PR #9, opened 2016-07-02
Handle channel updates when channel isn't cached

aemino:


abalabahaha:
The problem is, there should be no instance where the channel is not cached. This is not the right way to fix it.

aemino:
Other things like guild updates and message updates handle it like that.

PR #12, opened 2016-07-03
Fix permission check in default help command

btk5h:


abalabahaha:
Sorry, was unable to properly view PRs while abroad and duplicated this in one of my commits

PR #30, opened 2016-07-20
For for updating channel permission overwrites

hayleyxyz:
Previously the editChannelPermission() was returning 405 METHOD NOT ALLOWED from the API.

I have fixed the endpoint to use the correct URL, and added the required 'type' parameter to the request body.


abalabahaha:
Picked into dev, try to PR there next time

PR #33, opened 2016-07-26
Fix error where stream.destroy() function does not exist

hayleyxyz:
The following code:

`request.get(downloadUrl, function (error, response, body) {
    if(!error && response.statusCode == 200) {
        var bufferStream = new stream.PassThrough();
        bufferStream.end(body);

```
    vc.playStream(bufferStream);
}
```

});`

Generates the error:

`TypeError: stream.destroy is not a function`

This is because stream.PassThrough does not have a destory() method. I have updated Eris to check this function exists before calling it.

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -269,7 +269,7 @@ class Client extends EventEmitter {
     /**
     * Disconnects all shards
     * @arg {?Object} [options] Shard disconnect options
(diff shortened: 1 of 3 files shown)


abalabahaha:
Cherry-picked as f0b731b

PR #39, opened 2016-07-28
Another example

cernodile:
This is meant for people who find it difficult to use Promises, since the library uses lots of Promises, I believe this kind of example is required.


arunesh90:
lmfao

abalabahaha:
This is more of a general Javascript concept than a lib-specific one, so it's out of scope. It would be better to link people to one of the thousands of existing Intro To Promises 101 guides out there (or make your own).

Issue #45, opened 2016-08-02
editedTimestamp vs timestamp


hsiW:
`editedTimestamp` returns a Date String but is supposed to return a Number like `timestamp`
https://github.com/abalabahaha/eris/blob/master/lib/core/Message.js#L117
https://github.com/abalabahaha/eris/blob/master/lib/core/Message.js#L27

hsiW:
Fixed with https://github.com/abalabahaha/eris/commit/0b3c896fceb8069826c32c0fde766a274aa63a92

PR #46, opened 2016-08-02
Made editedTimestamp also return a number probably

hsiW:
Hopefully fixes https://github.com/abalabahaha/eris/issues/45


abalabahaha:
Only the first Date.parse is necessary

PR #52, opened 2016-08-04
Fix avatar update

MergHQ:
The base64 header was missing. Thanks to Aurieh#8140 for finding this bug :100:


hsiW:
This never really was a bug, you were just required to do it yourself.

MergHQ:
Ahh okay. Maybe it's still a good idea since Buffer.toString('base64') returns only the data and not the header?

hayleyxyz:
This explicitly hardcodes the mime type as image/jpeg, which will not be the case of all images.

hsiW:
_thats why part of why it was closed_

PR #54, opened 2016-08-07
Add more properties

cernodile:
Add some more properties, that I found while digging through data.


abalabahaha:
Modified and rebased

Issue #58, opened 2016-08-08
Prefix with @mention doesn't work with nickname


dsevillamartin:
If you have the prefix in the `CommandClient` set to anything including `@mention` it won't work if the bot has a nickname on that guild.

PR #62, opened 2016-08-09
Fix @mention prefix issue

MergHQ:
Should fix issue #58


hsiW:
This is fixed in b8208ddccacaa7bccdc44c4049a810a1bbb1b9f0 already

MergHQ:
It's not. Still broken

hsiW:
_ignore me, I'm not awake_

PR #67, opened 2016-08-10
Add "Bot" to authorization header

mnlkrs:


abalabahaha:
This already exists at https://github.com/abalabahaha/eris/blob/dev/lib/Client.js#L164, which is also user-account friendly

PR #77, opened 2016-08-26
Add a bunch of null checks

MergHQ:
This is not really a good fix for the caching problem, but still prevents crashes from happening.


abalabahaha:
This doesn't actually solve the caching problem, which shouldn't be happening in the first place

PR #89, opened 2016-09-13
Add deleted message to promise in deleteMessage

MergHQ:
This might be very useless, but some people wanted it so I thought I'd implement it too.


abalabahaha:
You can do this yourself... C'mon

PR #93, opened 2016-09-21
Add cooldown to Command

brussell98:
Because only bakalabahahas don't add colldowns to their commands.

### DIFF
--- a/lib/command/Command.js
+++ b/lib/command/Command.js
@@ -32,6 +32,7 @@ class Command {
     * In the above example, the user must not have administrator permissions, but must have manageMessages to use the command
     * @arg {Array<String>} [options.requirements.roleIDs] An array of role IDs that would allow a user to use the command
     * @arg {Array<String>} [options.requirements.roleNames] An array of role names that would allow a user to use the command
+    * @arg {Number} [cooldown] The cooldown between command usage in milliseconds
     * @returns {Command}
     */
     constructor(label, generator, options) {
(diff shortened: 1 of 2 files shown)


lib/command/Command.js
```diff
@@ -132,6 +137,17 @@ class Command {
...
 
+    cooldownCheck(userID) {
+        if(this.usersOnCooldown.has(userID)) {
+            return false;
+        }
+        this.usersOnCooldown.add(userID);
+        setTimeout(() => {
+            this.usersOnCooldown.delete(userID);
+        }, this.cooldown);
+        return true;
```
TehSeph: Shouldn't the two lines above only be set on a successful command (or at least after execute) imo it doesn't make sense to set them here during the check.

lib/command/Command.js
```diff
@@ -132,6 +137,17 @@ class Command {
...
     }
 
+    cooldownCheck(userID) {
+        if(this.usersOnCooldown.has(userID)) {
+            return false;
+        }
+        this.usersOnCooldown.add(userID);
+        setTimeout(() => {
+            this.usersOnCooldown.delete(userID);
+        }, this.cooldown);
```
TehSeph, line 147: Shouldn't the three lines above only be set on a successful command? (promise after execute?)
It doesn't make sense to set them here during the check stage since the command could fail and the user would still be on cooldown.

PR #98, opened 2016-10-01
Fix logic behind executeWebhook

eritbh:
ur logic is bad and u should feel bad

The logic of this if statement before looked like:

```
if (options.content and options.disableEveryone both exist):
  if (options.disableEveryone):
    remove everyones
else:
  if (bot.options.disableEveryone):
    remove everyones
```

What it should look like now:

```
if (options.content exists):
  if (options.disableEveryone exists && is true):
    remove everyones
  else if (options.disableEveryone doesn't exist && bot.options.disableEveryone is true):
    remove everyones
```

Fixes this error when calling `client.executeWebhook` with no content:

this error

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -584,7 +584,7 @@ class Client extends EventEmitter {
         if(!options.content && !options.file && !options.embeds) {
             return Promise.reject(new Error("No content, file, or embeds"));
         }
-        if(options.content && options.disableEveryone !== undefined ? options.disableEveryone : this.options.disableEveryone) {
+        if(options.content && (options.disableEveryone !== undefined ? options.disableEveryone : this.options.disableEveryone)) {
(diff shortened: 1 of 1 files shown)


hsiW (approved):

aemino:
Isn't the logic intentional? That way, setting disableEveryone to false when sending the message allows you to override the bot's default option...

eritbh:
@Programmix I changed it to what , as far as I can tell, was intended. Before, it was creating an error when `content` was undefined - which it should definitely not have been doing, since `content` is optional in the API docs.

What you're describing is actually what it does: If `options.disableEveryone` is defined, it will use that value; otherwise, it will use the client's value. SO yes, you can override the client value on a per-call basis.

aemino:
Oh, you're right. Should've looked at the actual code, rather than the pseudo-code. :P

abalabahaha:
;-;

PR #101, opened 2016-10-04
Fix insensitive command label option

### LINKED ISSUE #100: Command label with uppercase text disables caseInsensitive
`bot.registerCommand("PING", "Pong", {caseInsensitive: true});` command works with all uppercase, but the lowercase ping does not work. Command label must  be in all lowercase.

MergHQ:
Closes #100 (wow 100th issue 🍻 )


hsiW (commented):
Doesn't fully fix the issue, just fixes it if its all uppercase or all lowercase, if its a mix of both its still not fixed

PR #117, opened 2016-10-27
Automatically encode emoji if not already done.

Brayzure:
Can't imagine why `.getMessageReaction` needs a reaction as a parameter, but whatever.


abalabahaha:
Resolved conflicts and rebased.

PR #118, opened 2016-11-02
emit messageDeleteBulk event for bulk deletes

briantanner:
emits the messageDeleteBulk event with channel and array of message ids

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -456,14 +456,19 @@ class Shard extends EventEmitter {
                 if(!channel) {
                     break;
                 }
-                packet.d.ids.forEach((id) => {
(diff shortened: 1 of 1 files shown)


MinnDevelopment (approved):

Fuechschen:
I'm actually against this since it would only emit the message ids and not the cached message objects itself. I would prefer to either have 
1. both events emited 
or 
2. to have `messageDeleteBulk` emit an array of message objects.

Another option would be to pass a second parameter at the `messageDelete`-event indicating if it was triggerd through a bulk delete.

abalabahaha:
@briantanner 3a4a71d appears to have partially reverted ceb7356 and 5519fab

abalabahaha:
Fixed and rebased.

PR #121, opened 2016-11-06
Add argsRequired option

wrightk:


lib/command/Command.js
```diff
@@ -152,6 +154,10 @@ class Command {
             return;
         }
         if(args.length === 0) {
+            if(this.argsRequired) {
+            msg.channel.createMessage(`Invalid usage. Do \`${msg.prefix}help ${this.label}\` to view proper usage.`);
+            return;
```
MinnDevelopment: Nice indentation here.

abalabahaha:
I think this could be extended further, like defining a minimum number of args (maybe even an array of required arg types)... :thinking: 

You also forgot a few doc points here, too.

abalabahaha:
I'll merge this for now, but it would be nice to extend this later.

PR #122, opened 2016-11-07
Make null content not break createMessage

eritbh:
If `null` is passed as the content argument to `Client.createMessage()` it no longer breaks. If there's a file provided, it will send the file with no message; otherwise, it will error out as it should.


abalabahaha:
Patched and rebased.

PR #123, opened 2016-11-07
Add reaction docs to message

Ratismal:
How would one document sub-properties of an object when it comes to using `@prop`? It would be great if Object properties told you what properties they have, and I'd gladly implement this to improve readability.


amishshah:
This may be of use to you http://usejsdoc.org/tags-typedef.html

hsiW:
hydra get out

PR #134, opened 2016-11-23
Added an emoji cache

Ratismal:
Before abal instantly rejects:

PROS:
- have access to all emojis, and can retrieve via ID
- can do emoji searches based criteria such as guild or emoji name
- collection is built on call, so if you don't care about and/or won't use them this it won't affect you performance-wise

CONS:
- 14 more lines of code (including docs and newlines) 👀


abalabahaha:
This rebuilds the emoji list every time you want to access Client.emojis, which is pretty inefficient. Also having a Client-wide emoji cache isn't really necessary.

PR #135, opened 2016-11-23
Add emoji cache

Ratismal:
Now without recreating the collection every time it's called.

👀

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -41,6 +41,7 @@ try {
 * @prop {Object} channelGuildMap Object mapping channel IDs to guild IDs
 * @prop {Collection<Shard>} shards Collection of shards Eris is using
 * @prop {Collection<Guild>} guilds Collection of guilds the bot is in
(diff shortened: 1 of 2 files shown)


lib/Client.js
```diff
@@ -138,7 +140,7 @@ class Client extends EventEmitter {
             game: null,
             status: "offline"
         };
-
+
```
abalabahaha: :eyes:

abalabahaha:
Again, I still don't like the idea of a Client-wide emoji cache.

Also, you're missing the initial GUILD_CREATE/READY emojis

Ratismal:
I'm not missing GUILD_CREATE and READY emojis, as it updates every time the `guildCreate()` function is called.

I *am* missing the GUILD_DELETE emojis, and will push those soon.

I will not give up on this. ❤

abalabahaha:
Do it in this PR then, not a new one

PR #138, opened 2016-11-27
General Docs Fixes

hsiW:
You might not want the code box thing to be gone or whatever but atm its broked, I'd have properly fixed it instead of removing if I knew how exactly your site stuff handles it but I don't

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -424,8 +424,8 @@ class Shard extends EventEmitter {
                 * @prop {Object?} oldMessage.mentionedBy Object of if different things mention the bot user
                 * @prop {Boolean?} oldMessage.tts Whether to play the message using TTS or not
                 * @prop {String[]?} oldMessage.mentions Array of mentioned users' ids
-                * @prop {String[]?} oldMessage.roleMentions Array of mentioned roles' ids, requires client option moreMentions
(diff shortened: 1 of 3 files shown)


lib/gateway/Shard.js
```diff
@@ -424,8 +424,8 @@ class Shard extends EventEmitter {
                 * @prop {Object?} oldMessage.mentionedBy Object of if different things mention the bot user
                 * @prop {Boolean?} oldMessage.tts Whether to play the message using TTS or not
                 * @prop {String[]?} oldMessage.mentions Array of mentioned users' ids
-                * @prop {String[]?} oldMessage.roleMentions Array of mentioned roles' ids, requires client option moreMentions
-                * @prop {String[]?} oldMessage.channelMentions Array of mentions channels' ids, requires client option moreMentions
+                * @prop {String[]?} oldMessage.roleMentions Array of mentioned roles' ids.
+                * @prop {String[]?} oldMessage.channelMentions Array of mentions channels' ids.
```
eritbh, line 428: Aren't there generally not periods on these?

hsiW: Its hit or miss on what has periods and what doesn't in the docs

abalabahaha: The general trend is that the last sentence has no period

But whatever

PR #140, opened 2016-12-18
Message ID should be set to 'message_id'

### LINKED ISSUE #139: messageReactionRemove event returns message, with the id of undefined
`bot.on("messageReactionRemove", (message,emote,userid) => {
    console.log(message.id);
});`
will log undefined when a reaction is removed, *at least in the dev version*

foxbot:
This resolves #139

(I haven't tested this PR, but this should be the fix)


abalabahaha:
Incomplete, also duplicates b56e043

PR #148, opened 2016-12-31
Edited Guild.iconURL() to return a .webp icon.

cernodile:
I've noticed that guild icons also support .webp and .png format.


abalabahaha:
Let's wait for all Discord clients to fully support WebP

abalabahaha:
Whoops

PR #151, opened 2017-01-03
Documented Member Properties

hsiW:
Also added staticAvatarURL to Member

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -25,7 +25,7 @@ const VoiceState = require("./VoiceState");
 * @prop {Number} createdAt Timestamp of user creation
 * @prop {Boolean} bot Whether the user is an OAuth bot or not
 * @prop {String} username The username of the user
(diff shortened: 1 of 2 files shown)


lib/structures/Member.js
```diff
@@ -21,6 +21,15 @@ const VoiceState = require("./VoiceState");
 * @prop {String[]} roles An array of role IDs this member is a part of
 * @prop {User} user The user object of the member
 * @prop {Permission} permission The guild-wide permissions of the member
+* @prop {String} defaultAvatar The hash for the default avatar of a user if there is no avatar set
+* @prop {Number} createdAt Timestamp of user creation
+* @prop {Boolean} bot Whether the user is an OAuth bot or not
+* @prop {String} username The username of the user
+* @prop {String?} discriminator The discriminator of the user
```
abalabahaha: This shouldn't be nullable

hsiW: Tell that to your user docs c:

PR #155, opened 2017-01-06
Add support for shard latency checking

Brayzure:
- Each shard now stores the last time it sent a heartbeat, and when it was last acknowledged
- Shard getter returns latency if both values exist

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -31,6 +31,9 @@ var Inflator = typeof window !== "undefined" && Pako ? Pako.inflate : Zlib.infla
 * @prop {Number} guildCount The number of guilds this shard should be handling
 * @prop {Array<String>?} discordServerTrace Debug trace of Discord servers
 * @prop {String} status The status of the shard. "disconnected"/"connecting"/"handshaking"/"connected"
+* @prop {Number} lastHeartbeatReceived Timestamp when Discord last acknowledged a heartbeat
+* @prop {Number} lastHeartbeatSent Timestamp when shard last sent a heartbeat
+* @prop {Number} latency Current latency between shard and Discord
 */
 class Shard extends EventEmitter {
     constructor(id, client) {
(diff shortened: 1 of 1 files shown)


hsiW (changes requested):
Forgot to add docs for new things

hsiW (approved):

lib/gateway/Shard.js
```diff
@@ -42,6 +45,10 @@ class Shard extends EventEmitter {
         this.hardReset();
     }
 
+    get latency() {
+        return this.lastHeartbeatSent && this.lastHeartbeatReceived ? this.lastHeartbeatReceived - this.lastHeartbeatSent : null;
```
abalabahaha: `null` should probably be a JS number. Maybe `Infinity`?

lib/gateway/Shard.js
```diff
@@ -31,6 +31,9 @@ var Inflator = typeof window !== "undefined" && Pako ? Pako.inflate : Zlib.infla
 * @prop {Number} guildCount The number of guilds this shard should be handling
 * @prop {Array<String>?} discordServerTrace Debug trace of Discord servers
 * @prop {String} status The status of the shard. "disconnected"/"connecting"/"handshaking"/"connected"
+* @prop {Number} lastHeartbeatReceived Timestamp when Discord last acknowledged a heartbeat
+* @prop {Number} lastHeartbeatSent Timestamp when shard last sent a heartbeat
```
abalabahaha: Wording is a bit awkward for these

abalabahaha (changes requested):

lib/gateway/Shard.js
Brayzure: Yeah, I thought it was, but I couldn't figure out a better way to word it. Ideas?

lib/gateway/Shard.js
Brayzure: `Infinity` would certainly work, but I'm wondering if there's a better way to signify that the value the user is asking for doesn't make sense. Perhaps `NaN`?

hsiW (approved):

abalabahaha (approved):

PR #156, opened 2017-01-06
Fix stream stopping on remove()

cernodile:
I believe it came from unclosed socket, I've done several tests, and this one has worked out best for me.
This only affects SharedStream


cernodile:
Nevermind, it seems to randomly get broken AGAIN.

PR #158, opened 2017-01-06
Fixing voice state update

cryptiklemur:


aequasi:
If need be, i can alter this to just have the 4 properties, instead of the whole state.

PR #161, opened 2017-01-09
Voice and Client bug fixes

briantanner:
- Fix 'channels' of undefined in Client.getMessage
- Remove inconsistent promise rejection in leaveVoiceChannel
- Fix undefined pending guilds in readyHandler and disconnectHandler in VoiceConnectionManager

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -276,15 +276,7 @@ class Client extends EventEmitter {
     * @returns {GuildChannel | GroupChannel | PrivateChannel}
     */
     getChannel(channelID) {
-        const guild = this.guilds.get(this.channelGuildMap[channelID]);
-
-        let channel = guild ? guild.channels.get(channelID) : null;
-
-        if (!channel) {
-            channel = this.privateChannels.get(channelID) || this.groupChannels.get(channelID);
-        }
-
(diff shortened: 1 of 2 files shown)


briantanner:
The voice stuff may need looking at further

lib/Client.js
```diff
@@ -276,7 +276,16 @@ class Client extends EventEmitter {
     * @returns {GuildChannel | GroupChannel | PrivateChannel}
     */
     getChannel(channelID) {
-        return this.channelGuildMap[channelID] ? this.guilds.get(this.channelGuildMap[channelID]).channels.get(channelID) : this.privateChannels.get(channelID) || this.groupChannels.get(channelID);
+        const guildID = this.channelGuildMap[channelID];
```
brussell98: This seems to be uneeded

brussell98 (changes requested):

lib/Client.js
briantanner: Ahh right, thanks for pointing it out

abalabahaha:
Its probably better to figure out why channelGuildMap has data for guilds that don't appear to exist, rather than checking for null/undefined there

PR #169, opened 2017-01-20
Add guildMemberChunk Event

Fuechschen:
This adds an event when discord sends new members on startup or while caching new guilds.

This is especially useful if you want to add the users to a database. Since guildCreate is able to fire before all users are cached, it can lead to errors on new guilds when not all users are passed in guild.members and, after you added them to your database, you try to query for non-existent users.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -12,29 +12,29 @@ const Zlib = require("zlib");
 var EventEmitter;
 try {
     EventEmitter = require("eventemitter3");
-} catch(err) {
+} catch (err) {
     EventEmitter = require("events").EventEmitter;
 }
 var Pako;
 try {
     Pako = require("pako");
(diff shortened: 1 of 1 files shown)


lib/gateway/Shard.js
```diff
@@ -1014,10 +1014,17 @@ class Shard extends EventEmitter {
                         --this.getAllUsersCount[guild.id];
                     }
                 }
-                packet.d.members.forEach((member) => {
+
+                /**
+                 * Fired when discord sends members on startup
```
abalabahaha: Awkward wording. Maybe something like `Fired when Discord sends member chunks`

This also isn't limited to just startup

abalabahaha (changes requested):

hsiW (approved):

abalabahaha (approved):

PR #170, opened 2017-01-20
Add command error message

Fuechschen:
This adds an error message which is sent if the command generator somehow fails (e.g. if an exception is thrown or if the promise does not resolve).

I do not know if you are still accepting PRs for the CommandClient, however i thought this could be useful for some people. If you are rejecting, please consider this feature for the rewrite.

### DIFF
--- a/lib/command/Command.js
+++ b/lib/command/Command.js
@@ -37,6 +37,7 @@ class Command {
     * @arg {Number} [options.cooldown] The cooldown between command usage in milliseconds
     * @arg {String} [options.cooldownMessage] A message to show when the command is on cooldown
     * @arg {String} [options.permissionMessage] A message to show when the user doesn't have permissions to use the command
+    * @arg {String} [options.errorMessage] A message to show if the execution of the command handler somehow fails.
     */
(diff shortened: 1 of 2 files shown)


lib/command/CommandClient.js
```diff
@@ -140,6 +140,11 @@ class CommandClient extends Client {
                     if(resp != null) {
                         this.createMessage(msg.channel.id, resp);
                     }
+                }).catch((err) => {
+                    this.emit('warn', err);
```
abalabahaha: Use `"`

lib/command/Command.js
```diff
@@ -65,6 +66,7 @@ class Command {
         this.cooldown = options.cooldown || 0;
         this.cooldownMessage = options.cooldownMessage || false;
         this.permissionMessage = options.permissionMessage || false;
+        this.errorMessage = options.errorMessage || false;
```
abalabahaha: Default should be ""

abalabahaha (changes requested):

abalabahaha (approved):

PR #175, opened 2017-01-25
Add Client getUserSettings() request & cache from READY event if applicable

tigercat2000:


Brayzure:
A bit of an aside, but don't we usually do variables `likeThis` and not `like_this`? Idk if that impacts its ability to be merged.

tigercat2000:
You're right, actually, I wasn't paying attention to the styling.

lib/Client.js
```diff
@@ -1330,6 +1333,15 @@ class Client extends EventEmitter {
     }
 
     /**
+    * Get user settings (user accounts only)
+    * @arg {String} [userID="@me"] The user ID of the target user to get data for. "@me" refers to the logged in user's own settings.
```
abalabahaha: Pretty sure you can only get your own settings, so this is kinda useless

lib/Client.js
```diff
@@ -146,6 +146,9 @@ class Client extends EventEmitter {
         } else {
             this.voiceConnections = new VoiceConnectionManager();
         }
+
```
abalabahaha: Can you move these to L141-142

lib/Client.js
```diff
@@ -146,6 +146,9 @@ class Client extends EventEmitter {
...
+        this.userGuildSettings = [];
+        this.userSettings = [];
```
abalabahaha: If these are objects, it would be better to initialize them as {}

abalabahaha (changes requested):

lib/Client.js
tigercat2000: Sure.

abalabahaha (approved):

PR #177, opened 2017-01-29
Fix subcommands not being deleted when option is true

### LINKED ISSUE #176: Subcommands cannot delete messages.
Hi there,

I noticed that subcommands cannot control the deletion of their command message like top level commands can.

A simple test: Register any commands in this configuration:
bot -> command (deleteCommand: false) -> subcommand (deleteCommand: true)

if you then call **command subcommand** the command message wont be deleted.

Only top level commands can control deletion of the command message it would seem. 
Since Command.regiterSubcommand has the deleteCommand variable, is this intentional?

Thanks

MergHQ:
Closes #176

Don't know if this is really a good solution 🤔 , but it works! 🍻


lib/command/Command.js
```diff
@@ -156,13 +156,20 @@ class Command {
...
         if(!this.permissionCheck(msg)) {
             if(this.permissionMessage) {
                 msg.channel.createMessage(this.permissionMessage);
             }
             return;
         }
+        if (this.deleteCommand && msg.channel.guild && msg.channel.permissionsOf(client.user.id).has('manageMessages')) {
+            let originalExec = this.execute;
+            this.execute = (msg, args) => {
+                client.deleteMessage(msg.channel.id, msg.id);
```
abalabahaha: This can be done directly through the Message object, removing the need to pass client

lib/command/Command.js
```diff
@@ -156,13 +156,20 @@ class Command {
...
 
-    process(args, msg) {
+    process(args, msg, client) {
         if(!this.permissionCheck(msg)) {
             if(this.permissionMessage) {
                 msg.channel.createMessage(this.permissionMessage);
             }
             return;
         }
+        if (this.deleteCommand && msg.channel.guild && msg.channel.permissionsOf(client.user.id).has('manageMessages')) {
```
abalabahaha: Use a bitwise permissions check here

lib/command/Command.js
```diff
@@ -156,13 +156,20 @@ class Command {
...
+            let originalExec = this.execute;
+            this.execute = (msg, args) => {
```
abalabahaha: You shouldn't need to replace the execute method to delete the message

abalabahaha (changes requested):

lib/command/Command.js
MergHQ: If the main command has `deleteCommand: true` but the subcommand has `deleteCommand: false` the message will still be deleted even if the user calls the subcommand since the main command is processed before the subcommand.

I guess I could add the delete check before execute is called.

lib/command/Command.js
```diff
@@ -163,10 +163,12 @@ class Command {
             }
             return;
         }
-        if (this.deleteCommand && msg.channel.guild && msg.channel.permissionsOf(client.user.id).has('manageMessages')) {
+        if (this.deleteCommand && msg.channel.guild) {
             let originalExec = this.execute;
             this.execute = (msg, args) => {
-                msg.delete();
+                msg.delete().catch(e => {
```
abalabahaha: Should've left the preemptive check in

abalabahaha:
Whoops, outdated diff, ignore that.

EDIT: actually, can you add the perms check back?

abalabahaha (approved):

PR #187, opened 2017-02-21
VoiceState returns a class with id when not in voice channel;

TTtie:
use channelID property instead

### DIFF
--- a/examples/playFile.js
+++ b/examples/playFile.js
@@ -19,7 +19,7 @@ bot.on("messageCreate", (msg) => { // When a message is created
             bot.createMessage(msg.channel.id, "This command can only be run in a server.");
             return;
         }
-        if(!msg.member.voiceState) { // Check if the user is in a voice channel
+        if(!msg.member.voiceState.channelID) { // Check if the user is in a voice channel
             bot.createMessage(msg.channel.id, "You are not in a voice channel.");
             return;
         }


SapphoSys:
noice 10/10

matix2267:
Can't we instead add a property voiceState.isConnected or something like that?

hsiW:
why .-.

matix2267:
IMO it's better to check `voiceState.isConnected` (if it's `true` or `false`) than to check something like `if(voiceState.channelID !== null)` if you want to see if a user is connected to voice channel or not.

Or maybe not, dunno, I'm ok with either way, but I would change the line in the example to `if(msg.member.voiceState.channelID !== null)` instead of the current `if(!msg.member.voiceState.channelID)`

TTtie:
@matix2267 well, `VoiceState.channelID` is something like that too, also `VoiceState.channelID` will be undefined when not in a voice channel, so that's why `if(!msg.member.voiceState.channelID)`; we check if it is undefined.

TTtie:
^ also, where's [`VoiceState.isConnected`](https://github.com/abalabahaha/eris/blob/dev/lib/structures/VoiceState.js)

matix2267:
There isn't, I just asked if it couldn't be added, but it's more of a design issue and that is up to the library owner to decide. My subjective opinion is that `if(msg.member.voiceState.isConnected)` has the clearest meaning, next is `if(msg.member.voiceState.channelID !== null)` which I suggested in previous comment if we don't want to add another field to `VoiceState`, and then `if(!msg.member.voiceState.channelID)` is the most obscure because you can very easily miss `!`. Anyway, I'm just adding my 2 cents here, it's no big deal whatever you choose will be fine, it's just a matter of code style / examples.

abalabahaha:
I believe the line of reasoning is that if you're not in a voice channel (missing voiceState.channelID), you're probably not connected to a voice channel.

Feel free to open another PR/issue if you think this behavior should be changed.

PR #188, opened 2017-02-22
Allow embeds and attachments to be sent at the same time

mistval:
Discord allows this now.

Here's the testing I did: http://imgur.com/a/IyA88


abalabahaha:
Rebased as 112a9a0, thanks

PR #200, opened 2017-02-27
Possibly fixes voice clients disconnecting, may break other things.

macdja38:
This fixes the issue shown in this gif [here](https://pvpcraft.ca/i/57deh.gif)
Steps to reproduce:
1. have eris join a voice channel
2. call disconnect on the voice connection
3. remove the bot's permission to connect to that channel
4. have the bot attempt to join again
5. The bot will resolve the promise as if it joined, and pass back the original voice connection while in fact it has not, calling .play on this voice connection will fail.
(must be attempted without https://github.com/abalabahaha/eris/pull/199 having been merged, as this fixed a bug required to have the library attempt a voice connection that will fail)

Note that this very possibly breaks other things.

### DIFF
--- a/lib/voice/VoiceConnectionManager.js
+++ b/lib/voice/VoiceConnectionManager.js
@@ -77,7 +77,6 @@ class VoiceConnectionManager extends Collection {
             }
             this.pendingGuilds[data.guild_id].res(connection);
             delete this.pendingGuilds[data.guild_id];
-            connection.removeListener("disconnect", disconnectHandler);
         };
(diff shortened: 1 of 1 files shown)


abalabahaha:
Yeah, it feels like this would have some unintended consequences... I'll look more into alternate ways of fixing/avoiding this

macdja38:
Seems this fixed using voiceConnection.disconnect(null, false) as previously documented here https://abal.moe/Eris/docs/VoiceConnection while client.leaveVoiceChannel(voice.connection.channelID) was not broken.

Issue #204, opened 2017-03-03
Uncoherent behaviour on messageUpdate event and its documentation


enebe-nb:
In messageUpdate event documentation the message argument description says `The updated message. If oldMessage was undefined, it is not recommended to use this since it will be very incomplete` but...

Situation 1: Old message **is not** cached.
- In Documentation: is expected oldMessage to be undefined.
- In Reality: oldMessage contains a object with the only property being the id with the correct id string.

Situation 2: Old message **is** cached.
- In Documentation: is expected oldMessage to contains the previous message with all its properties.
- In Reality: oldMessage is defined and contains all properties except for its id.

hsiW:
The 2nd situation seems to be a bug, whereas the first one seems to be the docs not being fully clear, id should always(it doesn't if the message is cached currently) all the others are technically optional and can be undefined(even the the docs don't show it properly, the code to generate them does)

abalabahaha:
Situation 1: Docs are most correct, oldMessage should be ~~undefined~~ null

Situation 2: Since message.id should always exist, passing the message ID through oldMessage is kinda redundant

I updated the docs and code to reflect the intended behavior (ef73d67)

PR #205, opened 2017-03-03
Corrected messageUpdate to return the old message id if cached

hsiW:
Should resolve situation 2 of https://github.com/abalabahaha/eris/issues/204


abalabahaha:
Since message.id always exists, this is kinda redundant. I'll update the docs and code to match the intended behavior.

PR #212, opened 2017-03-14
Add reaction buttons to CommmandClient

7596ff:
This PR adds a reaction buttons option to command client, using similar generators for button actions.


abalabahaha:
The ratelimiting handler should already send requests in the order they were started, so the reaction chaining method seems a bit redundant

brussell98:
I feel like the `type=edit` `content` should allow a function instead of a string. I can see many situations where computation would be needed to display conditional output.

abalabahaha:
Rebased, thanks

Issue #225, opened 2017-04-21
Join message support


macdja38:
type: 7

Sample message:
```js
  {
    "attachments": [],
    "tts": false,
    "embeds": [],
    "timestamp": "2017-04-21T00:54:19.613000+00:00",
    "mention_everyone": false,
    "id": "304781866777706506",
    "pinned": false,
    "edited_timestamp": null,
    "author": {
      "username": "PvPCraft",
      "discriminator": "0625",
      "bot": true,
      "id": "103607047383166976",
      "avatar": "fe22d57db2c367ab96e466c1fd6d6a2b"
    },
    "mention_roles": [],
    "content": "",
    "channel_id": "304781394100355075",
    "mentions": [],
    "type": 7
  }
```

Discord code:
```js
function r(e) {
        var t = C(),
            n = R.default.extractTimestamp(e) % t.length;
        return S.default.locale.startsWith("en") ? t[n] : E.default.Messages.SYSTEM_MESSAGE_GUILD_MEMBER_JOIN
    }
```
```js
extractTimestamp: function(e) {
            return null == e ? 0 : r.default.fromString(e, !0).shiftRight(s).add(r.default.fromNumber(o)).toNumber()
        },
```

List of system messages
https://gist.github.com/macdja38/12dff2a6a496cd2d9185fec43c01d77b

the function C() returns an array with all the system messages starting at 0 and moving up if I remember correctly.

PR #226, opened 2017-04-22
Preliminary Type 7 Message Support

Brayzure:
Sets tentative content for type 7 (member join) messages until more information is known about it.

Addresses #225


abalabahaha:
Alternatively implemented with proper Message.content in 6b58105

PR #230, opened 2017-04-29
Do string checks

Ratismal:
String checks must be done, otherwise users will get "No content, file, or embed" whenever they use a string (because content.content and content.embed are undefined)


abalabahaha:
Fixed alternatively in 141fc4e

PR #231, opened 2017-04-30
Fix NSFW channel check

ghost:
Channels are dependent on their name. Channels are NSFW if one of 2
things is true:

1. the channel name is `nsfw`;
2. if 1 is not true, then the channel name must be prefixed with
`nsfw-`.

This means that:

- `nsfw` is true
- `nsfwa` is false
- `nsfw-test` is true
- `nsfw-` is true

This commit fixes broken behaviour that would allow the `nsfwa` case to
return true.

### DIFF
--- a/lib/structures/GuildChannel.js
+++ b/lib/structures/GuildChannel.js
@@ -89,7 +89,7 @@ class GuildChannel extends Channel {
     }
 
     get nsfw() {
-        return this.type !== 2 && this.name.length == 4 ? this.name == "nsfw" : this.name.startsWith("nsfw-");
+        return this.type !== 2 && (this.name.length === 4 ? this.name === "nsfw" : this.name.startsWith("nsfw-"));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


lib/structures/GuildChannel.js
```diff
@@ -89,7 +89,7 @@ class GuildChannel extends Channel {
     }
 
     get nsfw() {
-        return this.type !== 2 && this.name.startsWith("nsfw");
+        return this.type !== 2 && this.name.length == 4 ? this.name == "nsfw" : this.name.startsWith("nsfw-");
```
abalabahaha: The voice channel check is being combined with the length check, should use some parenthesis

Also use `===`

abalabahaha (changes requested):
Hmm somehow this didn't make it into the commit... thanks for pointing this out

ghost:
Changes made

PR #235, opened 2017-05-02
fix getGuildBans

na-2n:


abalabahaha:
I prefer to avoid unnecessary Array.map & object creation

Implemented alternatively in cd69156

PR #246, opened 2017-06-13
Add a fullname option?

bjxke:


Brayzure:
Seems a little trivial to me, but I'd use it if it was added.

abalabahaha:
At least make this a getter so it updates automatically

bjxke:
I screwed up with GitHub, I made it a getter on another pull request. #252

PR #252, opened 2017-06-28
User.fullName

bjxke:


itslukej:
This could be `<User>.tag`, alike https://discord.js.org/#/docs/main/stable/class/User?scrollTo=tag

bjxke:
Yeah, it's a bit similar to that.

itslukej:
also you forgot a `;` lol

abalabahaha:
fullname sounds weird

bjxke:
I'm not sure if `extendedName` would be better 😂  Or maybe something related to mentions 🤔

macdja38:
I don't really think that this is... necessary for the library but maybe .nameAndDiscriminator or .nameAndDiscrim

itslukej:
*discriminator dumb idiot

bjxke:
@macdja38 It could be for other people, for example other people adding modlogs to their bots.

sn8to:
~~You could be not lazy and just code the tag in your own code. It's not that hard, guys.~~

vladfrangu:
For those who want it already in Eris.
```javascript
  Object.defineProperty(Eris.Member.prototype, "tag", {
    get: function() { return `${this.username}#${this.discriminator}`; },
  });

  Object.defineProperty(Eris.User.prototype, "tag", {
    get: function() { return `${this.username}#${this.discriminator}`; },
  });
```

vladfrangu:
Also, to add on the naming: `tag` could sound like `mention`, or in my head it could. ~Yet I use the snippet from above~
This would be a nice feature, whatever it would be called, because convenience, but its not a **must-have** in the API.

foxbot (approved):

abalabahaha:
After some more discussion in Discord, I think we're going to reject this from the lib. You can always do a polyfill (see KingDGrizzle's code snippet) to replicate this behavior

PR #260, opened 2017-07-09
Fixed two issues and a few typos

LJNeon:
Sorry if I did this wrong somehow, first time using this feature.  Anyway tried to fix the two CommandClient bugs in the issues since both had well-documented causes and fixes.

**Edit:** Also fixed a few typos in the docs, though there are still two left I didn't know how to fix.

### DIFF
--- a/lib/command/CommandClient.js
+++ b/lib/command/CommandClient.js
@@ -169,7 +169,7 @@ class CommandClient extends Client {
     }
 
     onMessageReactionEvent(msg, emoji, userID) {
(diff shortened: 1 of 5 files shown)


macdja38:
Potentially it could be a bit more clear if you did !(thing1 && thing2 && thing3) but... other than that looks like it makes sense.

LJNeon:
Like that?

abalabahaha:
oh wait the `&&` should be `||`
whoops

Issue #237, opened 2017-05-08
Automatically generated Typescript typings


cryptiklemur:
Would be a wonderful thing to have.

abalabahaha:
same tbh

TTtie:
Honestly, that idea should be realized as soon as possible

aurieh:
Working on it at [aurieh/eris-typings](https://github.com/aurieh/eris-typings).
Still missing several features, like the eris factory (default export), webhook and embed types (`any` right now). As it's handwritten, it might (and will) have some issues.

EDIT: all done properly 🙏

abalabahaha:
Manually generated typings have been added, however automating typing generation through the doc generator is planned.

PR #264, opened 2017-07-16
Typescript typings (#237)

aurieh:
Features:
* Allows proper autocompletion when using VSCode intellisense and JS/TS LSP
* Allows library usage in TypeScript projects without type guessing
🙏

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,7 +5,7 @@ declare module 'eris' {
   type WebhookPayload = {
     content?: string,
     file?: { file: Buffer, name: string } | Array<{ file: Buffer, name: string}>,
-    embeds?: Array<Embed>,
+    embeds?: Array<EmbedOptions>,
     username?: string,
     avatarURL?: string,
     tts?: boolean,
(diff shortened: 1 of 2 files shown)


TTtie (commented):
:thinking: This should be optional IMO, I know it is always present in a message embed object but people omit it when sending embed messages.

Talking about line 19.

aurieh:
@TTtie I might separate it and some other type aliases (like EmojiOptions), as properties differ when sending and receiving data.

TTtie:
👌

cryptiklemur:
Is this easy enough to update every time abal updates something @aurieh? Is this automateable?

abalabahaha:
There are plans for integrating this into the docs generator but its manual for now. The API isn't changing much anyway

cryptiklemur:
👍 lets merge then :D

aurieh:
@aequasi @abalabahaha
It's really easy to add methods and props to definitions file.
About generating definitions, it's possible but fairly hard.
Take [Microsoft/dts-gen](https://github.com/Microsoft/dts-gen) as an example. Definitions it generates are lacking as it can't infer properties like a human can.
Along with that, even if you were to make something that works well, current documentation would be hard to work with, as some types and props are missing or are just `Object`. With webhook payload types, you can redirect users to discord api developer docs, but a generator can't infer what properties it should have.
What I'm saying is that it's only possible to write quality definition files by hand.

Issue #270, opened 2017-07-29
Suggested improvement for Command cooldowns


LJNeon:
So at the moment how cooldowns work is that every time you use a command it checks if the command is on cooldown and if so returns a cooldown message.  However, if someone spams the command the bot is quickly rate limited by all the cooldown messages.  This presents a problem especially since one of the main uses of cooldowns is to help prevent rate-limiting the bot.

**Suggested changes:**
`this.cooldownMessage` can be set to a function (or a string still) that is passed `msg`, `totalTime`, and `remainingTime` (param names could be changed) and returns a valid reply object or string. (object for things like embeds/attachments)
 * Pros: People will be less likely to spam commands and will be less annoyed if they at least know how much longer they have to wait.  Cooldown messages can have more detail added to them.
 * Cons: adds complexity to the cooldowns
 
Cooldowns restart every time the command is attempted to be used.
 * Pros: much stronger against spam
 * Cons: might annoy a few people when they have to wait longer because they were impatient

Cooldowns only return a message the first time it's used, after that it just doesn't respond until the cooldown ends.
 * Pros: much less likely to rate-limit the bot with spam
 * Cons: may conflict with the first suggestion, more likely to confuse people when the command just isn't doing anything.

You could implement these as individual options, replace the current system with this, etc.  All these changes would be trivial to code.

Issue #269, opened 2017-07-29
Command {argsRequired: true} sends incorrect messages


LJNeon:
The first issue is that it tells you to use the help command even when no help command is set, which is misleading.  The second problem is that when you have a subcommand like `.random list` it tells you to use `.help list` when it should say to use `.help random list`.
![screenshot 2017-07-28 at 10 31 38 pm](https://user-images.githubusercontent.com/23249107/28742412-8be21fd6-73e4-11e7-9124-fa878c63777a.png)

PR #273, opened 2017-08-06
Several CommandClient changes

LJNeon:
*Note: I have bug tested all of these new features thoroughly.  I also have valid use cases for all of them and will be using almost all of them.  If there are any issues or you want different behaviors just tell me and I will do my best to fix any problems.*

I implemented all of the suggested changes (by me) in issue #270.  However, I made them optional and didn't change the default behavior.  All existing bots will work the exact same with or without this update.  To recap on these changes:
 * `Command.cooldownMessage` can be optionally set to a function that is passed the Message object and returns a message to send.  I also gave `Command.permissionMessage` and `Command.errorMessage` the same behavior.
 * All the options inside of `Command.requirements` can be optionally set to a function that returns an array/object and is passed the Message object.
 * I added an option to Commands called `restartCooldown`.  When set to true the Command's cooldown will reset every time someone attempts to use it while on cooldown.
 * I added an option to Commands called `cooldownReturns`.  When set to a number other than 0 it will only return a cooldown message x times.  When the cooldown expires this is reset.

I also fixed the bug reported (by me) in issue #269.  The invalid usage message was already fixed to not tell you to use a help command that doesn't exist, but when using `.cmd subcmd` it would still tell you to use `.help subcmd` instead of `.help cmd subcmd` so I fixed that.

**Edit:** I also changed the event emitted when a command generator fails from `warn` to `error`, since logically an error should emit the error event.

### DIFF
--- a/lib/command/Command.js
+++ b/lib/command/Command.js
@@ -22,8 +22,8 @@ class Command {
     * @arg {String} [options.fullDescription="No full description"] A detailed description of the command to show in the default help command
     * @arg {String} [options.usage] Details on how to call the command to show in the default help command
     * @arg {Object} [options.requirements] A set of factors that limit who can call the command
-    * @arg {Array<String>} [options.requirements.userIDs] An array of user IDs representing users that can call the command
-    * @arg {Object} [options.requirements.permissions] An object containing permission keys the user must match to use the command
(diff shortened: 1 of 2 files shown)


LJNeon:
@hsiW wouldn't this also have the bug tag to match issue #269?

PR #275, opened 2017-08-09
Add toString methods to structures

minemidnight:


macdja38:
Assuming abal decides to go this way, some thought should be put into toString() turning channels and users into mentions or strings.

LJNeon:
Personally, I'd rather use @macdja38's [`.toJSON()`](https://github.com/macdja38/pvpcraft/blob/master/PvPCraft.js#L19-L47) over this.

minemidnight:
@macdja38 I wanted to keep the functionality that Jake previously put in his pull request recently for a "full name" property where you can get the username and discriminator, and mention is already a property where a user's username plus discriminator is not. 

@LJNeon Do you mean you would rather have the toString method return the result of toJSON? If so, I believe that would be completely pointless as some messages may be over 2000 characters with the stringified JSON. If that's not what you meant I'm afraid you don't understand toString.

LJNeon:
@minemidnight I made the [`.toJSON()`](https://github.com/macdja38/pvpcraft/blob/master/PvPCraft.js#L19-L47) a link to the exact code that I meant, please refer to that.

LJNeon:
How it essentially works is that it returns `"[ Eris ClassName ]"` for all the classes in Eris (for example a channel would turn into `"[ Eris Channel ]"` and a collection of roles would turn into `"[ Eris Collection<Role> ]"`).  Note that the linked code doesn't have the aforementioned collection support but it's trivial to add.  I would prefer this because it's more consistent, predictable, and readable; and doesn't return different things for different classes.

abalabahaha:
Just plain out returning `this.label` or `this.id` is pretty ambiguous and offers no explanation as to what's being stringified. At the very least, I'd prefer it be more like `[Command label]`/`[Shard <id>]`

minemidnight:
I was thinking of people doing stuff such as:
`Current Shard ${message.channel.guild.shard}` or something like that.

abalabahaha:
Closing in favor of #281

PR #280, opened 2017-08-17
Send type when editing status.

bjxke:
It is required by a new update to send the type when editing the game.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -1670,7 +1670,7 @@ class Shard extends EventEmitter {
             this.presence.status = status;
         }
         if(game !== undefined) {
-            game.type = +!!game.url
+            game.type = game.url ? 1 : 0;
             this.presence.game = game;
         }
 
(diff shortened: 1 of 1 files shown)


lib/gateway/Shard.js
```diff
@@ -1670,6 +1670,7 @@ class Shard extends EventEmitter {
             this.presence.status = status;
         }
         if(game !== undefined) {
+            game.type = +!!game.url
```
enhsaihan-d: i'd use
```javascript
game.type = +Boolean(game.url)
```
!! is bad practice

abalabahaha:
Would be nice to explicitly pick 0/1 instead of converting a Boolean to Number

enhsaihan-d:
```javascript
game.type = Boolean(game.url)
``` 
or
```javascript
game.type = !!game.url
```
it is then, i recommend the first one

bjxke:
Either way it works.

The next commit won't be bad practice, so that basically settles @necrowy 's point ;)

PR #281, opened 2017-08-17
Add toString methods to structures (edited)

LJNeon:
I edited it to match the style abal asked for in the original pull request.  I also added support for Maps/Sets and made it so that it would automatically affect new properties added to Eris at any point.  I did test this to make sure it worked.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,7 @@ function Eris(token, options) {
 Eris.Base = require("./lib/structures/Base");
 Eris.Bucket = require("./lib/util/Bucket");
 Eris.Call = require("./lib/structures/Call");
+Eris.CategoryChannel = require("./lib/structures/CategoryChannel");
 Eris.Channel = require("./lib/structures/Channel");
 Eris.Client = Client;
 Eris.Collection = require("./lib/util/Collection");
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -6,6 +6,28 @@ function Eris(token, options) {
...
 }
 
+function toString() {
+	var str;
+	Object.keys(Eris).filter(eris_prop => typeof Eris[eris_prop] === "function" && this instanceof Eris[eris_prop]).sort((a, b) => Eris[b].prototype instanceof Eris[a] ? 1 : 0).forEach(eris_prop => {
+		if(!str) {
+		    var ID = "";
+			if(this.id) {
+			    ID = " " + this.id;
+			}else if(this.user) {
```
macdja38: potentially this could maintain a consistent style with the rest of the lib (eg in this case space after })

for example https://github.com/abalabahaha/eris/blob/master/lib/Client.js#L951

LJNeon: I doubt something like that would make any major difference but sure I'll change that to match.

LJNeon:
I just added `Eris.CategoryChannel` (to index.js), removed the conflicts, and am a total idiot who replaced spaces with tabs .-.

abalabahaha:
spaces are the only real indentors (at least for this project)

LJNeon:
Well I managed to turn this into only one commit and cleaned it up, should be good now.

PR #282, opened 2017-08-20
Use new NSFW logic

Thewsomeguy:
Been a few weeks since NSFW was changed to not require `nsfw-` prefix.. should finally use new logic.

### DIFF
--- a/lib/structures/GuildChannel.js
+++ b/lib/structures/GuildChannel.js
@@ -49,6 +49,7 @@ class GuildChannel extends Channel {
         this.position = data.position !== undefined ? data.position : this.position;
         this.bitrate = data.bitrate !== undefined ? data.bitrate : this.bitrate;
         this.userLimit = data.user_limit !== undefined ? data.user_limit : this.userLimit;
+        this.nsfw = data.nsfw !== undefined ? data.nsfw : false;
         if(data.permission_overwrites) {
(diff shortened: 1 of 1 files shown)


macdja38:
IDK if this is still the case but from what I heard a little while back the official client is remaining backwards compatible with channels that are prefixed with nsfw-

Thewsomeguy:
yeah, the CHANNEL_UPDATE event returns `nsfw: true` for nsfw- prefixed channels, but the current check doesn't work for non-prefixed channels

macdja38:
Maybe it would be a good idea for the check to work with both?

lib/structures/GuildChannel.js
```diff
@@ -49,6 +49,7 @@ class GuildChannel extends Channel {
         this.position = data.position !== undefined ? data.position : this.position;
         this.bitrate = data.bitrate !== undefined ? data.bitrate : this.bitrate;
         this.userLimit = data.user_limit !== undefined ? data.user_limit : this.userLimit;
+        this.nsfw = data.nsfw !== undefined ? data.nsfw : false;
```
macdja38: Do you think something like this could work, so that it maintains consistency with the client and displays channels that start with nsfw- as nsfw?
```js
this.nsfw = this.type !== 2 && ((this.name.length === 4 ? this.name === "nsfw" : this.name.startsWith("nsfw-")) || data.nsfw);
```

PR #287, opened 2017-08-29
Fix case insensitive aliases

### LINKED ISSUE #286: Command client caseInsensitive option not working for aliases
Using the following code snippet, I'm currently unable to get aliases to be case insensitive despite setting the option for it. 

```javascript
bot.registerCommand('hugs', (msg, args) => {
  console.log('working');
}, {
  aliases: ['hug', 'cuddles'],
  caseInsensitive: true
});
```

For example, `+Hugs` or `+HuGs` will function as expected, whereas `+Hug` or `+hUg` will not be detected. 

*NOTE: `+` is the command prefix for my bot.*

germanoeich:
This fixes #286 

I feel this is a rather sloopy way to fix it, but since commandAliases only holds the name of the command, I tried to minimize the checks and keep things the way the are now. If people here feel the right way to go about this is to add the caseInsensitive flag to the commandAlias property, let me know and I'll edit this PR.

### DIFF
--- a/lib/command/CommandClient.js
+++ b/lib/command/CommandClient.js
@@ -140,8 +140,9 @@ class CommandClient extends Client {
             var args = msg.content.replace(/<@!/g, "<@").substring(msg.prefix.length).split(" ");
             var label = args.shift();
             label = this.commandAliases[label] || label;
+            var lowerCaseAlias = this.commandAliases[label.toLowerCase()]
             var command;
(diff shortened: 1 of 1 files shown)


abalabahaha:
Eh, the statement is kinda weird and long now. Maybe making a function like `resolveCommand(label)` would be cleaner and more readable

germanoeich:
Should be better now.

PR #296, opened 2017-09-09
Allow null as well as undefined for game in editStatus

macdja38:
Fixes issue when setting the status to null as described in the docs.

https://abal.moe/Eris/docs/Client#function-editStatus

If this is to be changed requiring users to use a status of undefined this should be done in a major change as it changes the api.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -1699,6 +1699,9 @@ class Shard extends EventEmitter {
     * @arg {String} [game.url] Sets the url of the shard's active game
     */
     editStatus(status, game) {
+        if (game === null) {
+            game = undefined;
+        }
         if(game === undefined && typeof status === "object") {
             game = status;
             status = undefined;


LJNeon:
Wouldn't it be easier to just change `=== undefined` to `== undefined` instead of adding a whole new if statement?

macdja38:
@LJNeon Possibly but that would add ambiguity and require methods this method calls to be able to handle both undefined and null, which could increase the number of bugs in the future. Having two values for undefined / null is not a great feature, arguably it should not be part of javascript but their is definitaly no point in requiring everything past the first line of that function to check for both undefined then null, then sent only one to discord in the end to ensure that the feature behaves consistently in the future.

macdja38:
Abal has a better way of fixing this.

abalabahaha:
Ook just took a second look and realized this PR does not do what I thought it would do. Setting game to null is the intended way to clear your game, so this is going to break that.

macdja38:
Yup, I mis-interpreted the game !== undefined as the method requiring undefined to clear the game. Sorry about the broken PR.

PR #299, opened 2017-09-13
fix some things in typings

na-2n:


index.d.ts
```diff
@@ -570,9 +570,9 @@ declare module 'eris' {
   }
 
   export class Collection<T extends { id: string }> extends Map<string, T> {
-    baseObject: new () => T;
+    baseObject: new (...args: any[]) => T;
     limit?: number;
-    constructor(baseObject: new () => T, limit?: number);
+    constructor(baseObject: new (...args: any[]) => T, limit?: number);
```
aurieh, line 575: The typings use `Array<any>` in most places, you should keep this in mind

na-2n: using `Array` for simple types is meme tho

aurieh: Then make a PR changing all array types I used to `type[]` 🙃

na-2n: me too thanks

cryptiklemur: `type[]` is standard for simple types in tslint

aurieh: @aequasi I'm fully aware of that, couldn't break my habit of using Array<T> before. I might PR the change at some point, unless someone else wants to do it. Before that, however, I would like the typings' style to stay consistent.

na-2n:
closing cuz aurieh added my changes to his pull

PR #301, opened 2017-09-18
Split GuildChannel into TextChannel and VoiceChannel

LJNeon:
Abal mentioned that he wanted to do this but didn't have the time. Other than adding aliases for `Client.joinVoiceChannel()` and `Client.leaveVoiceChannel()` I have made the minimum required changes.

I only tested this minimally to make sure that it populated `Guild.channels` properly. Due to both that and the fact that this is a potentially breaking change to Eris, I would highly recommend thoroughly testing this PR before merging it. I will be testing this a lot more myself but outside help is always welcome.

I'm completely open to suggestions from abal or anyone else, though obviously, abal's suggestions take priority.

I don't know TypeScript and wasn't able to update the typings, someone else will have to do that.


LJNeon:
I also moved the text channel only methods (like `createMessage`) from `Channel` to `TextChannel` and `PrivateChannel` since `GroupChannel` is extended from `PrivateChannel`. This made it so `CategoryChannel`s and `VoiceChannel`s don't have methods you can't actually use.

LJNeon:
If anyone wants to help me test this (for example none of my bots use voice channels and that needs to be tested too)
`npm install git://github.com/LJNeon/eris.git#dev`

PR #306, opened 2017-09-22
Adding ability to create channels in categories

cryptiklemur:


abalabahaha:
parentID not parent

cryptiklemur:
Fixed

PR #307, opened 2017-10-02
TypeScript typings additions and fixes

pyrox18:
Added a Promise\<string> return type to CommandGeneratorFunction in the typings file. This enables generator functions that use the ES7-style async/await function syntax to specify the required Promise return type for async functions.

Some example code:

```typescript
fakePromise(): Promise<string> {
  return Promise.resolve("Resolved promise");
}
```

```typescript
async generatorFunc(msg: Message, args: string[]): string {
  //                                               ^
  // TS error: The return type of an async function or method must be the global Promise<T> type
  let str: string = await fakePromise();
  return str;
}
```

```typescript
async generatorFunc(msg: Message, args: string[]): Promise<string> {
  // this is correct...
  let str: string = await fakePromise();
  return str;
}

// ...but this doesn't work (this PR fixes this error)
// note: bot is an Eris.CommandClient
bot.registerCommand("test", generatorFunc);
// TS error: Argument of type '(msg: Message, args: string[]) => Promise<string>' is not
// assignable to parameter of type 'CommandGenerator'
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -242,7 +242,7 @@ declare module "eris" {
     permissionMessage?: string | GenericCheckFunction<string>;
     errorMessage?: string | GenericCheckFunction<string>;
   }
-  type CommandGeneratorFunction = (msg: Message, args: string[]) => Promise<string> | string | void;
+  type CommandGeneratorFunction = (msg: Message, args: string[]) => Promise<string> | Promise<void> | string | void;
   type CommandGenerator = CommandGeneratorFunction | string | string[] | CommandGeneratorFunction[];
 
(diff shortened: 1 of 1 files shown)


aurieh:
You might also want to add `Promise<void>`

pyrox18:
Found an issue where CommandClient.registerCommand's return type was specified as void instead of a Command - fixed it in https://github.com/abalabahaha/eris/pull/307/commits/b3a2a803d5397044eacecc923c3c6571604f8903. Also changed the PR title to reflect the current nature of the PR.

PR #312, opened 2017-10-05
Update to typings

cryptiklemur:


aurieh:
Oh, great find! I've missed out on making everything `*Options` optional ever since I've separated event payload and option types. ~~You should also change `EmojiOptions`~~ Nevermind, that's how it should be.

cryptiklemur:
using `Partial<RoleOptions>` would work too, but this seems better.

abalabahaha:
Hacktoberfest is in the air~

PR #317, opened 2017-10-13
 Letting shards emit events

cryptiklemur:


cryptiklemur:
ohgod. Ignore. i'll re-pr without the style changes.

PR #319, opened 2017-10-14
Fix double event emitting

TTtie:
Fixes the behavior of https://github.com/abalabahaha/eris/blob/9d1016c918dd7881269b9afd05471fde1f87f7e8/lib/gateway/Shard.js#L1833-L1836


TTtie:
the (╯°□°）╯︵ ┻━┻ commit is just a semicolon addition to not make eslint cry

PR #325, opened 2017-10-30
Added tag property to Member and User

PassTheMayo:
The `.tag` property is used to show the full username and discriminator of a user. For example, if you wanted to get `Username#1337`, you could simply use `User.tag` or `Member.tag` instead of `User.username + "#" + User.discriminator`. This is mostly for ease of use. I have kept the same style of your code, but simply added this property.


abalabahaha:
Literally #252

PassTheMayo:
:O

PR #326, opened 2017-10-30
fix the special handling of days via query string param in /prune

DasWolke:
Discord uses query string params for the days to prune in POST /guilds/:id/prune
see [prune docs](https://discordapp.com/developers/docs/resources/guild#begin-guild-prune)

### DIFF
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -116,7 +116,7 @@ class RequestHandler {
                             throw new Error("Invalid file object");
                         }
                     } else if(body) {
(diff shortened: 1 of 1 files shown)


DasWolke:
On a sidenote: Why not add something like a useParams parameter to the request method of the request handler ? 
That way you could avoid that if check on [line 119](https://github.com/abalabahaha/eris/blob/dev/lib/rest/RequestHandler.js#L119).. 
To not have to change too much you could set the default value of it to `method === "GET"`
This would also make it easier in case future endpoints do something similar :D

PR #328, opened 2017-11-10
Added cooldown exclusions for CommandClient

4lch4:
When registering a command with a command client, the client now excepts a cooldown exclusions object, which contains an array of user, guild, and channel ids that aren't affected by command cooldowns.


4lch4:
This is my first ever pull request so if any changes are necessary just let me know and I'll get it done.

lib/command/Command.js
```diff
@@ -207,7 +215,19 @@ class Command {
         return !req;
     }
 
-    cooldownCheck(userID) {
+    cooldownExclusionCheck(msg) {
+        if(this.cooldownExclusions.userIDs.indexOf(msg.author.id) > -1) return true;
+        if(this.cooldownExclusions.guildIDs.indexOf(msg.channel.guild.id) > -1) return true;
+        if(this.cooldownExclusions.channelIDs.indexOf(msg.channel.id) > -1) return true;
```
Brayzure: Perhaps reorder these three checks? I'd imagine channel exclusions are going to be more likely than user exclusions, which will be more likely than guild exclusions.

PR #333, opened 2017-11-25
Fixed zlib/pako compression issues with uws

TheAkio:
Duplicate of #327 Because I moved the commit to another branch on my fork.

Eris wasn't working with compression set to true and UWS installed. This method I used works with zlib-sync and pako but still has issues with the regular zlib one in combination with UWS but somehow still works but throws weird errors. In addition this change will atm only apply if UWS is actually used and will not affect anything using the regular internal WebSocket library.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -34,8 +34,10 @@ try {
     } catch(err) { // eslint-disable no-empty
     }
 }
+var uws = false;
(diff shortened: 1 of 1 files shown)


TheAkio:
Made an issue (https://github.com/abalabahaha/eris/issues/336). Closing this here.
grafik

Issue #329, opened 2017-11-13
Clean up REST errors


PassTheMayo:
Hello. This really isn't an urgent thing, but I want to suggest something. I have been getting tons of 403 errors (no permission to send messages) a lot. When I was with Discord.js, there was a way to properly check the error code or message. With Eris, it seems a bit more *finicky*. Here is an example of a 403 error in Eris:
![image](https://user-images.githubusercontent.com/16949253/32711425-6f8eea16-c804-11e7-9bd6-bffe8218fd97.png)
As you can see here, there is also a problem with this. I am using `process.on("unhandledRejection")` but part of the error seems to bypass it, including a duplicate of it goes through the event handler. The red part at the top is the part that bypassed the `unhandledRejection`, while the green at the bottom means it was properly logged using the event. I believe that this is an error with the lib. Here is my code:
```js
process.on("unhandledRejection", (error) => {
	console.log(error);
});
```
With the fix to the duplicate errors, I also wish to request that the errors get some sort of fix-up, so that it can be better parsed by the bot. For example, here is an error that would be easily be parsable by the bot, and could be handled properly:
```js
{ Error: 50013 Missing Permissions
    at repl:1:13
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at REPLServer.defaultEval (repl.js:239:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:440:10)
    at emitOne (events.js:120:20)
    at REPLServer.emit (events.js:210:7)
    at REPLServer.Interface._onLine (readline.js:282:10)
    at REPLServer.Interface._line (readline.js:631:8) type: 'DiscordAPI', code: 50013, name: 'Missing permissions' }
```
That was just an example that I created within the Node.js REPL client. I believe that the errors could be cleaned up easily, and wouldn't be too hard to do. Thanks.

abalabahaha:
Your "duplicate error" appears because you aren't handling the rejection of the Promise returned by the `createMessage()` call, and NodeJS 8+ is telling you so, as you can see on the first line of the screenshot you posted. Adding a standard Promise `.catch(func)` should fix that.

Regarding formatting, the error you posted is technically in that format, but it seems adding the response as a property (`resp`) of the error is confusing the NodeJS inspector. For the error you screenshotted, `error.message` would return a string that looks like (in human form):
```
[...]

PR #335, opened 2017-11-29
Cleaned up REST Errors

LJNeon:
See #329.

As far as I know these were the only changes that were needed, if I missed anything please comment.


abalabahaha:
A nicer/less hacky version is in the works, thanks tho

PR #339, opened 2017-12-14
Fix Channel construction on REST mode

enebe-nb:
On `lib/Client.js#1765` and `lib/Client.js#1769` the constructors receive a guild object, not the client.

On `lib/structures/CategoryChannel.js#25` in REST mode `guild` is null, so here we avoid a undefined access error

On `lib/structures/GuildChannel.js#25` If guild is null (always in REST mode) then we store a mock to have access to guild id information.

### DIFF
--- a/lib/structures/CategoryChannel.js
+++ b/lib/structures/CategoryChannel.js
@@ -22,11 +22,13 @@ class CategoryChannel extends GuildChannel {
     constructor(data, guild) {
         super(data, guild);
         this.channels = new Collection(Channel);
-        guild && guild.channels.forEach((channel) => {
-            if(channel.parentID === this.id) {
-                this.channels.add(channel);
-            }
-        });
+        if(guild) {
+            guild.channels.forEach((channel) => {
+                if(channel.parentID === this.id) {
+                    this.channels.add(channel);
+                }
+            });
+        }
 
         this.update(data);
     }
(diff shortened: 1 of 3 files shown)


lib/structures/GuildChannel.js
```diff
@@ -22,7 +22,9 @@ const PermissionOverwrite = require("./PermissionOverwrite");
 class GuildChannel extends Channel {
     constructor(data, guild) {
         super(data);
-        this.guild = guild;
+        if (!guild && data.guild_id) {
+            this.guild = {id: data.guild_id};
+        } else this.guild = guild;
```
LJNeon: Would be nice if you used brackets for the `else`.

lib/structures/CategoryChannel.js
```diff
@@ -22,7 +22,7 @@ class CategoryChannel extends GuildChannel {
     constructor(data, guild) {
         super(data, guild);
         this.channels = new Collection(Channel);
-        guild.channels.forEach((channel) => {
+        guild && guild.channels.forEach((channel) => {
```
LJNeon: This would be better as an `if` statement.

LJNeon (changes requested):
Looks good, but would be nice if you held to the coding format of the rest of the lib. There's an eslint file you can use.

LJNeon (approved):

PR #341, opened 2017-12-15
Add ready/disconnect event typings

DusterTheFirst:
ready and disconnect events from the client were missing

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -732,6 +732,7 @@ declare module "eris" {
     public searchGuildMessages(guildID: string, query: SearchOptions): Promise<SearchResults>;
     // tslint:disable-next-line
     public on(event: string, listener: Function): this;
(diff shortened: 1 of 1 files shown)


aurieh:
You also have to implement the ready event in Client and Shard classes, as `Emittable` is an interface.

DusterTheFirst:
bump?

abalabahaha:
Trying to find a Typescript user to review :|

index.d.ts
```diff
@@ -317,7 +318,8 @@ declare module "eris" {
     connected_accounts: Array<{ verified: boolean, type: string, id: string, name: string }>;
   }
 
-  interface Connection {
```
aurieh: Was this intentional?

aurieh (changes requested):

index.d.ts
DusterTheFirst: no

PR #344, opened 2017-12-20
Implement CommandClient Command Hooks

Brayzure:
When working on another feature, I realized I wanted command logic to be applied before the checks were made, but after the command has begun processing. This led me to take a shot at implementing hooks into commands. Hooks are meant to be used by experienced users that want to further modify the behavior of their commands. Each hook and when it's used is detailed below.

`beforeCheck`: Called when the command has begun processing, but has not done any checks.
`beforeExecute`: Called after all checks have passed, but before the command has been executed.
`afterFailedCheck`: Called after any check has failed, but before the command has sent a response.
`afterExecute`: Called after the command has finished executing.
`afterSuccess`: Called after the command has successfully executed, but before the command has sent a response.
`postProcess`: Called after a response has been sent, regardless of whether or not the command failed to execute.

**Sample Hook Execution for Successful Command**
- `beforeCheck`
- `beforeExecute`
- `afterExecute`
- `afterSuccess`
- `postProcess`

**Sample Hook Execution for Failed Command**
- `beforeCheck`
- `afterFailedCheck`
- `postProcess`

This is an initial attempt, and will likely receive further updates to naming, exact timing, and other implementation details, but I'm submitting the PR now to solicit feedback.

### DIFF
--- a/lib/command/Command.js
+++ b/lib/command/Command.js
@@ -23,12 +23,10 @@ class Command {
     * @arg {String} [options.fullDescription="No full description"] A detailed description of the command to show in the default help command
     * @arg {String} [options.usage] Details on how to call the command to show in the default help command
     * @arg {Object} [options.hooks] A set of functions to be executed at different times throughout the command's processing
-    * @arg {Function} [options.hooks.beforeCheck] A function that is executed before any permission or cooldown checks is made
-    * @arg {Function} [options.hooks.beforeExecute] A function that is executed after all checks have cleared, but before the command is executed
-    * @arg {Function} [options.hooks.afterExecute] A function that is executed after the command is executed, regardless of the final failed state of the command
-    * @arg {Function} [options.hooks.afterSuccess] A function that is executed after the command has successfully executed, but before any response is posted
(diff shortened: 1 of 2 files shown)


Brayzure:
Hooks modified as follows:

Hooks
-
- `preCommand` - Run before any checks are performed. Accepts a function that passes the command message and arguments as parameters.
- `postCheck` - Run after checks are completed. Accepts a function that passes the command message, arguments, and the boolean result of the checks (pass/fail).
- `postExecution` - Run after the command has finished executing, but before any response has been sent. Accepts a function that passes the command message, arguments, and whether the command was executed successfully. This hook is not run if pre-execution checks were failed.
- `postCommand` - Run after the response or error message has been sent, and the command has finished processing. Accepts a function that passes the command message, arguments, and the newly created message (if applicable).

abalabahaha:
Only nit would be the use of `options.hooks` vs some sort of `registerHook()`-esque function... But that can wait until people who actually use these want to give input. I'll leave this open a bit for the critics.

Brayzure:
Can always have both options.

PR #345, opened 2017-12-20
Custom Command Requirements

Brayzure:
As requested by @aequasi , a property within a Command's requirements object, `custom`, that allows you to specify a custom requirement function. If it returns true, the command can be run, if it does not, then other requirements can still allow the command to run.


lib/command/Command.js
```diff
@@ -153,6 +154,12 @@ class Command {
 
     permissionCheck(msg) {
         var req = false;
+        if(this.requirements.custom) {
```
cryptiklemur: make sure its a function too?

PR #346, opened 2017-12-25
Fix CommandClient cooldowns for DM messages

Zoddo:
https://sentry.io/share/issue/1c5f5e1ed03a48d3bcf88b3cc83e3f43/

`msg.channel.guild` is undefined when the message is sent in DM. This throw a `Cannot read property 'id' of undefined` exception.

So, this commit check if the message was sent in a guild to not try to get the guild id on a DM channel.


Zoddo:
Discussed quickly on Discord (around 01:14 UTC)

PR #348, opened 2018-01-08
Have nsfw property default to false instead of nsfw if the nsfw prope…

macdja38:
…rty is not defined.


abalabahaha:
I'd rather change docs to reflect what Discord is sending

PR #350, opened 2018-01-17
Fix ban reasons

AnotherGenZ:
Currently eris puts the reason for audit log when banning in the query. However, api docs say to include it as a header. This PR fixes that.


abalabahaha:
Intentional behavior, nothing to fix here.

Discord uses the `X-Audit-Log-Reason` header to set the audit log reason only, but ban endpoint has a `reason` query parameter that sets the reason for **both** the ban list entry and the audit log entry. API docs are incomplete ~~as always~~.

PR #354, opened 2018-01-25
Add support for discriminator updates in presence_update events

### LINKED ISSUE #353: Discriminator updates are not handled properly
When a user updates only his discriminator Eris does not properly update that. This is caused by Eris not checking if the discriminator changed (https://github.com/abalabahaha/eris/blob/master/lib/gateway/Shard.js#L216) . Easily fixed by adding the discrim to that check.

DasWolke:
This pr adds support for discriminator updates in presence_update events, also fixes #353


TheAkio:
German bug fixing efficiency

DasWolke:
\>_<

PR #357, opened 2018-01-29
Fix for crash when getting an error from Discord's API

sylv:
Whenever Eris recieves a Discord API error, it crashes (atleast every time I've gotten one with the dev branch). This fixes it. 

First pull request so this might be a complete mess, but this keeps on annoying me so here I am.

### DIFF
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -275,7 +275,7 @@ class RequestHandler {
                             }
 
                             var stack = _stackHolder.stack;
-                            if(stack.startsWith("Error\n")) {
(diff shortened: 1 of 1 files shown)


lib/rest/RequestHandler.js
```diff
@@ -275,7 +275,7 @@ class RequestHandler {
                             }
 
                             var stack = _stackHolder.stack;
-                            if(stack.startsWith("Error\n")) {
+                            if(typeof stack === "string" && stack.startsWith("Error\n")) {
```
Brayzure, line 278: When would `stack` not be a string? As far as I can tell, there would never be a case where it's not a string.

PR #360, opened 2018-02-09
Dev

anthoeknee:
A commandclient embed since there’s only command ones, also looking to add a lot more commandclient stuff.


vitalkanev:
> Pulling from user's `dev` branch into `master`

cc repository owner

abalabahaha:
Basically a dupe of the embed + CommandClient examples. People should be able to connect those dots themselves , and I don't really want or need an example file for every possible function combination in the lib.

PR #364, opened 2018-02-24
Add Eris.VERSION

auguwu:
This pull request is telling the bot/code that Eris is `x` version (x = Current Release on NPM/package.json)


Brayzure:
Helpful I guess, but what's so challenging about just `require('./package.json').version`? It's a single line, and your solution still wouldn't be documented anywhere afaik.

PR #373, opened 2018-03-20
Typings: Renamed dynamicIconURL in User to dynamicAvatarURL

TheAkio:
By accident I noticed that the function dynamicIconURL() for User is defined in the typings but it's not really in the source code.


Pepijn98:
Actually that doesn't have to be removed, just renamed to dynamicAvatarURL. I guess it was copy pasted from somewhere else or a typo.

TheAkio:
Yep lol... Should have checked if dynamicAvatarURL existed.

PR #374, opened 2018-03-27
Image size can be any power of 2 between 16 and 2048

ghost:
(As per https://discordapp.com/developers/docs/reference#image-formatting)

Currently Eris only supports powers of 2 between 128 and 2048. This PR fixes that.


abalabahaha:
Sorry, I've decided to just remove ImageSizes entirely, since programatically checking for `2^n` feels cleaner than hand-writing all the possible sizes. Thanks for the PR though!

Issue #372, opened 2018-03-14
member.game only updates on game (or spotify) change, not asset/detail etc change.


CoalSephos:
The presenceUpdate event seems to only fire when either a Spotify rich presence updates or a user's game changes. This issue here is that it doesn't not seem to fire when attributes to member.game such as "Details" or "Assets" change. It only changes if the bot is restarted completely. 
Example: 
![discordcanary_2018-03-14_16-50-14](https://user-images.githubusercontent.com/31113295/37417369-fde0fa90-27a7-11e8-9b27-ffe797467c4c.png)
I waited to see if it was just a timing issue, but it wasn't:
![discordcanary_2018-03-14_16-52-44](https://user-images.githubusercontent.com/31113295/37417447-40d9cf7a-27a8-11e8-843e-174a8c36f172.png)

PR #376, opened 2018-04-06
Change how duplicate presences are detected

Brayzure:
The lengthy conditional was getting a bit silly and hard to maintain.


Brayzure:
Fixes #372 btw

PR #383, opened 2018-04-18
Fix disconnect() not sending close code

Runi-c:
`client.disconnect()` currently sends Discord an empty buffer instead of a close code, which causes the bot user to not be marked offline by Discord until the gateway connection times out. This fix causes the function to send a close code of `1000` - normal close, which causes Discord to immediately close the connection and mark the bot offline as expected.


abalabahaha:
Oh wrong branch

PR #386, opened 2018-05-04
Fixed overwriting commands or aliases in cmd client

MoordMachineYT:
Fixed commands possibly being overwritten instead of throwing an error (Derp and dErp both with caseInsensitive true) and the same for aliases (now throws if one of the aliases the user is trying to set already exists regardless of it being an alias or command)


abalabahaha:
Closed for #389

PR #387, opened 2018-05-09
Guild Role Update - Added mentionable to old role

motoenduroboy:
-Added mentionable to oldRole on event GUILD_ROLE_UPDATE

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -807,7 +807,8 @@ class Shard extends EventEmitter {
                         managed: role.managed,
                         name: role.name,
                         permissions: role.permissions,
-                        position: role.position
+                        position: role.position,
++                       mentionable: role.mentionable
                     };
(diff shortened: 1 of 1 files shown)


Brayzure:
Just some general commentary, but why don't we just pass the old role as-is? Why do we create another vanilla object, and copy properties over?

motoenduroboy:
Probably a good idea unless @abalabahaha has a reason behind it?

abalabahaha:
`update()` updates the roles in place so you would be getting two of the same object...
```js
var oldRole = {name: "foo"};

var newRole = oldRole; // Both pointing to the same object
(function update(role, data) { // Over-simplified update()
    role.name = data.name;
})(newRole, {name: "bar"});

emit("roleUpdate", newRole, oldRole); // ("roleUpdate", {name: "bar"}, {name: "bar"})
```

PR #393, opened 2018-05-31
Fix deprecation: Use symbol method for custom inspection

eritbh:
Setting `inspect()` on objects in order to modify the behavior of `util.insect(obj)` is deprecated and prints a warning as of Node 10. The alternative, `[util.inspect.custom]()`, has been available since node 6.

Deprecation ref: https://github.com/nodejs/node/issues/15549

This outdates the typescript defs - I don't know how to document symbol methods so I didn't touch them. If anyone else can, let me know and I'll update this PR.

### DIFF
--- a/lib/structures/Base.js
+++ b/lib/structures/Base.js
@@ -68,7 +68,7 @@ class Base {
 
 // Node 6+ util.custom.inspect symbol support - https://github.com/nodejs/node/issues/15549
 if(util.inspect.custom) {
-    Base.prototype[util.inspect.custom] = Base.prototype.inspect
+    Base.prototype[util.inspect.custom] = Base.prototype.inspect;
 }
 
 module.exports = Base;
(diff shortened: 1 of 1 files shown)


Brayzure:
I think this should be bundled in with node 6 changes, whenever we get to them. Additionally, we should probably grant users the same courtesy that node has, and deprecate this before removing it entirely.

eritbh:
What other Node 6 changes are there? Not familiar with that.

I guess that's a fair point, but I dunno if there's any case where it matters - does anyone use `#inspect` outside the `util.inspect` use?

If it does matter, that can be resolved by setting the two properties to the same value to silence the deprecation warning (the symbol version takes precedence): `Base.prototype.inspect = Base.prototype[util.inspect.custom]` would be backwards-compatible and not emit the warning on 10.x.

abalabahaha:
Agree with Brayzure, no sense breaking Node 4 compatibility for this single change. Probably would be nicer to do it the other way around - set `prototype[util.inspect.custom]` to `prototype.inspect` if the symbol exists.

PR #394, opened 2018-06-08
Use MessageContent instead of string in TS typings

retrixe:
Since Eris' CommandClient supports the CommandGenerator returning an Object, this commit edits the TypeScript typings to support a function returning one.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -469,8 +469,8 @@ declare module "eris" {
     defaultSubcommandOptions?: CommandOptions;
     hidden?: boolean;
   }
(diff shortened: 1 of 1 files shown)


retrixe:
https://github.com/abalabahaha/eris/blob/4e229d5b3204a3cce1e02762f337a1f27112332f/lib/command/Command.js#L12

https://github.com/abalabahaha/eris/blob/4e229d5b3204a3cce1e02762f337a1f27112332f/lib/command/CommandClient.js#L303

Documentation changes required here?

abalabahaha:
The JSDoc generator has a couple issues with custom types, seems fine as is

PR #398, opened 2018-06-16
Allow guild-less invites for group DMs

eritbh:
Updates the invite structure so guild is an optional property in the case of an invite being from a group DM and therefore not having a guild.

retro-ref discordapp/discord-api-docs#614


MoordMachineYT:
Invites are guild-only, when someone invites you to a DM channel, you are added straight away. It might be a good idea to get an understanding of the api first before you’re gonna submit pull requests on applications built with it.

PR #399, opened 2018-06-17
Allow return from command hooks

motoenduroboy:
This would allow for a advanced hooks

example of a good use
```javascript
bot.registerCommand('test2', (msg, args) => {
    console.log(args);
}, {
    hooks: {
        preCommand: (msg, args) => {
            if (args && args.length > 0) {
                return { args: args.join(' ').replace(/ {2,}/gmi, ' ').split(' ') };
            }
        }
    }
});
```
this would remove extra spaces but this would be set as a global hook tho not just a single command as there would be no point to that

What it does / changes
- Checks for a return
- if so waits for it, if not continue
- set values if they are present in the return

Possible suggestions:
Check for a var Continue from the return, then continue if its true?

### DIFF
--- a/lib/command/Command.js
+++ b/lib/command/Command.js
@@ -279,17 +279,29 @@ class Command {
         return true;
     }
 
-    process(args, msg) {
+    async process(args, msg) {
         var shouldDelete = this.deleteCommand && msg.channel.guild && msg.channel.permissionsOf(msg._client.user.id).has("manageMessages");
 
         if(this.hooks.preCommand) {
-            this.hooks.preCommand(msg, args);
+            if (this.hooks.preCommand(msg, args) !== undefined) {
+                const response = await this.hooks.preCommand(msg, args);
+                msg = response.msg ? response.msg : msg;
+                args = response.args ? response.args : args;    
+            } else {
+                this.hooks.preCommand(msg, args);
+            }
         }
 
         var reply;
         if(!this.permissionCheck(msg)) {
             if(this.hooks.postCheck) {
-                this.hooks.postCheck(msg, args, false);
+                if (this.hooks.postCheck(msg, args, false) !== undefined) {
+                    const response = await this.hooks.postCheck(msg, args, false);
+                    msg = response.msg ? response.msg : msg;
+                    args = response.args ? response.args : args;    
+                } else {
+                    this.hooks.postCheck(msg, args, false);
+                }
             }
 
             if(shouldDelete) {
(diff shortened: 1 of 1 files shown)


lib/command/Command.js
```diff
@@ -279,17 +279,29 @@ class Command {
...
         var shouldDelete = this.deleteCommand && msg.channel.guild && msg.channel.permissionsOf(msg._client.user.id).has("manageMessages");
 
         if(this.hooks.preCommand) {
-            this.hooks.preCommand(msg, args);
+            if (this.hooks.preCommand(msg, args) !== undefined) {
+                const response = await this.hooks.preCommand(msg, args);
+                msg = response.msg ? response.msg : msg;
+                args = response.args ? response.args : args;    
+            } else {
+                this.hooks.preCommand(msg, args);
```
eritbh: The `else` block here should be removed, it results in the hook being run twice if it doesn't return anything

lib/command/Command.js
```diff
@@ -279,17 +279,29 @@ class Command {
         return true;
     }
 
-    process(args, msg) {
+    async process(args, msg) {
         var shouldDelete = this.deleteCommand && msg.channel.guild && msg.channel.permissionsOf(msg._client.user.id).has("manageMessages");
 
         if(this.hooks.preCommand) {
-            this.hooks.preCommand(msg, args);
+            if (this.hooks.preCommand(msg, args) !== undefined) {
```
eritbh: Since this is expected to return an object and objects are always truthy, couldn't you omit the `!== undefined` here?

(all these comments apply to all the hooks, just gonna make them once on this one)

lib/command/Command.js
```diff
@@ -279,17 +279,29 @@ class Command {
...
+                const response = await this.hooks.preCommand(msg, args);
+                msg = response.msg ? response.msg : msg;
+                args = response.args ? response.args : args;
```
eritbh: These two lines could just be `msg = response.msg || msg`, etc.

lib/command/Command.js
```diff
@@ -354,15 +384,27 @@ class Command {
...
+            } else {
+                this.hooks.postCheck(msg, args, true);
+            }
         }
 
         let ret = this.execute(msg, args);
 
         if(this.hooks.postExecution) {
-            this.hooks.postExecution(msg, args, true);
+            if (this.hooks.postExecution(msg, args, true) !== undefined) {
```
eritbh: Do we really need to implement this for `postExecution` since `msg` and `args` aren't used afterwards?

lib/command/Command.js
```diff
@@ -279,17 +279,29 @@ class Command {
         return true;
     }
 
-    process(args, msg) {
+    async process(args, msg) {
```
Brayzure, line 272: One thing I would like to mention is that async/await isn't a feature that's enabled without a flag until node 7.6.x. You should make sure that what you're adding is compatible with node 4.

lib/command/Command.js
Brayzure: Unless I'm mistaken, this unnecessarily runs this hook twice. It's preferable to run it, then do something different based on the output (or use `Promise.resolve()`).

Brayzure (changes requested):
Additionally, I don't feel that hooks should be able to make changes to the content that a command sees when it's actually time to execute it.

PR #400, opened 2018-07-03
Fix inspect/toJSON methods on structures to only omit undefined values, not all falsy ones

eritbh:
Should be self-explanatory. Also:
- Fixed a typo which would prevent `editedTimestamp` from showing up in the output of `Message#toJSON()`.
- Fixed `Message#editedTimestamp` being `NaN` instead of `null` on unedited messages


LJNeon:
I wouldn't be surprised at all if `Message#editedTimestamp` wasn't a one time thing, maybe make sure the same problem doesn't occur anywhere else?

eritbh:
Quick GitHub search for `editTimestamp` doesn't show any other results, I can do a better full check tomorrow

PR #409, opened 2018-08-01
Add permission PRIORITY_SPEAKER

ghost:
[discord-api-docs #638](https://github.com/discordapp/discord-api-docs/pull/638)


lib/Constants.js
```diff
@@ -37,6 +37,7 @@ module.exports.REST_VERSION = 7;
 
 module.exports.Permissions = {
     createInstantInvite: 1,
+    voicePrioritySpeaker: 256,
```
abalabahaha: Use bitshift notation, sort it into the right place, and fix spacing alignment

abalabahaha (changes requested):

PR #411, opened 2018-08-07
Fix docs for emoji endpoints

Ratismal:
The docs say that bot accounts are unable to use the emoji create/edit/delete endpoints, but this is no longer the case. This PR updates the documentation.

Proof:
image showing an eval which adds an emote


Ratismal:
eugh it didn't rebase, i'll make another cleaner PR. sorry.

PR #10, opened 2016-07-02
Add easy access to message guild

aemino:

PR #425, opened 2018-10-10
Add "Message#guild"

auguwu:
**Added**:

* The getter: `Message#guild`
* Update the typings to have `Message#guild`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1269,6 +1269,7 @@ declare module "eris" {
     public reactions: { [s: string]: any, count: number, me: boolean };
     public prefix?: string;
     public command?: Command;
+    public guild?: Guild;
     public constructor(data: BaseData, client: Client);
     public edit(content: MessageContent): Promise<Message>;
     public pin(): Promise<void>;
--- a/lib/structures/Message.js
+++ b/lib/structures/Message.js
@@ -195,6 +195,10 @@ class Message extends Base {
 
         return (this._channelMentions = (this.content.match(/<#[0-9]+>/g) || []).map((mention) => mention.substring(2, mention.length - 1)));
     }
+    
+    get guild() {
+        return this.channel.guild;
+    }
 
     /**
     * Edit the message


Ratismal:
Considering that `Message#guild` was intentionally removed, I think that chances of this getting pulled are slim. You can implement this project-side with something like this:
```js
const { Message } = require('eris');

Object.defineProperty(Message.prototype, "guild", {
    get: function guild() {
        return this.channel.guild;
    }
});
```

ThatTonybo:
Why was it even removed in the first place?

abalabahaha:
As stated above, this _has_ been removed from the library. A similar PR (#10) has also been rejected, though I realize now I never copied the explanation at the time from Discord chat to GitHub.

The core of it, though, is that messages are directly tied to channels, but only indirectly tied to guilds through their respective channels. So removal was partially to remove lib complexity (the lib's design (wrongly) assumes cache is always complete, so more internal special cases to cover/objects to mock, especially for REST calls), and partially to make the relationship more apparent to users (`msg.channel`, _then_ `channel.guild`).

Brayzure:
>The core of it, though, is that messages are directly tied to channels, but only indirectly tied to guilds through their respective channels.

Unless I'm misunderstanding what you meant, this is no longer entirely true. `MESSAGE_CREATE` packets now explicitly specify guild ID in addition to channel ID.
```
{ type: 0,
  tts: false,
  timestamp: '2018-10-11T19:26:32.693000+00:00',
  pinned: false,
  nonce: '500026437042962432',
  mentions: [],
  mention_roles: [],
  mention_everyone: false,
  member:
   { roles: [ '367799054467596288' ],
     nick: 'Shope',
     mute: false,
     joined_at: '2018-04-15T22:15:07.584000+00:00',
     deaf: false },
  id: '500026437294751744',
  embeds: [],
  edited_timestamp: null,
  content: 'dude what hahaha link me that',
  channel_id: '218121084564799489',
  author:
   { username: 'Shope',
     id: '143068911083061249',
     discriminator: '7834',
     avatar: '02a0131145754ab162b318035eb59229' },
  attachments: [],
  guild_id: '152517096104919042' }
```

abalabahaha:
The packet does not contain the guild object, so `Message#guild` is still not always guaranteed. Plus, if the guild is cached, Eris can always recover guild ID from the `channelGuildMap`, so this doesn’t really change things.

Issue #420, opened 2018-09-18
URI malformed error is thrown when Audit-Log reasons include a percent sign


Santhosh-Annamalai:
The aforementioned error is thrown for guild related requests like `kickGuildMember` and `addGuildMemberRole` where Audit-Log reason is passed in the `X-Audit-Log-Reason` header. It doesn't occur for requests where the reason is passed as the query param like `banGuildMember` for example. It seems like the decodeURI check in [this](https://github.com/abalabahaha/eris/blob/master/lib/rest/RequestHandler.js#L87) line is causing the issue. Here is a screenshot of the error which I got when I tried to add a role to a member with `%` as the reason.

![image](https://user-images.githubusercontent.com/34129975/45713207-05bf6b00-bbac-11e8-97ab-389f0cec33c9.png)

As Discord supports adding special characters like `%` in its audit-log reason when a member is kicked manually, I would expect the same for API related requests. I tried removing that decodeURI check after which I didn't get that error and that member was successfully kicked out. So I guess that check could be removed if that's a viable solution. Or any other alternative solution would also be welcomed by me! 😄

AlekEagle:
One reason this could be caused is how it is sending the request, to my knowledge Discord is effectively one huge REST API handler, and percent signs in internet terms mean encode the next two hexadecimal numbers as x symbol, and it is returning malformed URI because it doesn't have anything to encode. (May not be correct but this is my theory.)

MoordMachineYT:
No, that's completely wrong. The reason it failed, was because the client was trying to *decode* a string. And if you were to encode a percent sign `%25` would be returned. So that's also no issue. And regardless, this issue has (likely) been fixed in #426.

AlekEagle:
Like I said, I may have been completely wrong, but thank you for assuring me that I was wrong.

PR #426, opened 2018-10-10
Change Audit Log Processes

Ratismal:
This PR does the following:
1. Change the banGuildMember request to use the `reason` option instead of `queryReason`
    - While the ban request still accepts the reason queryString, this is for backwards compatibility. Bans, like other requests, accept the `X-Audit-Log-Reason` header
2. Change the RequestHandler to remove the URI encoding/decoding, which creates problems when you use % in your audit reason. See: #420

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -1316,7 +1316,7 @@ class Client extends EventEmitter {
         }
         return this.requestHandler.request("PUT", Endpoints.GUILD_BAN(guildID, userID), true, {
             "delete-message-days": deleteMessageDays || 0,
-            queryReason: reason
+            reason
         });
     }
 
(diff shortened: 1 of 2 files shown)


abalabahaha:
The query reason has been the only way to set the ban reason (different from the audit log reason, shows up in the ban list). Unless that's changed, Eris should continue to use the query parameter, which both sets the ban list reason and overwrites the audit log reason header.

I think the `decodeURI` block was put there to temporarily patch a bug where spaces would show up as `%20` in the audit log. PR will need some testing to see if that can still be reproduced.

Ratismal:
I have tested bans with only the X-Audit-Log-Reason header, with and without URI encoding, and everything appears to be working as intended.

Screenshots:
- Without URI encoding: https://owo.whats-th.is/3a9f5c.png
- With URI encoding: https://owo.whats-th.is/c438d7.png
- Audit logs: https://owo.whats-th.is/48e4b2.gif

Edit: I was unaware that bans even had reasons separate from audit logs, considering that they don't show up in the event. my testing shows that the audit log reason doesn't show up in the ban reason. Why it's tracked twice is beyond me, but I'll revise my pr with this in consideration. sorry about that

PR #439, opened 2018-12-02
Collection Typings

auguwu:
# Fixes
- Replaces `{ id: string | number }` -> `T extends any`
  - Reason: If you put a class into Eris' collection modal, it will error this message: `Type 'x' does not satisfy the constraint '{ id: string | number; }'.
  Property 'id' is missing in type 'x' but required in type '{ id: string | number; }'.`
  - Note: The reason only works in TypeScript, not in JavaScript.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -945,7 +945,7 @@ declare module "eris" {
     public queue(func: Function): void;
   }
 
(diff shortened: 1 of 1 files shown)


abalabahaha:
Collections require objects to have `id` properties...
And `remove()` requires an object, not an ID

PR #444, opened 2018-12-28
fix maximum avatar size threshold (1024 -> 2048)

aetheryx:
[Image size can be any power of two between 16 and 2048.](https://discordapp.com/developers/docs/reference#image-formatting)


Brayzure:
You'll need to merge into the dev branch, not master.

abalabahaha:
Mind patching the other places with similar code? Search for `> 1024`, there's a couple.

PR #449, opened 2019-01-13
Member class updates

Brayzure:
Added two things, `Member.clientStatus` and `Member.activities`. Also adds the `"dnd"` status to `Member#status` docs, fixing #456.

## Member.clientStatus
The raw packet data includes `packet.d.client_status`, which is an object whose property names are the clients that user is not offline on, and whose property values are their respective statuses on those platforms.

### Examples
`{ mobile: 'dnd', desktop: 'dnd' }`
User is offline on the web client, and Do Not Disturb on mobile and desktop
`{ desktop: 'idle' }`
User is offline on mobile and web clients, and idle on the desktop client
`{}`
User is offline on all clients.

### Implementation
`Member.clientStatus` defaults all three clients to offline, then uses the data in the packet to overwrite the property values.

## Member.activities
The raw packet data also includes `packet.d.activities`, which is an array of activity objects. Since the included properties of each object is very variable, no formatting has been done for now, and no documentation of activity properties has been made. Figured we should probably discuss this to see what works best for the lib and the docs.

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -12,13 +12,13 @@ const VoiceState = require("./VoiceState");
 * @prop {String} mention A string that mentions the member
 * @prop {Guild} guild The guild the member is in
 * @prop {Number} joinedAt Timestamp of when the member joined the guild
-* @prop {String} status The member's status. Either "online", "idle", or "offline"
+* @prop {String} status The member's status. Either "online", "idle", "dnd", or "offline"
 * @prop {Object?} game The active game the member is playing
 * @prop {String} game.name The name of the active game
 * @prop {Object} clientStatus The member's per-client status
-* @prop {String} clientStatus.web The member's status on web. Either "online", "idle", or "offline". Will be "online" for bots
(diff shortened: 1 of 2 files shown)


lib/structures/Member.js
```diff
@@ -56,6 +61,8 @@ class Member extends Base {
         this.status = data.status !== undefined ? data.status : this.status || "offline";
         this.game = data.game !== undefined ? data.game : this.game || null;
         this.joinedAt = data.joined_at !== undefined ? Date.parse(data.joined_at) : this.joinedAt;
+        this.clientStatus = Object.assign({}, { web: "offline", desktop: "offline", mobile: "offline" }, data.client_status);
```
abalabahaha: What's the reason for the `{}`? Also this needs to handle client_status undefined/null, update docs accordingly.

Brayzure: Initially, it was so it wouldn't modify the object containing the defaults, but I suppose that's not necessary. Will be changed.

Is there ever a case where it's `null`? From what I can tell, it's always defined, or if not, the property just isn't there.

Issue #213, opened 2017-03-15
Webm Streams sometimes end instantly.


macdja38:
Tested this with abal the other day without finding a decent solution. Every now and then a webm result from youtube-dl / ytdl pops up that eris refuses to play and instantly skips. using the same https.get method eris uses I can pipe the stream into a file, then play that file though eris, but when fed directly the URL of the stream eris fires an end event instantly.

macdja38:
This seems to happen more with itag 250 (64kbit/s) than 249.

macdja38:
https://pvpcraft.ca/i/6lgu9.png which states the reason as the one emitted by this reason I've added in here https://github.com/macdja38/eris/blob/voice-error/lib/voice/VoiceConnection.js#L438

SinisterRectus:
Some Discordia users have reported similar occurences. Their streams stop while using youtube-dl and FFmpeg. Likewise, the audio can be fully downloaded to a file and played without issue. I'll keep an eye on this issue, but I'd sooner blame the reliability of a YouTube stream before anything else.

abalabahaha:
This problem appears to be specific to webm/opus streams though (which bypass FFmpeg entirely)... and I think Mac is using his thing that resumes the stream on ECONNRESET, so that wouldn't be an issue here

macdja38:
@abalabahaha @SinisterRectus I can confirm that downloading and playing from a file fixes it... and playing the same song over and over again will reliably skip it. the skip happens with or without my stream-resume library (none of the skips pictured were caused by econnresets, and my library handles those when they pop up).

SomeoneCares11:
i was having this problem using `npm install node-gyp`
solved after install node-gyp with -g `npm install -g node-gyp`

abalabahaha:
Interesting... because the Webm parser does not use node-gyp in any way

SomeoneCares11:
@abalabahaha i solved it in two bots with that command, try it.

macdja38:
@SomeoneCares11 I tried it and it didn't work.

Brayzure:
Are there any more instances of this behavior occurring? More than three months without any followup suggests there aren't.

abalabahaha:
Yeah, but still don't have a solid repro method

macdja38:
It definitely still happens, problem is the urls it happens on seem to be temporary and it seems to depend where they are fetched from. If I download the file using something like wget then eris can play them, but without doing that trying to stream skips.

[...]

PR #451, opened 2019-01-14
Fix for a rare FFMPEG stream closing error

linkjay:
There is an issue on some systems where an FFMPEG stream will close and stop the song at some duration throughout. 

Upon further investigation, it appears that the stdout buffer that FFMPEG reads from the network stream is reading more than is available and stops entirely. (something like that)

To fix this, bumping up the highWaterMark variable to the highest number possible will remove this issue.

It's hard to test this issue. Basically, it appears as if nothing is wrong for some people on any system. Anybody can just play what they want. Then, by pure luck, everything from there on out will stop halfway through. It's got to be some system-level thing that is going wrong, but this fixes it. 

We've tested this on multiple different bots with different libraries and they work just fine with this fix.


ghost:
Can confirm, also happens with other Discord libs like discord.js. I was debugging this for a few nights with Linkjay, and oddly enough it was only happening to me at first -- even with both of us testing on multiple machines. We even tried sending eachother our entire node_modules folders. Not sure what changed that it's happening for him now too. I could repro this on multiple machines, on multiple distros, multiple versions of node, while using old library versions, with or without ytdl-core (even while using a manual ffmpeg child_process). This only started happening about a month ago. Would be nice to figure out why this is suddenly happening, but for now this fixes it.

My understanding is that the child process should pause writing to the stdout stream once it has enough buffered to meet highWaterMark (in bytes), to allow the reading end of the stream to catch up, but it isn't for some reason. I couldn't figure out any reason why. Bumping up highWaterMark to the max allowed number (max 32bit int) seems to resolve the issue as the buffer simply won't get full anymore, and the full output of ffmpeg can be stored.

Simple code we used to test: [eristest.tar.gz](https://github.com/abalabahaha/eris/files/2754112/eristest.tar.gz)
Song should cut out **consistently** at around 2:38, right after Eminem says to "stop the beat".
I was using node v11.5.0 at the time, for reference.

abalabahaha:
Thanks for doing the investigative work here! _Possibly_ one cause of #213, along with a couple other reported WTFs.

Code-wise, comments should follow capitalization and spelling like the rest of the lib. `2^32-1` should not be hard-coded, maybe use the `Infinity` global or something.

More importantly though, nullifying `highWaterMark` has serious implications on memory management, especially for music bots with high/long-running stream concurrency. So this proposed fix cannot be merged as-is. Let's try to find a cause rather than treat a symptom.

linkjay:
I'm open to being wrong here, but I don't think that is true when it comes to the memory implications. I'm running my bot on a 2 GB RAM VPS and tested out a lot of different 10 - 30 minute videos and a 3 hour video with no memory implications. 

I and Doug have been running our bots (his on discord.js with special ffmpeg; mine with eris and hard coded highwatermark in the library) just fine with this fix. I'll continue to test and see if I can run into these memory issues that you say, but I believe this is a fix.

I will add another commit making my code consistent with the rest of the library.

Also, the integer we used in this commit is actually `2^31-1`. That number was from Doug and I assume it's just so it doesn't completely hit the limit, and it still does the job.

PR #461, opened 2019-02-09
Don't set speaking indicator before playing something

Runi-c:
Makes bots not have speaking indicators on joining a channel if they aren't playing anything yet


abalabahaha:
Thanks for the PR, but the way I want to handle this is a little more complicated due a couple oddities in the client and voice protocol

PR #467, opened 2019-02-19
d.ts: Break getOAuthApplication return into interface

eritbh:
Working on a TypeScript project that required me to reference this object's type, breaking it into its own thing is much easier to work with than `Unwrap<ReturnType<Eris.Client["getOAuthApplication"]>>`. Could be done for other object returns as well but I'm lazy.


abalabahaha:
Merged as 0fc778d

PR #469, opened 2019-03-01
1.0.0

auguwu:
I will attempt to rewrite Eris since it uses the Client a shit ton of times (check structures).


Brayzure:
In advance of this resurfacing, I highly advise opening an issue beforehand. This is a lot of work, and warrants serious discussion before you unilaterally restructure the entire library.

PR #474, opened 2019-03-08
Added support for Node HTTP(S) Agent

### LINKED ISSUE #473: Support network proxies
This library does not provide a way to set a proxy, so it's impossible to use it on networks that blocking unproxied requests, or even through a SOCKS proxy.

Moreover, it seams that [`ws` can proxy requests](https://www.npmjs.com/package/ws#how-to-connect-via-a-proxy), so it should not be complicated to implement it here.

Minigugus:
Fix #473

A very simple PR that add support for [HTTP Agents](https://nodejs.org/api/http.html#http_class_http_agent), allowing requests to be proxied by many kind of proxies (with [`http-proxy-agent`](https://www.npmjs.com/package/http-proxy-agent) or [`socks-proxy-agent`](https://www.npmjs.com/package/socks-proxy-agent) for instance).

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -130,6 +130,11 @@ class Client extends EventEmitter {
         if(defaultImageSize < 16 || defaultImageSize > 1024 || (defaultImageSize & (defaultImageSize - 1))) {
             throw new TypeError("Invalid default image size: " + defaultImageSize);
         }
+        // Set HTTP Agent on Websockets if not already set
+        if (this.options.agent && !(this.options.ws && this.options.ws.agent)) {
+            this.options.ws = this.options.ws || {};
+            this.options.ws.agent = this.options.agent;
+        }
 
         this.token = token;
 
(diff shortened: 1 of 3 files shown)


lib/Client.js
```diff
@@ -130,6 +130,11 @@ class Client extends EventEmitter {
         if(defaultImageSize < 16 || defaultImageSize > 1024 || (defaultImageSize & (defaultImageSize - 1))) {
             throw new TypeError("Invalid default image size: " + defaultImageSize);
         }
+        // Set HTTP Agent on Websockets if not already set
+        if (this.options.agent && !(this.options.ws && this.options.ws.agent)) {
+            this.options.ws = this.options.ws || {};
+            this.options.ws.agent = this.options.agent;
+        }
```
eritbh, line 139: The jsdoc comment for the client constructor should be updated to reflect this new option.

PR #482, opened 2019-04-09
Add a keepalive agent to RequestHandlers

CaramelFur:
This stops NodeJS from opening a new connection for every REST request.


abalabahaha:
Probably should be user-configurable, which #474 already does

PR #485, opened 2019-05-10
Ignore duplicate spaces between command args

noahm:
In my experience users often add multiple spaces between commands, especially after an `@mention` due to the way discord "helpfully" autocomplets a space all on its own. e.g. 
```
@botname  hello
```
Unfortunately, the current command client naively parses the commands into `["", "hello"]` and gives up when the empty string isn't a registered command.

This simply filters out any empty strings from the args to allow such messages to be processed as a user would expect.

### DIFF
--- a/lib/command/CommandClient.js
+++ b/lib/command/CommandClient.js
@@ -149,7 +149,7 @@ class CommandClient extends Client {
 
         msg.command = false;
         if((!this.commandOptions.ignoreSelf || msg.author.id !== this.user.id) && (!this.commandOptions.ignoreBots || !msg.author.bot) && (msg.prefix = this.checkPrefix(msg))) {
-            const args = msg.content.replace(/<@!/g, "<@").substring(msg.prefix.length).split(" ");
+            const args = msg.content.replace(/<@!/g, "<@").substring(msg.prefix.length).split(" ").filter((s) => !!s);
             const label = args.shift();
             const command = this.resolveCommand(label);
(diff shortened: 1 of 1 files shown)


LJNeon:
I would +1, but your implementation is lacking. `split(/\s+/g)` is up to 3x faster and always at least 25% faster, and (imo) much cleaner and better code than `filter(s => !!s)`.

noahm:
Fair enough, but leading and trailing spaces still add empty strings to the array in that case. How about `.trim().split(/\s+/g)` ?

LJNeon:
AFAIK Discord trims leading and trailing whitespace.

noahm:
It does, but we're not dealing with the entire message at this point. The command client has already stripped off the leading @mention of the bot user plus one space with the `substring` call immediately preceding the `split` call.

LJNeon:
Oh yea you're right, adding `trim` to handle that should be fine I think.

ak-s:
Not quite on topic, but how about multiline messages? Shouldn't we also split by new line `[\s\n]+`?

noahm:
Definitely an option, but I don't see any users sending multi-line messages to a bot and expecting them to work. My motivations here are purely to just remove a papercut I've personally seen happening time and time again.

abalabahaha:
Also to note that the official Discord client trims whitespace, but I've seen other things (bots, unofficial clients) send padded messages

PR #489, opened 2019-05-20
Remove user account support

reboxer:
What the title says, it should be more than enough to completely remove user accounts support.

Everyone feel free to suggest any changes.

*also removed the `var`s from the examples*


lib/structures/Message.js
```diff
@@ -108,7 +84,7 @@ class Message extends Base {
 
     update(data, client) {
         if(this.type === 3) { // (╯°□°）╯︵ ┻━┻
-            (this.channel.call || this.channel.lastCall).update(data.call);
+            (this.channel.call).update(data.call);
```
AnotherGenZ: No point in having parentheses around this if its just one value

AnotherGenZ (changes requested):

lib/gateway/Shard.js
```diff
@@ -280,44 +255,30 @@ class Shard extends EventEmitter {
...
-                            if(!flag) {
-                                this.emit("debug", new Error("VOICE_STATE_UPDATE for user leaving call not found"));
-                                break;
-                            }
-                        }
-                    } else {
-                        const channel = this.client.getChannel(packet.d.channel_id);
-                        if(!channel.call && !channel.lastCall) {
+                    const channel = this.client.getChannel(packet.d.channel_id);
+                        if(!channel.call) {
```
eritbh: `channel.call` will always be undefined because you removed that. `packet.d.guild_id === undefined` implies we're in a private channel channel; the only thing that should be inside that condition is a `return`, because we want to ignore private voice chats.

lib/gateway/Shard.js
```diff
@@ -1250,12 +1090,7 @@ class Shard extends EventEmitter {
                         this.client.token = "Bot " + this.client.token;
                     }
                 } else {
-                    this.client.bot = false;
-                    this.client.userGuildSettings = {};
-                    packet.d.user_guild_settings.forEach((guildSettings) => {
-                        this.client.userGuildSettings[guildSettings.guild_id] = guildSettings;
-                    });
-                    this.client.userSettings = packet.d.user_settings;
+                    throw "User bots are not allowed by Discord's Terms of Service, please use a bot account.";
```
eritbh: Throwing strings is bad practice, they don't contain stack traces. `throw new Error()` pls.
```suggestion
                    throw new Error("User bots are not allowed by Discord's Terms of Service, please use a bot account.");
```

lib/structures/GroupChannel.js
```diff
@@ -12,7 +12,6 @@ const User = require("./User");
 * @prop {String} id The ID of the channel
 * @prop {String} mention A string that mentions the channel
 * @prop {Call?} call The current group call, if any
-* @prop {Call?} lastCall The previous group call, if any
```
eritbh, line 15: My two cents: Bots can't join group channels, and very few bots are still in them (only bot accounts converted from user accounts way back when, which were already in group channels prior to being converted). I think it makes more sense to get rid of this class entirely, and in any case where a GroupChannel would be necessary, just no-op instead (ignore relevant events, refuse to add these channels to cache, etc.)

lib/structures/Message.js
```diff
@@ -65,36 +65,8 @@ class Message extends Base {
...
-                    data.call.id = this.id;
-                    this.channel.call = new Call(data.call, this.channel);
-                }
-                data.content = `${this.author.mention} started a call. — Join the call.`;
-            }
-        } else if(this.type === 4) {
-            data.content = `${this.author.mention} changed the channel name: ${data.content}`;
-        } else if(this.type === 5) {
-            data.content = `${this.author.mention} changed the channel icon.`;
+        else if(this.type === 1 || this.type === 2 || this.type === 3 || this.type === 4 || this.type === 5 ) { // (╯°□°）╯︵ ┻━┻
```
eritbh: this isn't really worthy of a table flip lol

lib/structures/Message.js
```diff
@@ -108,7 +80,7 @@ class Message extends Base {
 
     update(data, client) {
         if(this.type === 3) { // (╯°□°）╯︵ ┻━┻
-            (this.channel.call || this.channel.lastCall).update(data.call);
+            this.channel.call.update(data.call);
```
eritbh: as noted above, `this.channel.call` is no longer a thing

eritbh (changes requested):
meh

lib/structures/Message.js
eritbh: Wait, don't do this check from within the constructor. Remove instances of the constructor being called with these types in the first place.

reboxer: Will change this tomorrow

lib/structures/Message.js
eritbh: actually, messages will never have type 3 anymore anyway, just remove this condition entirely

lib/gateway/Shard.js
```diff
@@ -1220,7 +1047,6 @@ class Shard extends EventEmitter {
                 this.checkReady();
                 break;
             }
-            case "RESUMED":
```
eritbh: wait what we need this

eritbh (changes requested):

lib/gateway/Shard.js
reboxer: little mistake here, will be fixed in the next commit 👍

eritbh (approved):

lib/gateway/Shard.js
```diff
@@ -414,6 +372,9 @@ class Shard extends EventEmitter {
                 break;
             }
             case "MESSAGE_CREATE": {
+                if(packet.d.type <= 1 && packet.d.type >= 5) {
```
eritbh: i am the big dumb

```suggestion
                if(packet.d.type >= 1 && packet.d.type <= 5) {
```

lib/structures/Message.js
```diff
@@ -64,41 +64,14 @@ class Message extends Base {
...
-        } else if(this.type === 4) {
-            data.content = `${this.author.mention} changed the channel name: ${data.content}`;
-        } else if(this.type === 5) {
-            data.content = `${this.author.mention} changed the channel icon.`;
-        } else if(this.type === 6) {
+        else if(this.type === 6) {
             data.content = `${this.author.mention} pinned a message to this channel. See all the pins.`;
         } else if(this.type === 7) {
             data.content = SystemJoinMessages[~~(this.createdAt % SystemJoinMessages.length)].replace(/%user%/g, this.author.mention);
+        } else if(this.type === 8 || this.type === 9 || this.type === 10 || this.type === 11) {
```
Brayzure, line 73: Should not be included in the PR. Should be in a separate PR (#491 is one of them).

PR #490, opened 2019-05-23
Convert CategoryChannel#channels to a getter

eritbh:
The list of channels may change over time, so the old method isn't accurate. This method is just a shorthand for the current recommend method of getting this information.

### DIFF
--- a/lib/structures/CategoryChannel.js
+++ b/lib/structures/CategoryChannel.js
@@ -15,7 +15,7 @@ const GuildChannel = require("./GuildChannel");
 * @prop {Number} position The position of the channel
 * @prop {Boolean} nsfw Whether the channel is an NSFW channel or not
 * @prop {Collection<PermissionOverwrite>} permissionOverwrites Collection of PermissionOverwrites in this channel
-* @prop {Array<GuildChannel>} channels A collection of guild channels that are part of this category
+* @prop {Array<GuildChannel>} channels An array of guild channels that are part of this category
 */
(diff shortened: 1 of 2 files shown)


lib/structures/CategoryChannel.js
```diff
@@ -15,23 +15,18 @@ const GuildChannel = require("./GuildChannel");
 * @prop {Number} position The position of the channel
 * @prop {Boolean} nsfw Whether the channel is an NSFW channel or not
 * @prop {Collection<PermissionOverwrite>} permissionOverwrites Collection of PermissionOverwrites in this channel
-* @prop {Collection<GuildChannel>?} channels A collection of guild channels that are part of this category
+* @prop {Array<GuildChannel>} channels A collection of guild channels that are part of this category
```
TheEvilSocks: With the changed return type to Array, `A *collection* of guild channels` is no longer an accurate description.

abalabahaha:
Breaks code relying on channels being a Collection. Also turns it into a small helper getter, which makes it a bit less useful.

Also there should be code to update this Collection. Is it missing events?
- [Creating a channel](https://github.com/abalabahaha/eris/blob/247092d/lib/gateway/Shard.js#L898)
- [Adding/changing/removing the parent channel](https://github.com/abalabahaha/eris/blob/247092d/lib/gateway/Shard.js#L969-L976)

eritbh:
Just channel delete handler appears to be missing. Closing this; another PR will be made to fix that specific instance.

PR #503, opened 2019-07-17
Fix for upcoming change to Discord voice nodes

HcgRandon:
Discord is going to be deploying an update to their voice nodes. This change means that the endpoint and the IP found in Voice OPCode 2 are no longer the same.

This change now correctly uses the IP from OPCode 2. Instead of DNS resolving the IP of the endpoint.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -82,7 +82,7 @@ module.exports.VoiceOPCodes = {
     READY:               2,
     HEARTBEAT:           3,
     SESSION_DESCRIPTION: 4,
-    SPEAKING:            5,
+    SPEAKING:            5
 };
 
 module.exports.SystemJoinMessages = [
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -83,11 +83,6 @@ module.exports.VoiceOPCodes = {
     HEARTBEAT:           3,
     SESSION_DESCRIPTION: 4,
     SPEAKING:            5,
-    HEARTBEAT_ACK:       6,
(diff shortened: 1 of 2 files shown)


HcgRandon:
~~Changing the Voice OPCode Constants Might of broke stuff I forgot to check. Abal would know off hand. I need to check~~ Looks like only place that uses VoiceOPCodes is VoiceConnection

abalabahaha:
How's this doing on Ayana prod? ~~I can't remember why I was caching the DNS query, but it should have been non-trivial.~~ I lied, no need for DNS.

Also, what's with the unused Voice3 constants?

HcgRandon:
This is only tested on my dev bot ATM. I'm going to push it to Ayana PRO shortly.

As for DNS it is no longer needed here. Because it will always be an IP per Discord Dev
image

HcgRandon:
If you wish to test this change, Simply use Japan voice servers. Discord has deployed this change there. Prior to this fix Eris will fail to connect

lib/Constants.js
```diff
@@ -79,10 +79,10 @@ module.exports.Permissions = {
 module.exports.VoiceOPCodes = {
     IDENTIFY:            0,
     SELECT_PROTOCOL:     1,
-    HELLO:               2,
+    READY:               2,
     HEARTBEAT:           3,
     SESSION_DESCRIPTION: 4,
-    SPEAKING:            5
+    SPEAKING:            5,
```
abalabahaha: Trailing comma?

PR #504, opened 2019-07-18
update Member.js

attributeerror:
Added two new getters: tag and highestRole
tag will return the user's username and discriminator combined (e.g. AttributeError#2513)
highestRole will return the member's highest role in the server, or null if they don't have any roles.

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -108,7 +108,7 @@ class Member extends Base {
     }
     
     get tag() {
-        return `${this.user.username}#${this.user.discriminator}`;
+        return this.user.tag;
     }
 
     get avatar() {
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -106,6 +106,10 @@ class Member extends Base {
     get discriminator() {
         return this.user.discriminator;
     }
+    
+    get tag() {
+        return `${this.user.username}#${this.user.discriminator}`;
+    }
 
     get avatar() {
         return this.user.avatar;
@@ -137,6 +141,12 @@ class Member extends Base {
 
     get mention() {
         return `<@!${this.id}>`;
+    }
+    
+    get highestRole() {
+        return this.roles.map(rID => this.guild.roles.get(rID)).sort((a,b) => {
+           return a.position < b.position; 
+        })[0] || null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


TTtie:
Here are multiple issues with your code:

1. Your code doesn't comply with the eslint rules that are set for this repo
2. Your highestRole implementation doesn't even work
3. These methods can be easily implemented as utility methods for your bot.

If you'd like something more beginner friendly, please check out discord.js instead.

I recommend you to close this PR now, because it is probably going to be closed anyway.

Thanks,
TTtie.

attributeerror:
1) I just ran it through a checker and there were no issues, so I would like to see where you checked the code.
2) I ran multiple of my own checks, on different members, in different servers, and it worked perfectly every time, so I'd also like to see how you checked the code.
3) I'm aware, these started as utility methods but I felt like other people could find them useful - is there anything wrong with helping the community?

Last time I checked, Eris is far more beginner-friendly than Discord.js, and when I've recommended both D.JS and Eris to actual beginners, they've chosen Eris and stuck with it.

I appreciate your concern, but I'll wait for the owner of the repo to close the PR. Thank you.

knht:
Just wanted to make sure that you know that's not how actual tests work. Throwing something together beneficial to yourself and testing it on a few instances isn't covering merely enough test cases. 

Also, Eris has always striven to stick very closely to the API Discord provides. This change isn't part of it. Eris has never aimed to be a high level abstraction like Discord.js.

eritbh:
As far as the checker you ran it through, it must not have been checking eslint rules. Missing semicolon in User.js

attributeerror:
I wasn't aware that was Eris' aim - I've always seen it as a smaller library that has created way after D.JS and didn't yet have some of the more useful features, such as the two I added here.

As for the eslint checks, not sure what happened with that but I'll fix it when I get back on my computer.

Thank you both for the *constructive* criticism.

LJNeon:
For relevant discussion on why `Member#tag` won't be accepted into the lib, see #252. The same will very likely apply to `Member.highestRole`

abalabahaha:
Partial duplicate of #252. As people may have mentioned, Eris doesn't really include higher-level helpers like these. Leans closer to being simple and lower-level than to being beginner-friendly/batteries-included.

Also, some improvements: `sort()` wants `a - b` instead of `a < b`. It's also relatively inefficient compared to a `max()`-like implementation, since we only need the single highest object.

PR #505, opened 2019-07-20
splashURL/bannerURL use user given settings

bsian03:
Guild.splashURL and Guild.bannerURL both had URLs ending with .jpg and not the settings given in Client.options. Now uses settings given in Client.options (including the image size)

### DIFF
--- a/lib/structures/Guild.js
+++ b/lib/structures/Guild.js
@@ -165,7 +165,7 @@ class Guild extends Base {
     }
 
     get iconURL() {
-        return this.icon ? `${CDN_URL}/icons/${this.id}/${this.icon}.${this.icon.startsWith("a_") ? "gif" : this.shard.client.options.defaultImageFormat}?size=${this.shard.client.options.defaultImageSize}` : null;
+        return this.icon ? `${CDN_URL}/icons/${this.id}/${this.icon}.${this.icon.startsWith('a_') ? 'gif':this.shard.client.options.defaultImageFormat}?size=${this.shard.client.options.defaultImageSize}` : null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


abalabahaha:
The problem with reusing `defaultImageSize` is that the splashes and banners are usually larger than the normal square icons...

bsian03:
Should've also noted that - `defaultImageSize` takes the size of the width into account, height goes relative to the width

bsian03:
If it was excluded, width takes 128px which isn't ideal (obviously)

bsian03:
Just updated the JSDoc for it btw

abalabahaha:
A `dynamicBannerURL`/`dynamicSplashURL` (like the existing `dynamicIconURL`) would be nice here

lib/structures/Guild.js
```diff
@@ -165,7 +165,7 @@ class Guild extends Base {
     }
 
     get iconURL() {
-        return this.icon ? `${CDN_URL}/icons/${this.id}/${this.icon}.${this.icon.startsWith("a_") ? "gif" : this.shard.client.options.defaultImageFormat}?size=${this.shard.client.options.defaultImageSize}` : null;
+        return this.icon ? `${CDN_URL}/icons/${this.id}/${this.icon}.${this.icon.startsWith('a_') ? 'gif':this.shard.client.options.defaultImageFormat}?size=${this.shard.client.options.defaultImageSize}` : null;
```
eritbh: Not sure why the cosmetic changes were made here, they go against the repo's eslint (double quotes are required, not sure about ternary spacing off the top of my head though)

bsian03: I honestly don't know, I'm used to using single quotes and never considered that. I'll change it to double quotes

PR #506, opened 2019-07-20
Basic support for upcoming channel mention format

eritbh:
ref discordapp/discord-api-docs#1030

- [x] Make Message#channelMentions return channel IDs for mentions in this format (aca5075)
- [x] Update the matching in Message#cleanContent to replace this format with a channel name where possible (9f6396a)

Also fixes a bug in Message#cleanContent where only the first instance of a channel in a message would get replaced with the channel names and the rest would stay in the `<#id>` format. (9f6396a)

### DIFF
--- a/lib/structures/Message.js
+++ b/lib/structures/Message.js
@@ -197,7 +197,7 @@ class Message extends Base {
             return this._channelMentions;
         }
 
-        return (this._channelMentions = (this.content.match(/<#[0-9]+>/g) || []).map((mention) => mention.substring(2, mention.length - 1)));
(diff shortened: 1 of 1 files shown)


AnotherGenZ:
We might want to hold off on this as Discord themselves still seem uncertain about the new format.
(Based off of https://github.com/discordapp/discord-api-docs/pull/1030#issuecomment-513922868)

eritbh:
Will leave this open until we get further clarification.

abalabahaha:
This is overly optimistic for malformed mentions (matches `<#1234 oops i dropped a >`).

Anyway, Discord has axed this one, so it's no longer necessary.

PR #507, opened 2019-07-20
Code/style fixes

apacheli:
* Remove getters from Member & Role (reverted)
* Destruct properties on imports
* Simplify Base#toJSON() because the properties being checked for are from classes that don't extend Base
* Use Array#fill() on RequestHandler because it's less verbose
* Fix typos

### DIFF
--- a/lib/structures/Base.js
+++ b/lib/structures/Base.js
@@ -14,19 +14,7 @@ class Base {
     }
 
     toString() {
-        let ID = "";
-        if(this.id) {
-            ID = " " + this.id;
-        } else if(this.user) {
-            ID = " " + this.user.id;
-        } else if(this.baseObject) {
-            ID = "<" + this.baseObject.name + ">";
-        } else if(this.label) {
-            ID = " " + this.label;
-        } else if(this.code) {
-            ID = " " + this.code;
-        }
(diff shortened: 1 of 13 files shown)


Ratismal:
Removing every single getter from the member and role objects is a hugely breaking change, and in my opinion should go in it's own PR for discussion.

I don't understand all this hate for getters, anyways. they barely cost anything and offer a lot of convenience.

apacheli:
Alright, I have reverted the getter changes. I'll open another PR later.

DonovanDMC:
removing those getters would break a ton of stuff in my code, so I would rather not have that

lib/structures/Base.js
```diff
@@ -14,19 +14,7 @@ class Base {
...
-        let ID = "";
-        if(this.id) {
-            ID = " " + this.id;
-        } else if(this.user) {
-            ID = " " + this.user.id;
-        } else if(this.baseObject) {
-            ID = "<" + this.baseObject.name + ">";
-        } else if(this.label) {
-            ID = " " + this.label;
-        } else if(this.code) {
```
abalabahaha, line 26: Isn't this still used for `Invite`?

PR #508, opened 2019-07-20
using "auto" instead of string on maxShards throws ts error

DonovanDMC:
the maxShards property being number | "auto" on ClientOptions causes a typescript error to be thrown because string is not assignable to number | "auto"

[error message](https://butts-are.cool/_2360_2019-07-20_16-21-43.641.png)
[config](https://butts-are.cool/img2361_Code_-_Insiders_2019-07-20_16-27-24.827.png)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -433,7 +433,7 @@ declare module "eris" {
     guildCreateTimeout?: number;
     largeThreshold?: number;
     lastShardID?: number;
-    maxShards?: number | "auto";
+    maxShards?: number | string;
     messageLimit?: number;
     opusOnly?: boolean;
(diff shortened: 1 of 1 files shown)


eritbh:
This is a limitation of TS type inference with object literals and not a bug in Eris. For possible solutions, see: https://basarat.gitbooks.io/typescript/docs/types/literal-types.html#inference

I can help further in the Discord if you're confused about how to implement this.

DonovanDMC:
fixed via [this change in my config](https://butts-are.cool/img2365_Code_-_Insiders_2019-07-20_16-49-12.300.png)

PR #511, opened 2019-08-05
Update `defaultAvatar` calculation

zaza7:
You thought I wouldn't but I would, Abal. :^)

### DIFF
--- a/lib/structures/User.js
+++ b/lib/structures/User.js
@@ -45,7 +45,7 @@ class User extends Base {
     }
 
     get defaultAvatarURL() {
-        return `https://cdn.discordapp.com/embed/avatars/${this.defaultAvatar}.png`;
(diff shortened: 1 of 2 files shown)


lib/structures/User.js
```diff
@@ -41,11 +41,11 @@ class User extends Base {
     }
 
     get defaultAvatar() {
-        return Constants.DefaultAvatarHashes[this.discriminator % Constants.DefaultAvatarHashes.length];
+        return this.discriminator % 5;
     }
 
     get defaultAvatarURL() {
-        return `https://discordapp.com/assets/${this.defaultAvatar}.png`;
+        return `https://cdn.discordapp.com/embed/avatars/${this.defaultAvatar}.png`;
```
abalabahaha: Use `CDN_URL`, see below for examples

abalabahaha (changes requested):

PR #514, opened 2019-08-07
Adds support for channel following and messageCreate type 12.

Khaaz:
This PR adds support for the new channel follow feature according to: discordapp/discord-api-docs#1051

This was tested so it support messageCreate type 12 and adds the new `messageReference` and `flags` properties to the message object.

A MessageFlags constant is also added in the Constants file.

### DIFF
--- a/lib/structures/Message.js
+++ b/lib/structures/Message.js
@@ -21,11 +21,11 @@ const User = require("./User");
 * @prop {Number?} editedTimestamp Timestamp of latest message edit
 * @prop {Boolean} tts Whether to play the message using TTS or not
 * @prop {Boolean} mentionEveryone Whether the message mentions everyone/here or not
-* @prop {Object} messageReference An object containing the reference to the original message if it is a crossposted message
+* @prop {Object?} messageReference An object containing the reference to the original message if it is a crossposted message
 * @prop {String} messageReference.messageID The id of the original message this message was crossposted from
 * @prop {String} messageReference.channelID The id of the channel this message was crossposted from
 * @prop {String} messageReference.guildID The id of the guild this message was crossposted from
-* @prop {Object} flags Message flags (see constants) 
+* @prop {Number} flags Message flags (see constants) 
 * @prop {Object[]} attachments Array of attachments
 * @prop {Object[]} embeds Array of embeds
 * @prop {Object?} activity The activity specified in the message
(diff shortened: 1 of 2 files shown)


Khaaz:
I fixed the naming and added the documentation.
This is pretty much ready, the last question concerns the message content. I used the channel mention to try to replicate discord showing the server where the channel is from. However should we use the raw channel name here?

lib/structures/Message.js
```diff
@@ -21,6 +21,11 @@ const User = require("./User");
 * @prop {Number?} editedTimestamp Timestamp of latest message edit
 * @prop {Boolean} tts Whether to play the message using TTS or not
 * @prop {Boolean} mentionEveryone Whether the message mentions everyone/here or not
+* @prop {Object} messageReference An object containing the reference to the original message if it is a crossposted message
```
abalabahaha: Should be nullable

lib/structures/Message.js
```diff
@@ -21,6 +21,11 @@ const User = require("./User");
...
+* @prop {String} messageReference.messageID The id of the original message this message was crossposted from
+* @prop {String} messageReference.channelID The id of the channel this message was crossposted from
+* @prop {String} messageReference.guildID The id of the guild this message was crossposted from
+* @prop {Object} flags Message flags (see constants)
```
abalabahaha: Seems like a number, not object

lib/structures/Message.js
```diff
@@ -103,7 +121,10 @@ class Message extends Base {
             data.content = `${this.author.mention} just boosted the server!`;
         } else if(this.type === 9 || this.type === 10 || this.type === 11) {
             data.content = `${this.author.mention} just boosted the server! ${this.channel.guild ? this.channel.guild.name : data.guild_id} has achieved **Level ${this.type - 8}!**`;
-        }  else {
+        } else if(this.type === 12) {
+            const guild = data.content.split('#');
+            data.content = `${this.author.mention} has added ${guild.length && guild[0]}<#${this.messageReference && this.messageReference.channelID}> to this channel`;
```
abalabahaha: The channel ID is not going to resolve for many people, think it would be better to just use `data.content` here instead of the whole `guildName+channelID` thing, and keep ID accessible for bots from `messageReference`

abalabahaha (changes requested):

Khaaz:
Fixed the documentation and changed the content to the raw data.content without the channel mention.
I thought so as well, it makes more sense to just use the name.

PR #519, opened 2019-08-13
Fix Client#getGateway() return value

apacheli:
yes


exoRift:
same

apacheli:
I also decided to fix a typo that should've been fixed in my last PR, but accidentally I reverted them.

PR #516, opened 2019-08-10
Add stream perm

AnotherGenZ:
discordapp/discord-api-docs#1059

PR #520, opened 2019-08-14
Adds stream to Permissions.all, and Permissions.allVoice.

Khaaz:
Just realised that #516 added the stream permission without adding it to `Permissions.all`, `Permissions.allGuild` and `Permissions.allVoice`.
This PR addresses that.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -64,7 +64,7 @@ module.exports.Permissions = {
     manageRoles:          1 << 28,
     manageWebhooks:       1 << 29,
     manageEmojis:         1 << 30,
-    all:      0b1111111111111111111110111111111,
+    all:      0b1111111111111111111111111111111,
     allGuild: 0b1111100000010000000000010111111,
     allText:  0b0110000000001111111110001010001,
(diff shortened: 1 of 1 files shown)


lib/Constants.js
```diff
@@ -63,10 +63,10 @@ module.exports.Permissions = {
     manageRoles:          1 << 28,
     manageWebhooks:       1 << 29,
     manageEmojis:         1 << 30,
-    all:      0b1111111111101111111110111111111,
-    allGuild: 0b1111100000000000000000010111111,
+    all:      0b1111111111101111111111111111111,
+    allGuild: 0b1111100000000000000001010111111,
```
Zoddo: I don't think the stream permission should be added in `allGuild`. It's a voice-channel only permission, it doesn't have any effect guild-wide.

abalabahaha:
Along with the review above, also needs to be updated since #523 was merged first

Khaaz:
Sorry for the late update, merged and removed from guild permissions (obviously didn't fit there indeed).

PR #525, opened 2019-08-28
Added lastPinTimestamp property to TextChannel

ghost:
According to the [docs](https://abal.moe/Eris/docs/TextChannel) `TextChannel` class contains `lastPinTimestamp` property which returns `number` or `null`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1158,7 +1158,7 @@ declare module "eris" {
     public lastMessageID: string;
     public rateLimitPerUser: number;
     public messages: Collection<Message>;
-    public lastPinTimestamp: number | null;
+    public lastPinTimestamp?: number;
     public constructor(data: BaseData, guild: Guild, messageLimit: number);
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1158,6 +1158,7 @@ declare module "eris" {
     public lastMessageID: string;
     public rateLimitPerUser: number;
     public messages: Collection<Message>;
+    public lastPinTimestamp: number | null;
```
abalabahaha: Isn't the `?` suffix used for this?

ghost: I thought of it. But it seems not. Because `?` is used for optional ones. But this property always exists even if there are no pins in the channel. (returns `null` then)

abalabahaha: `?` seems to be used everywhere else in the typings for this purpose (a specific subset of "optional"), better to follow that convention for now

PR #526, opened 2019-08-29
Add the guild property onto Guild channels

resynth1943:
This adds a "guild" property onto of the channels from a Guild, which keeps these typings in line with the documentation at https://abal.moe/Eris/. I've discussed this at length with members of the Discord API channel at https://discordapp.com/channels/81384788765712384/381890266052689920/616695091935051792.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -20,7 +20,9 @@ declare module "eris" {
 
   type TextableChannel = TextChannel | PrivateChannel | GroupChannel;
   type AnyChannel = TextChannel | VoiceChannel | CategoryChannel | PrivateChannel | GroupChannel;
-  type AnyGuildChannel = TextChannel | VoiceChannel | CategoryChannel;
+  type AnyGuildChannel = (TextChannel | VoiceChannel | CategoryChannel) & {
+    guild: Guild
+  };
 
   interface CreateInviteOptions {
     maxAge?: number;


ghost:
@resynth1943 Well, there is actually a [GuildChannel](https://github.com/resynth1943/eris/blob/5542c3ee2cc22dee98f95713abc1cdb6643762fc/index.d.ts#L1120)

eritbh:
It seems like the types `AnyChannel` and `AnyGuildChannel` are obsolete synonyms for the `Channel` and `GuildChannel` types. The person that initially wrote those definitions isn't quite sure what's up there, so we need to do some testing to see if we should just remove those definitions and modify the rest of the typings to match.

(Responding to your initial comment from the Discord, since you've left the API server: Since both these types are at best equivalent, I'm inclined to say that the `guild` property not showing up for you is an issue with Atom's typings parser, not the typings here.)

resynth1943:
I would be more than happy to rewrite the Typings, too, if you need help with that @abalabahaha.

abalabahaha:
Would be nice, if the situation is as described in Geo's comment

eritbh:
Typings definitely need some consistency work, and since it seems the one largely responsible for what we have now isn't really active anymore, rewriting from scratch might be the easiest way to go at this point.

I can also help with authoring/reviewing if we do want to do a full rewrite.

resynth1943:
I'll try and start it tommorow (人*´∀｀)｡*ﾟ+

eritbh:
I've started work on this and will make a WIP PR from my fork in a bit.

resynth1943:
I'm no longer donating my time to the Open Source community.

PR #533, opened 2019-09-19
Rewrite .d.ts typings

eritbh:
Full rewrite of Typescript definitions as discussed in #526 and on Discord. Fixes #526. 

TODOs:
- Formatting (indentation, set up a typescript-eslint thingy?)
- Address TODOs in code
- actually finish the typings


index-new.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module "eris" {
```
apacheli, line 1: Make it callable, and maybe(?) use `namespace` instead of `module` since that's the preferred terminology [according to TypeScript](https://www.typescriptlang.org/docs/handbook/modules.html).

apacheli (changes requested):

DonovanDMC:
tslint exists

DonovanDMC:
For that "typescript-eslint thingy", tslint exists.

Skillz4Killz:
@DonovanDMC TSLint is being deprecated. Any rewrite of the typings should be done via typescript-eslint

DonovanDMC:
I was not aware of this, I'll make a note of that for my own projects.

eritbh:
let's be real i'm not gonna do this

PR #536, opened 2019-09-22
make error properties writable

### LINKED ISSUE #535: Warning: a promise was rejected with a non-error
Since some recent Node.js version, Node will now emit warnings if you reject a Promise with a non-`Error` type. Some things in Eris aren't rejected with errors, such as `Client#getInvite()`. I don't know of any other ones, but I'd assume there is more. Any ETA on a fix? This is currently filling up my logs, and I'd really like to stay away from `--no-warnings`.

After digging some more, I found something odd...
`(node:3317) Warning: a promise was rejected with a non-error: [object Error]`

alex-taxiera:
Fixes #535

### DIFF
--- a/lib/errors/DiscordHTTPError.js
+++ b/lib/errors/DiscordHTTPError.js
@@ -18,7 +18,7 @@ class DiscordHTTPError extends Error {
         });
 
         this.code = res.statusCode;
-        let message = `${this.name}: ${res.statusCode} ${res.statusMessage} on ${req.method} ${req.path}`;
+        let message = `${res.statusCode} ${res.statusMessage} on ${req.method} ${req.path}`;
         const errors = this.flattenErrors(response);
         if(errors.length > 0) {
             message += "\n  " + errors.join("\n  ");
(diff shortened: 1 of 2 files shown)


lib/errors/DiscordHTTPError.js
```diff
@@ -17,13 +17,19 @@ class DiscordHTTPError extends Error {
             value: response
         });
 
-        this.code = res.statusCode;
+        Object.defineProperty(this, "code", {
+            enumerable: false,
```
apacheli, line 21: Object.defineProperty() sets `enumerable` to `false` by default, so it can be removed, but it probably wouldn't matter much

apacheli (changes requested):

lib/errors/DiscordHTTPError.js
alex-taxiera: considering that people suggested I don't use `Object.defineProperty` if "all I was doing" was `value: thing` it may be best to be more declarative.

PR #538, opened 2019-09-26
refactor command permission check

### LINKED ISSUE #537: dmOnly get's ignored when requirements.userIDs is used
When `dmOnly` is set to true and `requirements.userIDs` is used, `dmOnly` get's ignored and the bot respond in guild channels too.
```
bot.registerCommand('ping', 'pong',
  {
    dmOnly: true,
    requirements: {
      userIDs: ['userId1', 'userId2', 'userId3'],
    },
  });
```

alex-taxiera:
Fixes #537

### DIFF
--- a/lib/command/Command.js
+++ b/lib/command/Command.js
@@ -173,11 +173,11 @@ class Command {
                 return false;
             }
         }
-        if(!msg.channel.guild) {
-            if (this.guildOnly) {
+        if(msg.channel.guild) {
+            if(this.dmOnly) {
                 return false;
             }
-        } else if(this.dmOnly) {
+        } else if(this.guildOnly) {
             return false;
         }
         const keys = typeof this.requirements.permissions === "function" ? Object.keys(this.requirements.permissions(msg)) : Object.keys(this.requirements.permissions);
--- a/lib/command/Command.js
+++ b/lib/command/Command.js
@@ -164,73 +164,52 @@ class Command {
     }
 
     permissionCheck(msg) {
-        let req = false;
-        if(this.requirements.custom && typeof this.requirements.custom === "function") {
-            req = true;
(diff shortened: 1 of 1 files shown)


lib/command/Command.js
```diff
@@ -164,73 +164,52 @@ class Command {
...
+            if(userIds.length > 0 && !this.requirements.userIDs.includes(msg.author.id)) {
+                return false;
             }
         }
         if(!msg.channel.guild) {
-            return !this.guildOnly && !req;
+            if (this.guildOnly) {
+                return false;
+            }
         } else if(this.dmOnly) {
```
eritbh: I would rearrange this `if` to avoid the negated condition, makes it a bit clearer:

```js
if (msg.channel.guild) {
	if (this.dmOnly) {
		return false;
	}
} else if (this.guildOnly) {
	return false;
}
```

lib/command/Command.js
```diff
@@ -164,73 +164,52 @@ class Command {
     }
 
     permissionCheck(msg) {
-        let req = false;
-        if(this.requirements.custom && typeof this.requirements.custom === "function") {
-            req = true;
-            if(this.requirements.custom(msg)) {
-                return true;
-            }
+        if(this.requirements.custom && typeof this.requirements.custom === "function" && !this.requirements.custom(msg)) {
```
eritbh, line 167: Why not add `await` to the function calls to add support for async functions? Would require a docs update. This also applies to a bunch of other things that can be functions rather than plain values. Not required, just something to consider.

eritbh (commented):
Nothing critical, just a couple things I noticed to be considered.

lib/command/Command.js
alex-taxiera: working on it

PR #542, opened 2019-10-13
Fixed inconsistencies in documentation and types

jacquesamsel:
- index.d.ts says that client.registerSubcommand returns `void`, but the documentation says it returns `Command`. Documentation is correct.


resynth1943:
The typings are being redone by Geo1088.

eritbh:
As I said in Discord, PRs to fix the existing typings should still be accepted here since I won't have the chance to finish my larger rewrite in the near future.

PR #544, opened 2019-10-15
Add permissionOverwrites option when creating guild channels

Astraeus-:
Adds the option to specify an array of overwrite objects to be applied when creating a guild channel as specified in the [Discord API Documentation](https://discordapp.com/developers/docs/resources/guild#create-guild-channel).

I'm not experienced with writing typings for TypeScript, so I did not update any of those.

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -379,7 +379,7 @@ class Client extends EventEmitter {
     * @arg {Number} [options.userLimit] The channel user limit (voice channels only)
     * @arg {Number} [options.rateLimitPerUser] The time in seconds a user has to wait before sending another message (does not affect bots or users with manageMessages/manageChannel permissions) (text channels only)
     * @arg {String?} [options.parentID] The ID of the parent channel category for this channel
-    * @arg {Array} [options.permissionOverwrites] An array containing overwrite objects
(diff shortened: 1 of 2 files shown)


lib/Client.js
```diff
@@ -379,6 +379,7 @@ class Client extends EventEmitter {
     * @arg {Number} [options.userLimit] The channel user limit (voice channels only)
     * @arg {Number} [options.rateLimitPerUser] The time in seconds a user has to wait before sending another message (does not affect bots or users with manageMessages/manageChannel permissions) (text channels only)
     * @arg {String?} [options.parentID] The ID of the parent channel category for this channel
+    * @arg {Array} [options.permissionOverwrites] An array containing overwrite objects
```
abalabahaha: Would be better to explicitly state "permission overwrite objects"

abalabahaha (changes requested):
Has this been tested? I'd expect this to fail because `PermissionOverwrite#toJSON()` omits the ID...

Astraeus-:
I tested creating a channel with permission overwrites specified as per the [Discord docs](https://discordapp.com/developers/docs/resources/channel#overwrite-object) and it worked as expected.

abalabahaha:
Ah, you meant raw overwrite objects.

PR #547, opened 2019-10-18
Custom reconnect

jacquesamsel:
- Added ClientOptions.maxReconnectAttempts
- Added ClientOptions.reconnectDelay

N.B. The reconnect number starts at 0
N.B. If connect() fails in any way, the error will be caught and the connection will retry until maxReconnectAttempts is met.

The default increases by `(n+1)^0.7 * 20000ms` where `n` is the amount of previous reconnects.
i.e.
Attempt    | Timeout
0               | 20000ms
1               | 32490ms
2               | 43153ms
3               | 52780ms
4               | 61703ms


resynth1943:
This was basically what I was going to add. Thanks!

PR #548, opened 2019-10-20
reduce memory usage of voice states

ghost:
(✿◡‿◡)

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -54,7 +54,6 @@ class Member extends Base {
         } else {
             this.user = null;
         }
-        this.voiceState = new VoiceState(data);
         this.update(data);
     }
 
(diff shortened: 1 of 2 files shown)


.eslintrc.yml
```diff
@@ -12,3 +12,4 @@ rules:
   semi: [2, "always"]
   require-atomic-updates: 1
   no-prototype-builtins: 0
+  quotes: ["error", "double"]
```
alex-taxiera: use `2` instead of `"error"`

alex-taxiera (changes requested):

lib/structures/Member.js
```diff
@@ -76,6 +83,10 @@ class Member extends Base {
         }
     }
 
+    get voiceState() {
+        return this.guild.voiceStates.get(this.id) || new VoiceState({ id: this.id });
```
abalabahaha: Most of the rest of the lib doesn't do one-line object definitions

.eslintrc.yml
```diff
@@ -12,3 +12,4 @@ rules:
   semi: [2, "always"]
   require-atomic-updates: 1
   no-prototype-builtins: 0
+  quotes: [2, "double"]
```
abalabahaha: This should have been a separate PR (or folded into #549)

abalabahaha (changes requested):

lib/structures/Member.js
```diff
@@ -76,6 +82,11 @@ class Member extends Base {
         }
     }
 
+    get voiceState() {
+        const voiceState = this.guild.voiceStates.get(this.id);
+        return voiceState ? voiceState : new VoiceState({id: this.id});
```
Skillz4Killz: How about merging these into 1 line? Would probably look a lot cleaner in my opinion.
```js
return this.guild.voiceStates.get(this.id) || new VoiceState({ id: this.id })
```

ghost: https://github.com/abalabahaha/eris/pull/548#discussion_r340955586

abalabahaha (approved):

PR #549, opened 2019-10-21
Add eslint rules

alex-taxiera:


abalabahaha (changes requested):
Pls steal double-quotes from #548

PR #551, opened 2019-10-27
channelMentions is never undefined

Skillz4Killz:
This PR fixes the typings and JSDocs stating that `message.channelMentions` can be optional/undefined.

Based on the getter this can never be undefined. It will always return an array.
image


Skillz4Killz:
Closed and moved commit into #552

PR #555, opened 2019-11-10
Prevent Member.activities to be overriden by undefined values

Khaaz:
This PR adds a test to make sure we are not overriding the `Member.activities` property with undefined value on update.
Here we will only override it if `data.activities` is defined.


Skillz4Killz:
Please if possible also add `activities` to the Member class in the typings file.

If you want to go above and beyond maybe we could fix all the typings for the member class 🙏 

Missing props in the typings are as follows:

1. activities
2. clientStatus

Typings that can be improved:
1. status

abalabahaha:
This fixes a bug, typings belong in a separate PR.

PR #556, opened 2019-11-13
typings fixes

DonovanDMC:
## Return a Message object when wait is true on executeWebhook
When setting `wait` to true, it should say that it returns a Message object, like it does. This can be accomplished via an intersection (&) to still include an interface, or directly with a property.
```typescript
function a(params: value, specificProp: specificValue): specificReturn;
function a(params: value, specificProp: otherValue): otherReturn;


function b(params: value, specificProp: Interface & { overrideProp: specificValue; }): specificReturn;
function b(params: value, specificProp: Interface & { overrideProp: otherValue; } ): otherReturn;
```
just makes it return a different type whenever that property is that specific value, ex:
```typescript
interface Payload {
	id?: string;
	wait?: boolean;
}

function j(s: string, j: string, o: Payload & { wait: true }): Promise<null>;
function j(s: string, j: string, o: Payload & { wait?: false }): Promise<void>;
function j(s: string, j: string, o: Payload): Promise<null | void> {
	return;
}

j(null, null, { wait: true }); // null
j(null, null, { wait: false }); // void
j(null, null, {}); // void
```
[true (first)](https://butts-are.cool/Code_-_Insiders_NMbyVdjHia.png)
[false (second)](https://butts-are.cool/Code_-_Insiders_PeemMiJEVS.png)
[none (third)](https://butts-are.cool/Code_-_Insiders_lXg92MjFgQ.png)
The return on the function implementation gets thrown out because of the overrides, though the return would be ideally a union of all the possible returns, at least in my opinion, i.e. `void | Message`

## add "userID" property on Guild.getBan()
`Guid.getBan()` is missing the `userID` property, effectively making it useless.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -644,7 +644,7 @@ declare namespace Eris {
       reason?: string,
     ): Promise<Webhook>;
     public executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait: true; }): Promise<Message>;
-    public executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait?: false; ): Promise<void>;
+    public executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait?: false; }): Promise<void>;
     public executeSlackWebhook(webhookID: string, token: string, options?: { wait?: boolean }): Promise<void>;
     public deleteWebhook(webhookID: string, token?: string, reason?: string): Promise<void>;
     public getGuildWebhooks(guildID: string): Promise<Webhook[]>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -643,7 +643,7 @@ declare namespace Eris {
       token?: string,
       reason?: string,
     ): Promise<Webhook>;
-    public executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait: true; }): Promise<Eris.Message>;
+    public executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait: true; }): Promise<Message>;
     public executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait?: false; ): Promise<void>;
     public executeSlackWebhook(webhookID: string, token: string, options?: { wait?: boolean }): Promise<void>;
     public deleteWebhook(webhookID: string, token?: string, reason?: string): Promise<void>;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -643,7 +643,8 @@ declare namespace Eris {
       token?: string,
       reason?: string,
     ): Promise<Webhook>;
-    public executeWebhook(webhookID: string, token: string, options: WebhookPayload): Promise<void>;
+    public executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait: true; }): Promise<Message>;
+    public executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait?: false; ): Promise<void>;
```
abalabahaha: Missing `}`?

Also, does the `wait?` mean that it'll also match an undefined `wait?

abalabahaha (changes requested):

index.d.ts
eritbh: `wait?` means the key can be missing, it can't be undefined if it's present.

Skillz4Killz:
```suggestion
    public executeWebhook(webhookID: string, token: string, options: WebhookPayload): Promise<void>;
```

The `wait: true` is useful as an overload but the base should just default back to the WebhookPayload. This will also allow users to input `undefined` and wait will be set as the `wait?: boolean` allowing it to accept true and false as well.

Skillz4Killz (changes requested):

index.d.ts
DonovanDMC: You have to include the false, else it will complain that false is not assignable to true.

DonovanDMC (commented):
I meant to put this elsewhere, and it won't let me delete it

index.d.ts
Skillz4Killz: You can leave it as is, as discussed in the discord channel keeping it strictly expressed in this way is preferred.

However, you don't really need the `?:false`. 

The `false`, `true`, and `undefined` will all be inherited from WebhookPayload. The ?: false is redundant code imo.

DonovanDMC: @Skillz4Killz You have to leave the false, else it will complain that [false is not assignable to true](https://butts-are.cool/Code_-_Insiders_KrU6Zyung1.png), as well as [the property "wait" being missing if not included](https://butts-are.cool/Code_-_Insiders_a6jmNLPR2j.png)

abalabahaha (approved):

DonovanDMC:
@Skillz4Killz was right, the `{ wait?: false }` is redundant, just need basically what was already there along with the `{ wait: true }`.

Skillz4Killz (approved):

PR #563, opened 2019-11-20
Speed up resume attempts

abalabahaha:
`RESUME` attempts are time-sensitive, and do not count towards the `IDENTIFY` rate limit. If a shard is potentially resumable (`sessionID` exists), it can bypass the reconnect queue.


ghost (approved):
works fine with 400 shards

PR #565, opened 2019-11-21
Channel.from()

apacheli:
* Add a new static method to Channel which converts based on `.type`
* Clean up Client channel methods and Guild constructor
* Add a class for store channels

### DIFF
--- a/lib/structures/Channel.js
+++ b/lib/structures/Channel.js
@@ -18,7 +18,7 @@ class Channel extends Base {
     get mention() {
         return `<#${this.id}>`;
     }
-    
+
     static from(data, extra) {
         switch(data.type) {
             case 0: {
--- a/lib/structures/Channel.js
+++ b/lib/structures/Channel.js
@@ -22,25 +22,25 @@ class Channel extends Base {
     static from(data, client) {
         switch(data.type) {
             case 0: {
-                return new TextChannel(data, client.guilds.get(data.guild_id));
+                return new TextChannel(data, client);
             }
             case 1: {
                 return new PrivateChannel(data, client);
             }
             case 2: {
-                return new VoiceChannel(data, client.guilds.get(data.guild_id));
+                return new VoiceChannel(data, client);
             }
             case 3: {
                 return new GroupChannel(data, client);
             }
             case 4: {
-                return new CategoryChannel(data, client.guilds.get(data.guild_id));
+                return new CategoryChannel(data, client);
             }
             case 5: {
-                return new NewsChannel(data, client.guilds.get(data.guild_id));
+                return new NewsChannel(data, client);
             }
             case 6: {
-                return new StoreChannel(data, client.guilds.get(data.guild_id));
+                return new StoreChannel(data, client);
             }
         }
         return new Channel(data);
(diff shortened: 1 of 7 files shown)


alex-taxiera (changes requested):
someone needs to test this
Annotation 2019-11-20 225632

lib/structures/Channel.js
```diff
@@ -19,6 +19,30 @@ class Channel extends Base {
         return `<#${this.id}>`;
     }
 
+    static from(data, extra) {
```
abalabahaha: Looks like it would be better to pass `client` consistently, then have the `GuildChannel`s get the guild as needed.

lib/structures/StoreChannel.js
```diff
@@ -0,0 +1,28 @@
...
+* @prop {Guild} guild The guild that owns the channel
+* @prop {String?} parentID The ID of the category this channel belongs to
+* @prop {String} name The name of the channel
+* @prop {Number} position The position of the channel
+* @prop {Boolean} nsfw Whether the channel is an NSFW channel or not
+* @prop {Collection<PermissionOverwrite>} permissionOverwrites Collection of PermissionOverwrites in this channel
+*/
+class StoreChannel extends GuildChannel {
+    constructor(data, guild) {
+        super(data, guild);
```
abalabahaha: Redundant

lib/structures/StoreChannel.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    update(data) {
+        super.update(data);
```
abalabahaha: Redundant

abalabahaha (changes requested):

lib/structures/Channel.js
```diff
@@ -19,6 +19,33 @@ class Channel extends Base {
         return `<#${this.id}>`;
     }
 
+    static from(data, client) {
+        switch(data.type) {
+            case 0: {
+                return new TextChannel(data, client.guilds.get(data.guild_id));
```
abalabahaha: Also nitpick, but passing client and fetching the guild in GuildChannel would remove the need to do this every time

lib/structures/StoreChannel.js
apacheli: They are there for easier editing in the future.

abalabahaha: Add em in in the future then

lib/structures/TextChannel.js
```diff
@@ -23,12 +23,9 @@ const Message = require("./Message");
...
 class TextChannel extends GuildChannel {
-    constructor(data, guild, messageLimit) {
-        super(data, guild);
-        if(messageLimit == null && guild) {
-            messageLimit = guild.shard.client.options.messageLimit;
-        }
-        this.messages = new Collection(Message, messageLimit);
+    constructor(data, client, messageLimit) {
+        super(data, client);
+        this.messages = new Collection(Message, messageLimit || client.options.messageLimit);
```
abalabahaha: Needs to permit `messageLimit = 0`, use a `null` check here

abalabahaha (changes requested):

abalabahaha (approved):

PR #567, opened 2019-11-23
Document flags for editMessage

Khaaz:
As documented in discordapp/discord-api-docs#1003 we can now pass flags in editMessage to set/unset SUPRESS_EMBED flags.
I simply documented this change since we can directly use content in the `editMessage` method to apply the flags to the message (`content.flags`).
The `editMessage` wasalso edited to allow passing only `content.flags`
*This PR also updates typings*

Note on `editMessage`:
- you can now edit a message by only passing flags
- editing a message without flags will let the current flags (as expected)
- editing a message with only `content.content` will just edit the content of the embed without affecting the embed (as expected)
- editing a message with only `content.embed` will just edit the embed without affecting the content (as expected)
- editing the embed of a message with the SUPRESS_EMBED flag will actually edit the embed and suppress it. Unsetting SUPPRESS_EMBED will then just show the embed again.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -495,7 +495,7 @@ declare namespace Eris {
     [key: string]: {};
   }
 
-  type MessageContent = string | { content?: string; tts?: boolean; disableEveryone?: boolean; embed?: EmbedOptions };
+  type MessageContent = string | { content?: string; tts?: boolean; disableEveryone?: boolean; embed?: EmbedOptions; flags?: number };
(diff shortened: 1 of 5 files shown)


abalabahaha (commented):
Would be nice to add a [link to the flags](https://discordapp.com/developers/docs/resources/channel#message-object-message-flags)

abalabahaha (approved):

PR #572, opened 2019-11-24
Use Channel.from() in channel event handlers

Khaaz:
This PR changes the implementation of `CHANNEL_CREATE` and `CHANNEL_UPDATE` to works with changes in #565 .
Since the guild is resolved inside the channel constructor, I first instantiate the Channel and then check for the guild.

This change was tested and it works / actually fix the bug. However please double check that everything is fine.
I am also concerned about the instanceof check toknowif the Guild is valid (cached) or not. Maybe this can be replaced with checking specifically for one of the property always included in `Guild` (eg: `guild.name`)

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -874,13 +874,13 @@ class Shard extends EventEmitter {
                     }
                     let channel;
                     if(packet.d.type === 2) {
-                        channel = guild.channels.add(new VoiceChannel(packet.d, guild), guild);
+                        channel = guild.channels.add(new VoiceChannel(packet.d, this.client), guild);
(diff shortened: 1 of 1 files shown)


Khaaz:
Switched that to actually use `Channel.from` which makes more sense overall.
This also entirely add support to the new StoreChannel added in #565 .

In CHANNEL_CREATE, with `Channel.from` it means that if channel.type is undefined or not one of known type, it will create a `Channel`. 
That would mean it doesn't have a guild property so we then manually try to add the Guild.
In case of uncached Guild we don't add the Channel to the cache.

abalabahaha:
Implemented alternatively in b0c7a07

PR #576, opened 2019-11-26
Promisified Guild#fetchAllMembers

LJNeon:
I did a basic test to make sure this works, more testing may or may not be useful. I used the `requestTimeout` client option for this, as I don't believe that making a new option is worth it here and it seemed to be the best fitting of all existing options. I also never reject the promise as the point here was to mirror previous behavior, only with a Promise that resolves at the end. I'm open to feedback on those two points.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -146,8 +146,9 @@ class Shard extends EventEmitter {
         this.preReady = false;
         if(this.requestMembersPromise !== undefined) {
             for(const guildID in this.requestMembersPromise) {
-                if(!this.requestMembersPromise.hasOwnProperty(guildID))
+                if(!this.requestMembersPromise.hasOwnProperty(guildID)) {
                     continue;
+                }
                 clearTimeout(this.requestMembersPromise[guildID].timeout);
                 this.requestMembersPromise[guildID].res();
             }
@@ -1510,8 +1511,8 @@ class Shard extends EventEmitter {
         } else if(this.ready) {
             this.requestGuildMembers([guildID]);
 
-            return new Promise(res => this.requestMembersPromise[guildID] = {
-                res,
+            return new Promise((res) => this.requestMembersPromise[guildID] = {
+                res: res,
                 timeout: setTimeout(() => {
                     delete this.requestMembersPromise[guildID];
                     res();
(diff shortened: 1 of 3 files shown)


lib/gateway/Shard.js
```diff
@@ -144,6 +144,15 @@ class Shard extends EventEmitter {
         this.connecting = false;
         this.ready = false;
         this.preReady = false;
+        if(this.requestMembersPromise !== undefined) {
+            for(const guildID in this.requestMembersPromise) {
+                if(!this.requestMembersPromise.hasOwnProperty(guildID))
```
apacheli: Missing curly

apacheli (changes requested):

lib/gateway/Shard.js
LJNeon: Fixed

lib/gateway/Shard.js
```diff
@@ -144,6 +144,16 @@ class Shard extends EventEmitter {
         this.connecting = false;
         this.ready = false;
         this.preReady = false;
+        if(this.requestMembersPromise !== undefined) {
+            for(const guildID in this.requestMembersPromise) {
+                if(!this.requestMembersPromise.hasOwnProperty(guildID)) {
+                    continue;
+                }
+                clearTimeout(this.requestMembersPromise[guildID].timeout);
+                this.requestMembersPromise[guildID].res();
```
abalabahaha: `.received` should be returned here too

lib/structures/Guild.js
```diff
@@ -149,9 +149,11 @@ class Guild extends Base {
 
     /**
     * Request all guild members from Discord
+    * @arg {Number} [timeout] The number of milliseconds to wait before resolving early. Defaults to the `requestTimeout` client option
+    * @returns {Promise<Number>} Resolves with the total amount of fetched members.
```
abalabahaha: Nit: `total number of`

abalabahaha (changes requested):

lib/structures/Guild.js
LJNeon: should be fixed

lib/gateway/Shard.js
LJNeon: should be fixed

abalabahaha (approved):

PR #577, opened 2019-11-28
Intellisense Support

hiimjasmine00:
This will allow intellisense support. UGH!!! was always wondering why it wouldn't autocomplete...


LJNeon:
Please change your PR to merge into dev instead of master.

Skillz4Killz:
I am not sure if this change has any impact. Please someone with a JS bot, should test this. I thought by default, typings are read from the root file.

EDIT: Confirmed by Donovan as well on a new project in JS the intellisense worked.

DonovanDMC:
This makes no difference with my testing.
image

By default, a `index.d.ts` is looked for.

abalabahaha (approved):

abalabahaha:
I would love to find out what causes TypeScript to do this, but I also don't care enough to... Oh well, seems backwards compatible.

PR #578, opened 2019-11-28
executeWebhook returns a bare object, not a Message object.

DonovanDMC:
the executeWebhook method just directly returns what Discord gives to us, so it needs to be made into a Message object.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -740,6 +740,32 @@ declare namespace Eris {
     (event: "resume", listener: () => void): T;
   }
 
+  interface WebhookResponse {
+    id: string;
+    type: number;
+    content: string;
+    channel_id: string;
+    author: {
+      bot?: true;
+      id: string;
+      username: string;
+      avatar?: string;
+      discriminator: string;
+    };
+    attachments: Attachment[];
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -873,7 +873,7 @@ declare namespace Eris {
       reason?: string
     ): Promise<Webhook>;
     executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait: true }): Promise<Message>;
-    executeWebhook(webhookID: string, token: string, options: WebhookPayload): Promise<void>;
+    executeWebhook(webhookID: string, token: string, options: WebhookPayload): Promise<any>;
```
alex-taxiera: should this be `Promise<Message | any>`?
JSDocs look like that.

alex-taxiera (changes requested):

index.d.ts
DonovanDMC: No, it returns something different based on the wait parameter

alex-taxiera (dismissed):

abalabahaha (approved):

PR #579, opened 2019-11-29
Fix guildCreate missing event and prop

zapteryx:
seems to have been removed on accident in a commit on October 18th 🤔


abalabahaha:
Deprecated by #580

PR #580, opened 2019-11-29
Fix guildCreate docs

zapteryx:
essentially, it appears that it was accidentally removed in a commit on October 18th


abalabahaha (approved):
Oops

PR #582, opened 2019-11-30
feat: add commandNotFound event

ItsANameToo:
This PR adds a new event to the `CommandClient` called `commandNotFound`. It gets triggered when the CommandClient could not resolve the user's message to any of the existing commands. This offers the ability for developers to hook into this event to show a response to their users that it could not find a matching command. An example would be:

```
bot.on("commandNotFound", async (message) => {
  // can also inspect the message itself here as that gets passed along
  await message.channel.createMessage(`Sorry, that command does not exist`);
})
```

Unless I missed something, I could not find an existing way to have the bot handle non-existing commands when using the `CommandClient`.

### DIFF
--- a/lib/command/CommandClient.js
+++ b/lib/command/CommandClient.js
@@ -187,6 +187,8 @@ class CommandClient extends Client {
                     }
                 }
             }
+        } else {
+            this.emit("commandNotFound", msg);
         }
     }
 


alex-taxiera (changes requested):
needs typings and jsdocs

Khaaz:
Also would be emitted on every message insteaf of "on every message that starts with the prefix but that are not an actual command"

ItsANameToo:
@Khaazz message prefix is checked before on this line: https://github.com/abalabahaha/eris/blob/master/lib/command/CommandClient.js#L143, so it should only be emitted on messages with the correct prefix but that call a non-existing command

lib/command/CommandClient.js
```diff
@@ -187,6 +187,8 @@ class CommandClient extends Client {
                     }
                 }
             }
+        } else {
```
abalabahaha, line 190: Yeah, this should be on the previous line if you only wait it for valid prefixes. Currently this runs if the prefix does not match too.

abalabahaha (commented):
Personally I'm not a fan of command not found messages. It gets spammy and essentially hogs the entire command range for a prefix. Thanks for the PR though.

PR #583, opened 2019-11-30
Add support for self_stream field on voice state

Skillz4Killz:
This PR implements support for self_stream field on voice state, otherwise known as Go Live

- [x] Fully tested
- [x] Typings Updated
- [x] JSDocs updated

image

Before user is online
image

Once user is online
image

Once user is offline or leaves the channel
image

### DIFF
--- a/lib/structures/VoiceState.js
+++ b/lib/structures/VoiceState.js
@@ -32,7 +32,7 @@ class VoiceState extends Base {
         this.suppress = data.suppress !== undefined ? data.suppress : this.suppress || false; // Bots ignore this
         this.selfMute = data.self_mute !== undefined ? data.self_mute : this.selfMute || false;
         this.selfDeaf = data.self_deaf !== undefined ? data.self_deaf : this.selfDeaf || false;
-        this.selfStream = data.self_stream !== undefined ? data.self_stream : this.selfStream || false;
+        this.selfStream = data.self_stream || false;
     }
 
     toJSON(props = []) {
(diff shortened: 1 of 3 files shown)


lib/structures/VoiceState.js
```diff
@@ -31,6 +32,7 @@ class VoiceState extends Base {
         this.suppress = data.suppress !== undefined ? data.suppress : this.suppress || false; // Bots ignore this
         this.selfMute = data.self_mute !== undefined ? data.self_mute : this.selfMute || false;
         this.selfDeaf = data.self_deaf !== undefined ? data.self_deaf : this.selfDeaf || false;
+        this.selfStream = data.self_stream || false;
```
abalabahaha, line 35: Copy the undefined check from above

abalabahaha (changes requested):

lib/structures/VoiceState.js
Skillz4Killz: @abalabahaha Sadly the way that discord handles the `self_stream` property is different from the others. Initially, I used the undefined check above it, and during testing discovered that it was broken. After much debugging, I realized that discord doesn't pass the prop when the user goes offline so it was never becoming false.

```js
this.selfStream = data.self_stream !== undefined ? data.self_stream : this.selfStream || false;
```

So basically discord does NOT pass in the `self_stream` when the user is offline. Aka it will be undefined. When it is `undefined` it resolves to `this.selfStream` (which is currently true) `|| false`. So it always remained true

Once they become true it stayed true, so I had to remove the undefined check
if discord passes that value they just went live. If they don't pass it the user went offline

abalabahaha (approved):

PR #586, opened 2019-12-02
Type fixes

DonovanDMC:
The `OldChannel` object, which is used in the [`channelUpdate`](https://github.com/abalabahaha/eris/blob/dev/index.d.ts#L665) event is missing the `nsfw`, and `type` properties, which are present in the [jsdoc](https://abal.moe/Eris/docs/Client#event-channelUpdate).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -79,6 +79,8 @@ declare namespace Eris {
     type: 1 | 2 | 4 | 5 | 6;
     bitrate?: number;
     permissionOverwrites: Collection<PermissionOverwrite>;
+    parentID?: string;
+    rateLimitPerUser?: number;
   }
 
   type FriendSuggestionReasons = { type: number; platform_type: string; name: string }[];
--- a/index.d.ts
+++ b/index.d.ts
@@ -663,7 +663,7 @@ declare namespace Eris {
       event: "channelRecipientAdd" | "channelRecipientRemove",
       listener: (channel: GroupChannel, user: User) => void
     ): T;
-    (event: "channelUpdate", listener: (channel: AnyChannel, oldChannel: GenericOldChannel & ({ type: 0; topic?: string; rateLimitPerUser: number; } | { type: 2; bitrate: number; } | { type: 5; topic?: string; rateLimitPerUser: 0 } | { type: 4 | 6; })) => void): T;
+    (event: "channelUpdate", listener: (channel: AnyChannel, oldChannel: GenericOldChannel & ({ type: 0; topic?: string; rateLimitPerUser: number } | { type: 2; bitrate: number } | { type: 5; topic?: string; rateLimitPerUser: 0 } | { type: 4 | 6 })) => void): T;
     (event: "friendSuggestionCreate", listener: (user: User, reasons: FriendSuggestionReasons) => void): T;
     (event: "friendSuggestionDelete", listener: (user: User) => void): T;
     (event: "guildAvailable" | "guildBanAdd" | "guildBanRemove", listener: (guild: Guild, user: User) => void): T;
--- a/index.d.ts
+++ b/index.d.ts
@@ -486,22 +486,22 @@ declare namespace Eris {
     banner?: string;
   }
   interface OldGuild {
-    name: string,
-    verificationLevel: 0 | 1 | 2 | 3 | 4,
-    splash?: string,
-    banner?: string,
-    region: string,
-    ownerID: string,
-    icon: string,
-    features: string[],
+    name: string;
+    verificationLevel: 0 | 1 | 2 | 3 | 4;
+    splash?: string;
+    banner?: string;
+    region: string;
+    ownerID: string;
+    icon: string;
+    features: string[];
     emojis: Emoji[],
-    afkChannelID?: string,
-    afkTimeout: number,
-    mfaLevel: 0 | 1,
-    large: boolean,
-    maxPresences?: number,
-    explicitContentFilter: 0 | 1 | 2,
-    systemChannelID?: string
+    afkChannelID?: string;
+    afkTimeout: number;
+    mfaLevel: 0 | 1;
+    large: boolean;
+    maxPresences?: number;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -74,7 +74,9 @@ declare namespace Eris {
   interface OldChannel {
     name: string;
     position: string;
+    nsfw: boolean;
     topic?: string;
+    type: 1 | 2 | 4 | 5 | 6;
     bitrate?: number;
     permissionOverwrites: Collection<PermissionOverwrite>;
```
```suggestion
    permissionOverwrites: Collection<PermissionOverwrite>;
    parentID?: string;
    rateLimitPerUser?: number;
```

Skillz4Killz (changes requested):
More missing types from oldChannel

DonovanDMC:
if we really wanted to, we could get really specific with the `type` parameter so only the properties that would be on the specific channel are there, though that would probably be done in the event definition instead.

Skillz4Killz:
@DonovanDMC Silly question, but how does one even change the channel type? Is that even possible? I didn't know discord supported that.

DonovanDMC:
> @DonovanDMC Silly question, but how does one even change the channel type? Is that even possible? I didn't know discord supported that.

I don't think you can change the type of channels.

Skillz4Killz:
Then i think the best way to handle this would be to remove the `type` property from OldMessage and the typings. `type` can never change.

index.d.ts
```diff
@@ -662,7 +662,7 @@ declare namespace Eris {
       event: "channelRecipientAdd" | "channelRecipientRemove",
       listener: (channel: GroupChannel, user: User) => void
     ): T;
-    (event: "channelUpdate", listener: (channel: AnyChannel, oldChannel: OldChannel) => void): T;
+    (event: "channelUpdate", listener: (channel: AnyChannel, oldChannel: GenericOldChannel & ({ type: 0; topic?: string; rateLimitPerUser: number } | { type: 2; bitrate: number } | { type: 5; topic?: string; rateLimitPerUser: 0 } | { type: 4 | 6 })) => void): T;
```
eritbh: Wouldn't this be cleaner if we split it up into multiple named interfaces that all inherit from `GenericOldChannel`?

eritbh (changes requested):

index.d.ts
DonovanDMC: We can do that, and add ~4-5 more interfaces, it would be cleaner, yes.

Zoddo:
A text channel can be converted to a news channel and vice versa.

DonovanDMC:
> A text channel can be converted to a news channel and vice versa.

True, forgot about that.

Skillz4Killz:
Ahh very good point so we do need the `type` property on the oldChannel incase the text-news types change.

DonovanDMC:
Decided to convert the types back to just `OldChannel`, because it's much cleaner, and a lot less of a mess since nothing else is working. As well made some changes to `guildRoleUpdate`, as it was missing properties, and they were all optional since they were using `RoleOptions`.

index.d.ts
```diff
@@ -74,9 +74,12 @@ declare namespace Eris {
...
     name: string;
     position: string;
-    topic?: string;
-    bitrate?: number;
+    nsfw: boolean;
     permissionOverwrites: Collection<PermissionOverwrite>;
+    parentID?: string;
+    topic?: string;
+    rateLimitPerUser?: number;
+    type: 0 | 2 | 4 | 5 | 6;
```
Skillz4Killz, line 82:
```suggestion
    type: 0 | 2 | 4 | 5 | 6;
    bitrate?: number;
```

index.d.ts
```diff
@@ -487,7 +490,19 @@ declare namespace Eris {
     color?: number;
     hoist?: boolean;
     mentionable?: boolean;
+    managed?: boolean;
```
Skillz4Killz: I don't believe this is a valid option to pass.

index.d.ts
```diff
@@ -662,7 +677,7 @@ declare namespace Eris {
       event: "channelRecipientAdd" | "channelRecipientRemove",
       listener: (channel: GroupChannel, user: User) => void
     ): T;
-    (event: "channelUpdate", listener: (channel: AnyChannel, oldChannel: OldChannel) => void): T;
+    (event: "channelUpdate", listener: (channel: AnyGuildChannel, oldChannel: OldChannel) => void): T;
```
Skillz4Killz: Although I agree, with this, it would basically be a breaking change for any GroupChannel bots.

Skillz4Killz (changes requested):

index.d.ts
DonovanDMC: @Skillz4Killz to make that work, OldChannel would have to be changed a lot, it is completely wrong for Group DM bots.

index.d.ts
DonovanDMC: Not a valid option? It might not be able to be changed, but it is definitely sent
image

index.d.ts
DonovanDMC: Ohh

index.d.ts
DonovanDMC: Oh, I see what you mean, I'm not 100% sure, let me go test things.

Skillz4Killz: What you are editing is RoleOptions which is used for creating roles. Bots can't create managed roles so passing a managed thing is not a thing.

index.d.ts
Skillz4Killz: I agree I don't think it is worth it to support GroupChannels so I agree this should be changed. I just think when merging this should be labelled as breaking.

index.d.ts
DonovanDMC: Doesn't make a difference, so basically isn't a valid option
image

Skillz4Killz (dismissed):

abalabahaha (approved):

PR #587, opened 2019-12-02
Allow embed timestamp to be Date in typings

oathompsonjones:
Type should be `string | Date` in order to allow for doing `timestamp: new Date()`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -316,7 +316,7 @@ declare namespace Eris {
     title?: string;
     description?: string;
     url?: string;
-    timestamp?: Date;
+    timestamp?: string | Date;
     color?: number;
     footer?: { text: string; icon_url?: string; proxy_icon_url?: string };
     image?: { url?: string; proxy_url?: string; height?: number; width?: number };
--- a/index.d.ts
+++ b/index.d.ts
@@ -316,7 +316,7 @@ declare namespace Eris {
     title?: string;
     description?: string;
     url?: string;
-    timestamp?: string | Date;
+    timestamp?: Date | string;
     color?: number;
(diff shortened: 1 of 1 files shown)


Skillz4Killz (dismissed):

zneix (dismissed):

Skillz4Killz (dismissed):

Skillz4Killz:
I think this PR is fine, but fair warning this could be considered a breaking change. Since it was accepting `string` before, it would not accept the `Date` option so most devs probably opted to do something like what I have

`new Date(time).toISOString()`

Im pretty sure this would break on compile time because you cant set a string to a `Date`. Will need testing when im able to do so

Skillz4Killz:
Please, can we consider rewinding this back to the original commit of `Date | string`? I am certain I am not the only dev who was probably sending strings instead of Date. Here is how many errors I got from this change. 

image

I believe the best type would be `Date | string`

index.d.ts
```diff
@@ -316,7 +316,7 @@ declare namespace Eris {
     title?: string;
     description?: string;
     url?: string;
-    timestamp?: string;
+    timestamp?: string | Date;
```
```suggestion
    timestamp?: Date | string;
```

Skillz4Killz (dismissed):
It's perfect. Slight modification to `Date | string` to have a Strict => Loose typings format would be nicer on the eyes imo.

DonovanDMC:
I have always used `new Date().toISOString()`, making it take only a `Date` object would break literally every single one of my projects, thousands of files total (yes, I see that it has been changed back)

Skillz4Killz (approved):

DonovanDMC (approved):

abalabahaha (approved):

PR #588, opened 2019-12-03
Fix Client#createGuild() bug where shard was undefined

apacheli:
This fixes 2 issues that were recently discovered in the Eris channel.

1. The first issue was that it was a race condition between `Guild#shard` and `Client#guildCreate`.
2. The second issue comes from the first one. The guild obtained from REST was not cached.

### DIFF
--- a/lib/structures/Guild.js
+++ b/lib/structures/Guild.js
@@ -122,29 +122,75 @@ class Guild extends Base {
     }
 
     update(data) {
-        if(data.name !== undefined) this.name = data.name;
-        if(data.verification_level !== undefined) this.verificationLevel = data.verification_level;
-        if(data.splash !== undefined) this.splash = data.splash;
-        if(data.banner !== undefined) this.banner = data.banner;
-        if(data.region !== undefined) this.region = data.region;
-        if(data.owner_id !== undefined) this.ownerID = data.owner_id;
-        if(data.icon !== undefined) this.icon = data.icon;
-        if(data.features !== undefined) this.features = data.features;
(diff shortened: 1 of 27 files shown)


Skillz4Killz:
Why not have a property dedicated to the client? `this.client = client;`. You'd never have bugs like this again and users don't need to go `guild.shard.client`. One way or the other, `client` exists in the class.

Note: This needs typings updated to reflect the changes in the PR.

lib/structures/Guild.js
```diff
@@ -50,7 +50,10 @@ const VoiceState = require("./VoiceState");
 class Guild extends Base {
     constructor(data, client) {
         super(data.id);
-        this.shard = client.shards.get(client.guildShardMap[this.id]);
+        this._client = client;
+        this.shard = client.shards.get(client.guildShardMap[this.id]) || {
```
abalabahaha: `||` is kind of unnecessary with `_client`

lib/structures/GuildChannel.js
```diff
@@ -23,6 +23,9 @@ class GuildChannel extends Channel {
     constructor(data, client) {
         super(data);
         this.guild = client.guilds.get(data.guild_id) || {
+            shard: {
```
abalabahaha: Could use `guild._client` here and this wouldn't be necessary either.

abalabahaha (changes requested):

apacheli:
Sorry for delaying so long. I completely forgot about this. Happy late holidays lol

abalabahaha (approved):

PR #590, opened 2019-12-04
fix voice channel leave docs/typings

Skillz4Killz:
This PR always sends a valid `member`in `voiceChannelLeave` event. 
It also removes Category and Text Channels from the channel type in jsdocs
It also makes Collection.find() allowed to be undefined when there is nothing found.

image

image


apacheli (changes requested):
Should probably return the raw member data from the gateway instead of `null`

abalabahaha (approved):

PR #595, opened 2019-12-11
Addition and improvements to Collection

Khaaz:
This PR adds several methods in Collection in order to ease Collection usage.

Allow to create Collection with an iterable.
`.from()` to return a Collection from an Array.
`.toArray()` to return an Array from the Collection.
`.toObject()` to return an Object from the Collection.
`.apply()` to run any Collection method but return a Collection.

Not sure exactly about the typings, would like review!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1156,11 +1156,11 @@ declare namespace Eris {
   export class Collection<T extends { id: string | number }> extends Map<string | number, T> {
     baseObject: new (...args: any[]) => T;
     limit?: number;
-    constructor(baseObject: new (...args: any[]) => T, limit?: number, iterable?: object | any[]);
+    constructor(baseObject: new (...args: T[]) => T, limit?: number, iterable?: object | T[]);
     static from(array: any[], key: string, limit: number): Collection<any>;
-    toArray(): any[];
+    toArray(): T[];
     toObject(): object;
-    apply(key: string, func: string, ...args: any[]): Collection<any>;
+    apply(key: string, func: string, ...args: any[]): Collection<T>;
     add(obj: T, extra?: any, replace?: boolean): T;
     find(func: (i: T) => boolean): T | undefined;
     random(): T;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1156,7 +1156,11 @@ declare namespace Eris {
   export class Collection<T extends { id: string | number }> extends Map<string | number, T> {
     baseObject: new (...args: any[]) => T;
     limit?: number;
-    constructor(baseObject: new (...args: any[]) => T, limit?: number);
+    constructor(baseObject: new (...args: any[]) => T, limit?: number, iterable?: object | any[]);
+    static from(array: any[], key: string): Collection<any>;
+    toArray(): any[];
+    toObject(): object;
+    apply(key: string, func: string, ...args: any[]): Collection<any>;
     add(obj: T, extra?: any, replace?: boolean): T;
     find(func: (i: T) => boolean): T | undefined;
     random(): T;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1157,7 +1157,7 @@ declare namespace Eris {
     baseObject: new (...args: any[]) => T;
     limit?: number;
     constructor(baseObject: new (...args: any[]) => T, limit?: number, iterable?: object | any[]);
-    static from(array: any[], key: string): Collection<any>;
+    static from(array: any[], key: string, limit: number): Collection<any>;
     toArray(): any[];
     toObject(): object;
     apply(key: string, func: string, ...args: any[]): Collection<any>;
--- a/lib/util/Collection.js
+++ b/lib/util/Collection.js
@@ -10,12 +10,68 @@ class Collection extends Map {
     /**
     * Construct a Collection
     * @arg {Class} baseObject The base class for all items
-    * @arg {Number} [limit] Max number of items to hold
+    * @arg {Number} [limit=null] Max number of items to hold
+    * @arg {Object} [iterable=null] An iterable to initialise the Collection
     */
-    constructor(baseObject, limit) {
-        super();
+    constructor(baseObject, limit= null, iterable = null) {
+        if(iterable && iterable instanceof Array) {
+            super(iterable);
+        } else if(iterable && iterable instanceof Object) {
+            super(Object.entries(iterable) );
+        } else {
+            super();
+        }
+
         this.baseObject = baseObject;
         this.limit = limit;
+    }
+
+    /**
+     * Creates a collection from an array. Uses the first element instances as baseObject
+     *
+     * @static
+     * @param {Array<Class>} array The array of object
+     * @param {String} key The property to use as key
+     * @returns {Collection<Class>} A newly created Collection
+     */
+    static from(array, key) {
+        return new Collection( array[0].constructor, null, array.map((e) => [e[key], e] ) );
+    }
+
+    /**
+     * Return the Collection as an Array
+     * [value, value]
+     *
+     * @returns {Array<Class>} The Collection as an Array
+     */
+    toArray() {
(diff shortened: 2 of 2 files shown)


LJNeon:
This PR is coming from the POV that as a Collection extends a Map using it as a Map and not how it was made to be used should be supported, which is fundamentally wrong in my opinion. All the methods added here are extremely simple and multiple additions are missing crucial parts of how a Collection works, such as enforcing the `limit` and all entries having an `id` property. Therefore I believe this PR is unnecessary.

index.d.ts
```diff
@@ -1156,7 +1156,11 @@ declare namespace Eris {
   export class Collection<T extends { id: string | number }> extends Map<string | number, T> {
     baseObject: new (...args: any[]) => T;
     limit?: number;
-    constructor(baseObject: new (...args: any[]) => T, limit?: number);
+    constructor(baseObject: new (...args: any[]) => T, limit?: number, iterable?: object | any[]);
+    static from(array: any[], key: string, limit: number): Collection<any>;
+    toArray(): any[];
```
```suggestion
    toArray(): T[];
```

eritbh:
Can you show a usage example of the new `Collection#apply`? I'm not sure what it's trying to accomplish.

Khaaz:
I want to point out that if we don't want to see Collection as Map but only as Eris storage structure, we can implement these methods using `item.id` and slicing items over the limits when calling super().
Most of them would be kind of useless since it was aimed toward poeple using Eris.Collection for their own extended Map.
But i'd argue that `.apply` is somewhat useful. It's purpose being to just use any Map method but returning a Collection instead of an array.
Usage example:
`client.users.apply('id', 'filter', (e) => !e.bot)` Would return a Collection of users without bot.

index.d.ts
```diff
@@ -1156,7 +1156,11 @@ declare namespace Eris {
   export class Collection<T extends { id: string | number }> extends Map<string | number, T> {
     baseObject: new (...args: any[]) => T;
     limit?: number;
-    constructor(baseObject: new (...args: any[]) => T, limit?: number);
+    constructor(baseObject: new (...args: any[]) => T, limit?: number, iterable?: object | any[]);
+    static from(array: any[], key: string, limit: number): Collection<any>;
```
Skillz4Killz, line 1160:
```suggestion
    static from(array: T[], key: string, limit: number): Collection<T>;
```

index.d.ts
```diff
@@ -1156,7 +1156,11 @@ declare namespace Eris {
   export class Collection<T extends { id: string | number }> extends Map<string | number, T> {
     baseObject: new (...args: any[]) => T;
     limit?: number;
-    constructor(baseObject: new (...args: any[]) => T, limit?: number);
+    constructor(baseObject: new (...args: any[]) => T, limit?: number, iterable?: object | any[]);
```
```suggestion
    constructor(baseObject: new (...args: T[]) => T, limit?: number, iterable?: object | T[]);
```

index.d.ts
```diff
@@ -1156,7 +1156,11 @@ declare namespace Eris {
...
+    static from(array: any[], key: string, limit: number): Collection<any>;
+    toArray(): any[];
+    toObject(): object;
```
Skillz4Killz, line 1162: I don't like this. Theoretically, this is useless to typings. All you will get is `{}` and TS won't let you use it at all since `anyprop` does not exist on `{}`. I would prefer `unknown` to be more typesafe.
```suggestion
    toObject(): unknown;
```

index.d.ts
```diff
@@ -1156,7 +1156,11 @@ declare namespace Eris {
...
+    apply(key: string, func: string, ...args: any[]): Collection<any>;
```
```suggestion
    apply(key: string, func: string, ...args: T[]): Collection<T>;
```

index.d.ts
eritbh: Why would you make this `unknown`? That conveys even less than `object` does (`object` matches any non-primitive, not just `{}`.)

A better type, since you've proposed adding generics to the existing typings, would be:
```suggestion
    toObject(): {[key: T["id"]]: T};
```

eritbh:
@Khaazz Wouldn't it be significantly simpler to just do `Collection.from(myCollection.map(whatever => whatever))` - especially considering that's all it's doing anyway? I know Javascript is dynamic and all, but passing a method name as a string rubs me the wrong way.

Under what situation do you need a `Collection` instead of an array or object, anyway?

lib/util/Collection.js
```diff
@@ -10,14 +10,71 @@ class Collection extends Map {
...
+        } else {
+            super();
+        }
+
         this.baseObject = baseObject;
         this.limit = limit;
     }
 
+    /**
+     * Creates a collection from an array. Uses the first element instances as baseObject
```
apacheli: Asterisks should align with opening slash.

lib/util/Collection.js
```diff
@@ -10,14 +10,71 @@ class Collection extends Map {
...
+     *
+     * @static
+     * @param {Array<Class>} array The array of object
```
apacheli: Use `@arg` instead of `@param`

apacheli (changes requested):

index.d.ts
Khaaz: This doesn't work because it is static (I guess?).
TS on vscode doesn't seem to be happy with it at least?

index.d.ts
Khaaz: Same here, doesn't seem to be able to do what Geo suggests.
Complaining about the `key` part;
Maybe I should do it anyway, not really familiar with ts / typings

Khaaz:
Sorry for the late reply.
I agree, I don't really like going too far with Javascript dynamic aspect either. But it's just a convenient method that wouldn't actually do any bad imho.
I also agree about using Collection.from() like you mentionned. Would probably be the cleanest way to do that. But apply is really just here to make it all more convenient.
As to why: it was mainly in case for some reason you want a Collection instead of an array because you possibly want to do smthing else with it later? I can remember of some djs users coming by and being surprised it was returning an array everytime sooo 🤷‍♂ 

I really just took these methods for my own Collection implementation and thought it could potentially be useful to anyone else.

Honestly i'm just waiting for abal's review at this point.

index.d.ts
eritbh: If my suggestion above causes issues, using the `object` type is probably a safe bet. I'll look into it in a bit.

Skillz4Killz: I disagree about using `object`. Object is too broad and can cause incorrect typings issues. It is not typesafe. 

```js
const test = (idk: object) => {
    console.log(idk)
  }

  // Valid
  test(test)
  // Valid
  test({ idk: 'test' })
```

Having it return `object` can break a lot of typings which lead to potential bugs/errors. `object` will make it so it will be treated as an array, function, anything really. object is a useless and bad type to be using.

The worst part is if you tried to do object.id it will throw typings errors.
```js
const test = (idk: object) => {
    console.log(idk.id)
  }
```

Using `unknown` is a much better alternative. It is typesafe.

If possible id actually prefer using ur suggested change. I think that is better than `unknown` but if its a choice between object or unknown, `unknown` is the better choice.

eritbh: Most of the issues with using `object` apply when using it as the type of a parameter, but disappear when using it as the type of a return value. It just requires the consumer to do more type checks at runtime since TS doesn't have the type info.

Regardless, I've verified a fixed version of my suggestion above. The way TS handles index signatures is weird, so what we really want is this:

```suggestion
    toObject(): {[key: string]: T | undefined};
```

Using a `number` as a key works fine even when the index signature is of type `string` since all object keys are strings (or symbols, but we don't care about those) internally. I also forgot to use `| undefined` before.

abalabahaha:
I think this is a little out of scope for Eris. Can be useful for other purposes though. Thanks for the PR

PR #596, opened 2019-12-16
Slight change to RequestHandler debug message

DonovanDMC:
Changed debug message to look cleaner, and provide more info, instead of
`undefined 1576528937750 /channels/614408413492936705/messages/:id/reactions/:id/@me 204: 196ms (330ms avg) | 0/1 left | Reset 1576528937750 (0ms left)`, we have `None 1576528937750 POST /channels/614408413492936705/messages/:id/reactions/:id/@me 204: 196ms (330ms avg) | 0/1 left | Reset 1576528937750 (0ms left)`, removing that nasty undefined when there wasn't a body, and adding in the request method.

### DIFF
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -236,7 +236,7 @@ class RequestHandler {
                         }
 
                         if(resp.statusCode !== 429) {
-                            this._client.emit("debug", `${body && body.content} ${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
+                            this._client.emit("debug", `${body && body.content || "None"} ${now} ${method} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
                         }
 
                         if(resp.statusCode >= 300) {


lib/rest/RequestHandler.js
```diff
@@ -236,7 +236,7 @@ class RequestHandler {
                         }
 
                         if(resp.statusCode !== 429) {
-                            this._client.emit("debug", `${body && body.content} ${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
+                            this._client.emit("debug", `${body && body.content || "None"} ${now} ${method} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
```
Skillz4Killz, line 239: None is too common and isn't explicit to being what exactly is None. Also sometimes errors actually respond with None so its safer to use a different term imo. `undefined` is fine if it is prefixed with what is undefined. In this case just adding Body Content is enough imo.
```suggestion
                            this._client.emit("debug", `Body Content: ${body && body.content} ${now} ${method} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
```

abalabahaha:
I've been meaning to convert this to a JSON-encoded message actually, which would be a lot more usable ("what even is this number???").

DonovanDMC:
It would be nice if all of the debug events were JSON and just had data properties on them instead of just a bunch of strings

abalabahaha:
Rejecting in favor of upcoming JSON messages

PR #597, opened 2019-12-17
Added bannerURL/splashURL typings

oathompsonjones:


reboxer (commented):
maybe also add splashURL?

index.d.ts
```diff
@@ -1225,7 +1225,9 @@ declare namespace Eris {
     joinedAt: number;
     ownerID: string;
     splash?: string;
+    splashURL?: string;
```
```suggestion
    splashURL: string | null;
```

index.d.ts
```diff
@@ -1225,7 +1225,9 @@ declare namespace Eris {
...
     banner?: string;
+    bannerURL?: string;
```
```suggestion
    bannerURL: string | null;
```

Skillz4Killz (commented):
Using `?` to mark something as optional makes typescript understand it as if it is `undefined`. The actual behavior in Eris never allows this to be undefined. It marks it as `null` making this optional and not explicitly null can cause issues to end users.

oathompsonjones:
Wasn't aware those would return null as opposed to undefined, thanks for letting me know.

Skillz4Killz (approved):

abalabahaha (approved):

PR #599, opened 2019-12-20
Added `4` as an option for `Activity#type`

oathompsonjones:
Also edited `GamePresence#type` to match


oathompsonjones:
Has anybody found any difference between `Activity` and `GamePresence` when in use? From what I can see it might be better to just have one interface for them as they seem to return the same values in practice.

alex-taxiera:
@oathompsonjones I'll take a look at this over the next few weeks. My bot deals exclusively with presences, but I never had to pay much attention to Activity before.

Skillz4Killz (approved):

abalabahaha (approved):

PR #600, opened 2019-12-23
Bump lowest supported Node to v10

AnotherGenZ:
On December 31st, 2019, Node v8 will reach EOL. Thus, we should no longer support it either. Additionally, this'll open us up to using features from newer versions without worrying about backwards compatibility.


abalabahaha:
This is fine as a development guideline, but I see no reason to break compatibility prematurely. This should be done when the code actually starts to use Node >=10 features.

PR #602, opened 2019-12-27
Improved ESM support

LJNeon:
ECMAScript modules can import CommonJS modules, however this is simply as a default export and can't take advantage of named exports. This PR adds named exports to Eris when imported by an ECMAScript module.
 - CommonJS support for Node.js 8+ (and for obsolete versions before 8).
 - Full ECMAScript support for Node.js 12+ (using an `eris/esm` import and/or flags).
 - Forwards compatibility with future versions of Node.js using [conditional exports](https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_conditional_exports).
 - `Eris.VERSION` is temporarily removed from the ECMAScript exports, as importing JSON files is [experimental and behind a flag](https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_experimental_json_modules). The code needed to add it back is commented out.
 - Updated dependencies, as every single one of them were outdated.

One side effect of this PR is that it prevents using [subpath imports](https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_package_exports). Whether this is a good thing or not is worth discussion, but it causes no problems with normal use nor with TypeScript. An alternative method could be using the `eris/esm.mjs` subpath import, as while it's less appealing and lacks forwards compatibility it doesn't prevent use of subpaths.

### DIFF
--- a/esm.mjs
+++ b/esm.mjs
@@ -1,0 +1,40 @@
+import Client from "./lib/Client.js";
+// import pkg from "./package.json";
+
+export default function(token, options) {
+  return new Client(token, options);
+}
+
+export {default as Base} from "./lib/structures/Base.js";
+export {default as Bucket} from "./lib/util/Bucket.js";
+export {default as Call} from "./lib/structures/Call.js";
+export {default as CategoryChannel} from "./lib/structures/CategoryChannel.js";
+export {default as Channel} from "./lib/structures/Channel.js";
+export {Client};
+export {default as Collection} from "./lib/util/Collection.js";
+export {default as Command} from "./lib/command/Command.js";
+export {default as CommandClient} from "./lib/command/CommandClient.js";
+export {default as Constants} from "./lib/Constants.js";
+export {default as ExtendedUser} from "./lib/structures/ExtendedUser.js";
+export {default as GroupChannel} from "./lib/structures/GroupChannel.js";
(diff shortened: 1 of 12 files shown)


esm.mjs
```diff
@@ -0,0 +1,40 @@
+import Client from "./lib/Client.js";
+// import pkg from "./package.json";
+
+export default function(token, options) {
+  return new Client(token, options);
+}
+
+export {default as Base} from "./lib/structures/Base.js";
```
apacheli, line 8: Little nitpick, but you can save a few characters by doing `import name from 'module'` and `export { name1, name2, ...names }`

apacheli (changes requested):

esm.mjs
LJNeon: Personally I think that it's far nicer to only import modules as needed, and to just export the rest as I did currently, if enough people disagree then I could change that easily though.

eritbh: I agree with eljay, I prefer using the more explicit `export from` whenever doing pass-through exports from other files. With `import`/`export`, it's not immediately obvious that the imported modules aren't being used directly within the file. I find this form a bit more readable.

abalabahaha:
This needs to be split into a couple different PRs/commits, and we decided that re-exporting `index.js` would be a better approach for now.

PR #607, opened 2020-01-02
Fix Member#clientStatus

### LINKED ISSUE #605: ClienStatus
Hello,
```
Eris version ==> 0.11.1
Eris Sharder ==> 1.9.0
```
ClientStatus returns to an undefined state when offline on the web, desktop and mobile.

with the following code 
```js
if (message.mentions.length === 0) {
    return message.channel.createMessage("Please mention a user");
             }
         let mem = message.channel.guild.members.get(message.mentions[0].id);
         console.log(JSON.stringify(mem.clientStatus))

apacheli:
Fixes #605

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1495,7 +1495,7 @@ declare namespace Eris {
 
   export class Member extends Base {
     activities?: Activity[];
-    clientStatus?: ClientStatus;
+    clientStatus: ClientStatus;
     id: string;
     mention: string;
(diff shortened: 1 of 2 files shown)


Skillz4Killz:
Needs typings updated if memory serves we just added a PR that made clientStatus?: and it is no longer an optional prop. Have to remove the ?

Skillz4Killz (approved):

PR #610, opened 2020-01-07
Add missing properties to Invite typings & Invite class

DonovanDMC:
Adding in some invite properties that seem to have not been accounted for.
[api screenshot](https://butts-are.cool/Postman_HLPXScXn7h.png)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1461,8 +1461,6 @@ declare namespace Eris {
       features: string[];
       verificationLevel?: 0 | 1 | 2 | 3 | 4;
       vanityUrlCode?: string;
-      textChannelCount?: number;
-      voiceChannelCount?: number;
     };
     inviter?: User;
     uses?: number;
@@ -1470,7 +1468,6 @@ declare namespace Eris {
     maxAge?: number;
     temporary?: boolean;
     createdAt?: number;
-    revoked?: boolean;
     presenceCount?: number;
     memberCount?: number;
     constructor(data: BaseData, client: Client);
--- a/index.d.ts
+++ b/index.d.ts
@@ -1308,7 +1308,7 @@ declare namespace Eris {
     reason?: string;
     user: User;
     targetID: string;
-    target?: Guild | AnyGuildChannel | Member | RestChannelInvite | Role | any;
+    target?: Guild | AnyGuildChannel | Member | Role | any;
     before?: any;
     after?: any;
     count?: number;
--- a/index.d.ts
+++ b/index.d.ts
@@ -935,8 +935,8 @@ declare namespace Eris {
     getPruneCount(guildID: string, days: number): Promise<number>;
     pruneMembers(guildID: string, days: number, reason?: string): Promise<number>;
     getVoiceRegions(guildID: string): Promise<VoiceRegion[]>;
-    getInvite(inviteID: string, withCounts?: boolean): Promise<RestInvite>;
-    acceptInvite(inviteID: string): Promise<RestInvite>;
+    getInvite(inviteID: string, withCounts?: boolean): Promise<RESTInvite>;
+    acceptInvite(inviteID: string): Promise<RESTInvite>;
     deleteInvite(inviteID: string, reason?: string): Promise<void>;
     getSelf(): Promise<ExtendedUser>;
     editSelf(options: { username?: string; avatar?: string }): Promise<ExtendedUser>;
@@ -1444,8 +1444,8 @@ declare namespace Eris {
     sync(): Promise<void>;
   }
 
-  type RestInvite = RestChannelInvite | RestPrivateInvite;
-  type AnyInvite = RestInvite | ChannelInvite;
+  type RESTInvite = RESTChannelInvite | RESTPrivateInvite;
+  type AnyInvite = RESTInvite | ChannelInvite;
   interface BaseInvite {
     code: string;
     channel: {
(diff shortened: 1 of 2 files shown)


DonovanDMC:
Also, this bit at the top of [Invite.js](https://github.com/DonovanDMC/eris/blob/patch-12/lib/structures/Invite.js#L6) really isn't true.

> The nullable properties will be null if the bot user does not have manage channel or manage server permissions for the invite's channel/server.

In that [first screenshot](https://butts-are.cool/Postman_HLPXScXn7h.png), I fetched that invite with a bot that has administrator permissions, and yet those properties still didn't show. [[1]](https://butts-are.cool/Postman_p29GCvobAX.png) [[2]](https://butts-are.cool/DiscordCanary_5A0ZVBk5Zp.png) [[3]](https://butts-are.cool/DiscordCanary_oPITygwWSd.png)

Same story for a server the bot client isn't even in ([screenshot](https://butts-are.cool/Postman_QyTQ4ISM8P.png))

LJNeon:
Going by your screenshots it looks like there's more parts of the Invite class that could use updating, also you need to use 4 spaces for indentation, not tabs.

DonovanDMC:
I'm aware of the tab indenting, github is screwing with my indenting

index.d.ts
```diff
@@ -1446,22 +1446,28 @@ declare namespace Eris {
 
   export class Invite implements SimpleJSON {
     code: string;
-    channel: { id: string; name: string };
+    channel: {
+      id: string;
+      name: string;
+      type: 0 | 2 | 4 | 5 | 6;
+    };
     guild: {
```
Skillz4Killz, line 1680: Im not sure if we are still supporting this but according to the discord docs you can also have Invites for Group DMs. This means guild could be optional.

image

If this is accurate and it is a "partial", if memory serves the only guaranteed props on partials are ids. This means all props like name and features should be optional 🤔 Not sure on this.

lib/structures/Invite.js
```diff
@@ -3,28 +3,32 @@
...
+ * @prop {String[]} guild.features An array of the invite's guild's features
+ * @prop {Number?} guild.verificationLevel The verification level of the invite's guild
+ * @prop {String?} guild.vanityUrlCode The vanity url of the invite's guild
+ * @prop {User?} inviter The invite creator
+ * @prop {Number?} uses The number of invite uses
+ * @prop {Number?} maxUses The max number of invite uses
+ * @prop {Number?} maxAge How long the invite lasts in seconds
+ * @prop {Boolean?} temporary Whether the invite is temporary or not
+ * @prop {Number?} createdAt Timestamp of invite creation
+ * @prop {Boolean?} revoked Whether the invite was revoked or not
```
Skillz4Killz: This was removed in typings, should it not also be removed here?

Skillz4Killz: I would also check if there is code below in the constructor for this as well and remove that.

index.d.ts
```diff
@@ -1446,22 +1446,28 @@ declare namespace Eris {
 
   export class Invite implements SimpleJSON {
     code: string;
-    channel: { id: string; name: string };
+    channel: {
```
Skillz4Killz, line 1663: Similar thoughts on this as well since it is a partial channel are we certain that all these props will exist?

lib/structures/Invite.js
```diff
@@ -33,12 +37,15 @@ class Invite extends Base {
         this.channel = data.channel;
         if(data.guild) {
             this.guild = {
-                splash: data.guild.splash,
-                icon: data.guild.icon,
                 id: data.guild.id,
                 name: data.guild.name,
-                textChannelCount: data.guild.text_channel_count !== undefined ? data.guild.text_channel_count : null,
-                voiceChannelCount: data.guild.voice_channel_count !== undefined ? data.guild.voice_channel_count : null
+                splash: data.guild.splash,
```
Skillz4Killz, line 42: What happens if `guild` is undefined? 
image

DonovanDMC: image

index.d.ts
DonovanDMC: Channels will always have a type and id, if it's an invite to a group DM name will be null, and icon will exist, so I'll update that

DonovanDMC: name will be null if the group dm doesn't have a name*

lib/structures/Invite.js
DonovanDMC: revoked was removed from the typings, but the rest were not

index.d.ts
DonovanDMC: guild has been made optional

DonovanDMC:
Different fixes:
Fixed CategoryChannel only having Text/Voice channels as child channels (+News/Store)

Also fixed `getGuildVanity` returning `{ code: Invite }`, it returns ` { code?: string; uses: number; }`
[docs](https://discordapp.com/developers/docs/resources/guild#get-guild-vanity-url)

Skillz4Killz (dismissed):

abalabahaha (changes requested):
Consistency nit: `RESTInvite` instead of `RestInvite`

lib/structures/Invite.js
```diff
@@ -3,28 +3,32 @@
...
-* @prop {Number?} uses The number of invite uses
-* @prop {Number?} maxUses The max number of invite uses
-* @prop {Number?} maxAge How long the invite lasts in seconds
-* @prop {Boolean?} temporary Whether the invite is temporary or not
-* @prop {Number?} createdAt Timestamp of invite creation
-* @prop {Boolean?} revoked Whether the invite was revoked or not
-* @prop {Number?} presenceCount The **approximate** presence count for the guild
-* @prop {Number?} memberCount The **approximate** member count for the guild
-*/
+ * Represents an invite. Some properties are only available when fetching invites from channels, which requires the Manage Channel permission.
```
abalabahaha: This block shouldn't have been indented

abalabahaha (approved):

PR #612, opened 2020-01-13
Fix PCM decoding that break by shared opus instance

### LINKED ISSUE #443: Voice receiving on pcm mode is broken when multiple users are speaking
Reuse opus instance between different voice streams causes decode error

https://github.com/abalabahaha/eris/blob/e6208fa8ab49d526df5276620ac21eb351da3954/lib/voice/VoiceConnection.js#L667

mmis1000:
- User voice decoding in `VoiceConnection.js` is now using own codec per user.
- All opus codec are now initiated per stream (both decode/encode) because share the instance cause decode/encode error.
- Add `disconnect` event to `VoiceConnection.js` in order to trigger codec clean up on user leaved.
- Move opus codec creation into own function to prevent duplication.

This PR is aimed to Fix #443 

~~Not ready yet, only `VoiceConnection` was tested~~

VoiceConnection was tested with my own bot https://github.com/mmis1000/discord-record-bot-eris to ensure sound no longer garbled(sounds exactly same like you sound on discord itself). Other things may still requires a test to ensure it does not go wrong.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -76,7 +76,8 @@ module.exports.VoiceOPCodes = {
     READY:               2,
     HEARTBEAT:           3,
     SESSION_DESCRIPTION: 4,
-    SPEAKING:            5
+    SPEAKING:            5,
+    DISCONNECT:          13
 };
 
 module.exports.SystemJoinMessages = [
--- a/lib/util/Opus.js
+++ b/lib/util/Opus.js
@@ -1,0 +1,29 @@
+let NodeOpus;
+try {
+    NodeOpus = require("node-opus");
+} catch(err) { // eslint-disable no-empty
+}
+let OpusScript;
+try {
+    OpusScript = require("opusscript");
+} catch(err) { // eslint-disable no-empty
+}
+
+
+module.exports.createOpus = function createOpus (samplingRate, channels, bitrate) {
+    let opus;
+
+    if(NodeOpus) {
+        opus = new NodeOpus.OpusEncoder(samplingRate, channels);
+    } else if(OpusScript) {
+        opus = new OpusScript(samplingRate, channels, OpusScript.Application.AUDIO);
+        if(opus.setBitrate) {
+            opus.setBitrate(bitrate);
+        } else if(opus.encoderCTL) {
+            opus.encoderCTL(4002, bitrate);
+        }
+    } else {
+        throw new Error("No opus encoder found, playing non-opus audio will not work.");
+    }
+    return opus;
+};
--- a/lib/util/Opus.js
+++ b/lib/util/Opus.js
@@ -10,20 +10,22 @@ try {
 }
 
 
-module.exports.createOpus = function createOpus (samplingRate, channels, bitrate) {
-    let opus;
-
+module.exports.createOpus = function createOpus(samplingRate, channels, bitrate) {
     if(NodeOpus) {
-        opus = new NodeOpus.OpusEncoder(samplingRate, channels);
-    } else if(OpusScript) {
-        opus = new OpusScript(samplingRate, channels, OpusScript.Application.AUDIO);
+        return new NodeOpus.OpusEncoder(samplingRate, channels);
+    }
+
+    if(OpusScript) {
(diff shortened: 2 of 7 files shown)


lib/Constants.js
```diff
@@ -76,7 +76,8 @@ module.exports.VoiceOPCodes = {
     READY:               2,
     HEARTBEAT:           3,
     SESSION_DESCRIPTION: 4,
-    SPEAKING:            5
+    SPEAKING:            5,
+    DISCONNECT:          13
```
Skillz4Killz, line 80: Typings file line 207 needs updating for this plz

lib/util/Opus.js
```diff
@@ -0,0 +1,29 @@
...
+} catch(err) { // eslint-disable no-empty
+}
+let OpusScript;
+try {
+    OpusScript = require("opusscript");
+} catch(err) { // eslint-disable no-empty
+}
+
+
+module.exports.createOpus = function createOpus (samplingRate, channels, bitrate) {
```
Skillz4Killz:
```suggestion
module.exports.createOpus = function createOpus(samplingRate, channels, bitrate) {
```

lib/util/Opus.js
```diff
@@ -0,0 +1,29 @@
...
+        if(opus.setBitrate) {
+            opus.setBitrate(bitrate);
+        } else if(opus.encoderCTL) {
+            opus.encoderCTL(4002, bitrate);
+        }
+    } else {
+        throw new Error("No opus encoder found, playing non-opus audio will not work.");
+    }
+    return opus;
+};
```
Skillz4Killz: I believe this code has potential to be improved for readability and maintainability.

I would suggest considering something like:

```js
if (!NodeOpus && !OpusScript) {
  throw new Error("No opus......")
}

if (NodeOpus) {
  return new NodeOpus.OpusEncoder(samplingRate, channels)
}

const opus = new OpusScript(samplingRate, channels, OpusScript.Application.AUDIO)
if (opus.setBitrate) {
  opus.setBitrate(bitrate)
} else if (opus.encoderCTL) {
  opus.encoderCTL(4002, bitrate)
}

return opus
```

lib/util/Opus.js
Skillz4Killz: Consistency. See `pickCommand` on VoiceConnection

lib/voice/VoiceConnection.js
```diff
@@ -325,9 +308,21 @@ class VoiceConnection extends EventEmitter {
...
                 }
+                case VoiceOPCodes.DISCONNECT: {
+                    // opusscript requires explicit release wasm instance
+                    if(this.opus[packet.d.user_id] && this.opus[packet.d.user_id].delete) {
+                        this.opus[packet.d.user_id].delete();
+                    }
+
+                    delete this.opus[packet.d.user_id];
+
+                    this.emit("disconnect", packet.d.user_id);
```
Skillz4Killz: This will also need typings please

lib/Constants.js
mmis1000: Added

lib/voice/VoiceConnection.js
mmis1000: Added

lib/util/Opus.js
mmis1000: I am not very confident about what did you mean.

Did you means some structure like

```js
module.exports.opusFactory = {
  impl: null,
  selectImpl () {
    // whatevet...
  },
  create () {
     if (this.impl == null) this.selectImpl()
    // whatevet...
     return opus
  }
}
```

?

lib/util/Opus.js
mmis1000: Refactored with early return

lib/util/Opus.js
Skillz4Killz: I just meant that the other time there is something like `= function x(` there is no space between the name and the (
image

mmis1000: ok, fixed

abalabahaha (approved):

PR #616, opened 2020-01-16
Intent Support

Brayzure:
Initial PR
- Adds support for registering intents.

ToDo
- Check to make sure this doesn't horrifically break things, and fix conflicts as they arise.

Also worth considering, should intent defining be moved elsewhere? It's going to be required in gateway v7 (whenever it releases), so moving it out of options into its own argument preemptively may be a good idea.

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -148,8 +148,8 @@ class Client extends EventEmitter {
             this.options.ws.agent = this.options.agent;
         }
 
-        // Resolve intents option to the proper integer
         if(this.options.intents) {
+            // Resolve intents option to the proper integer
             if(Array.isArray(this.options.intents)) {
                 let bitmask = 0;
                 for(const intent of this.options.intents) {
@@ -158,6 +158,11 @@ class Client extends EventEmitter {
                     }
                 }
                 this.options.intents = bitmask;
+            }
+
+            // Ensure requesting all guild members isn't destined to fail
+            if(this.options.getAllUsers && !(this.options.intents & Constants.Intents.guildMembers)) {
+                throw new Error("Cannot request all members without guildMembers intent");
             }
         }
 
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -148,7 +148,7 @@ class Client extends EventEmitter {
             this.options.ws.agent = this.options.agent;
         }
 
-        if(this.options.intents) {
+        if(this.options.hasOwnProperty("intents")) {
             // Resolve intents option to the proper integer
             if(Array.isArray(this.options.intents)) {
                 let bitmask = 0;
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -98,7 +98,7 @@ class Client extends EventEmitter {
     * @arg {Object} [options.agent] A HTTP Agent used to proxy requests
     * @arg {Number} [options.maxReconnectAttempts=Infinity] The maximum amount of times that the client is allowed to try to reconnect to Discord.
     * @arg {Function} [options.reconnectDelay] A function which returns how long the bot should wait until reconnecting to Discord.
-    * @arg {Number | String | String[]} [options.intents] An intent, list of intents, or raw bitmask value describing the intents to subscribe to. "presence" intent must be enabled on your application's page to be used.
+    * @arg {Number | String[]} [options.intents] A list of intents, or raw bitmask value describing the intents to subscribe to. "presence" intent must be enabled on your application's page to be used.
     */
     constructor(token, options) {
         super();
@@ -162,11 +162,6 @@ class Client extends EventEmitter {
                     }
                 }
                 this.options.intents = bitmask;
-            }
-            if(typeof this.options.intents === "string") {
-                if(Constants.Intents[this.options.intents]) {
-                    this.options.intents = Constants.Intents[this.options.intents];
-                }
             }
         }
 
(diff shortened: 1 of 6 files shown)


Skillz4Killz:
Typings for ClientOptions need to be updated for this please.

1. add `intents` as a possible prop
2. i would even go so far as making all the `strings` available as well for autocompletion by having the strings in a type somewhere

```js
type IntentStrings = 'guilds' | '...' | '...'..........

interface ClientOptions {
  intents?: number | IntentStrings[]
}
```

abalabahaha:
The pattern of having required props in an options object isn't uncommon, I'd say this is fine

Dramex:
abal please don't forget this PR 😆

LJNeon:
It's not forgotten, it's missing crucial parts which prevents it from being merged. Hence why it has the postponed tag.

Dramex:
Great job guys! i can't wait to see intents support in Eris 🥰, yeah i wrote the comment because of the postponed tag + inactivity for 17 days, i might misunderstand the tag.

Brayzure:
Nah I've been otherwise occupied. Needed to add typings and an example before it's ready for merging.

Astraeus-:
Probably need to document that they're also adding a whitelist requirement for guild members https://github.com/discordapp/discord-api-docs/pull/1307#issuecomment-581561519

lib/Client.js
```diff
@@ -98,6 +98,7 @@ class Client extends EventEmitter {
     * @arg {Object} [options.agent] A HTTP Agent used to proxy requests
     * @arg {Number} [options.maxReconnectAttempts=Infinity] The maximum amount of times that the client is allowed to try to reconnect to Discord.
     * @arg {Function} [options.reconnectDelay] A function which returns how long the bot should wait until reconnecting to Discord.
+    * @arg {Number | String[]} [options.intents] A list of intents, or raw bitmask value describing the intents to subscribe to. "presence" intent must be enabled on your application's page to be used.
```
Astraeus-, line 101: `Guild members` also needs to be enabled on the application's page to be used as per https://github.com/discordapp/discord-api-docs/issues/1363#issue-565525845

lib/Constants.js
```diff
@@ -201,3 +201,21 @@ module.exports.ChannelTypes = {
...
+    guildInvites:           1 << 6,
+    guildVoiceStates:       1 << 7,
+    guildPresences:         1 << 8,
+    guildMessages:          1 << 9,
+    guildMessageReactions:  1 << 10,
+    guildMessageTyping:     1 << 11,
+    directMessages:         1 << 12,
+    directMessageReactions: 1 << 13,
+    directMessageTyping:    1 << 14
+};
```
Zoddo, line 221: Maybe we can use the same format than in the [Discord's documentation](https://discordapp.com/developers/docs/topics/gateway#list-of-intents) (ex: `GUILD_MEMBERS` instead of `guildMembers`), like what we did for OPCodes?

Brayzure: Would prefer to keep it as is, since it's the same format as we use for event names. It's formatted like that in Constants because intent integer calculating checks directly against Constants.

Brayzure: See also: Permissions, which use camelCase. There's nowhere in the lib where a dev using the lib has to use Discord case, would not be the greatest idea to change that now.

Skillz4Killz:
Have this running in production with 5k servers and 1.5mil users atm with 5 shards. Everything looks good after 12hrs running in production! Great stuff! LGTM to merge!

lib/util/Bucket.js
```diff
@@ -13,23 +13,30 @@ class Bucket {
     * Construct a Bucket
     * @arg {Number} tokenLimit The max number of tokens the bucket can consume per interval
     * @arg {Number} interval How long (in ms) to wait between clearing used tokens
+    * @arg {Number} reserved How many tokens to reserve for priority operations
     * @arg {Object} [latencyRef] An object
     * @arg {Number} latencyRef.latency Interval between consuming tokens
     */
-    constructor(tokenLimit, interval, latencyRef = {latency: 0}) {
+    constructor(tokenLimit, interval, reserved = 0, latencyRef = {latency: 0}) {
```
abalabahaha: Not a fan of adding it like this as it breaks things... Could use an options object as 4th param

abalabahaha (changes requested):

lib/util/Bucket.js
Brayzure: Strikes me as a bit odd to do it like that. Alternatively, we could merge `latencyRef` and `reserved` into a single options object. I don't even see anywhere in the lib where we explicitly define `latencyRef`, though search could be failing me.

Skillz4Killz:
Updated to latest in production. Everything working as intended. Noticed a slight delay on bootup but i think that could be attributed to a number of factors like network and the discord API was also being a bit unstable earlier today. After 2 hours and thousands of commands/events etc run, this should be safe to merge in its current state.

lib/Client.js
```diff
@@ -147,6 +148,24 @@ class Client extends EventEmitter {
             this.options.ws.agent = this.options.agent;
         }
 
+        if(this.options.intents) {
```
apacheli: This will cause unintended behavior if someone sets `intents: 0` and `getAllUsers: true`

apacheli (changes requested):

lib/Client.js
Brayzure: Ah yes, took me a moment to realize what you meant. Well, checking for the existence of the property instead should work. Thanks for pointing that out.

abalabahaha (approved):

PR #617, opened 2020-01-16
Add guildID prop to Message

Skillz4Killz:
This PR adds a `guildID` prop onto the Message class. The data that discord sends has a `guild_id` prop attached to it and it is made unavailable to the eris user. Instead it is made that users do `message.channel.guild.id`.

I still don't truly understand why `message.guild` is not a thing but having the guild_id prop exposed would be HUGE. 99% of the times i do `message.channel.guild` is just to get the guildID. 

image

The current dilemma with having to do `message.channel.guild` is it forces any TS user to have to add additional checks.

Everytime i make a function I have to run these two checks just to get a guild id. 
image

It's silly!
image

Eris should not make a prop that Discord passes unavailable to the end user, please 🙏.

Side Note: Some good points were brough up in discord chat about having a `message.guild` getter. If desired i can add this in.
```js
get guild() {
  return this.guildID ? this.client.guilds.get(guildID) || { id: this.guildID } : null
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1530,7 +1530,7 @@ declare namespace Eris {
     id: string;
     createdAt: number;
     channel: TextableChannel;
-    guildID: string;
+    guildID?: string;
     timestamp: number;
     type: number;
     author: User;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1530,6 +1530,7 @@ declare namespace Eris {
     id: string;
     createdAt: number;
     channel: TextableChannel;
+    guildID: string;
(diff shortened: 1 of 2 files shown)


abalabahaha:
> Eris should not make a prop that Discord passes unavailable to the end user

Much of the current code does this actually, it's not out of line.

That being said, this is a rather useful property, especially with subscriptions/intents now.

`Message#guild` does need to be re-evaluated later, now that intents are coming.

abalabahaha (approved):

PR #621, opened 2020-01-17
Add type parameter to Message typings

alex-taxiera:


index.d.ts
```diff
@@ -1526,10 +1526,10 @@ declare namespace Eris {
     unban(reason?: string): Promise<void>;
   }
 
-  export class Message extends Base {
+  export class Message<T extends Textable = TextableChannel> extends Base {
```
eritbh, line 1576: Is making the default here `TextableChannel` okay? I'm not sure if this will allow you to type-check the `Message` into a `Message<GuildChannel>` since TS is a bit weird about subclasses. It might need to be this instead:

```suggestion
  export class Message<T extends Textable = TextChannel | PrivateChannel> extends Base {
```

alex-taxiera: `GuildChannel` can never even satisfy the constraint of `Textable`, because a `GuildChannel` is not always `Textable` (`VoiceChannel`)

eritbh: I was thinking about this wrong; `TextableChannel` is a type alias for the union I'm talking about. I was thinking it was a class for some reason. This is fine as-is.

eritbh (approved):

abalabahaha (approved):

PR #622, opened 2020-01-18
Expose Channel.client

LJNeon:
Currently, there's no way to access the `Client` from a `Message` that was sent in DMs. Use cases for this aren't that uncommon and for those, the only option is to use internals. This can be effortlessly solved by exposing `Channel.client`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1239,6 +1239,7 @@ declare namespace Eris {
     id: string;
     mention: string;
     type: 0 | 1 | 2 | 3 | 4 | 5 | 6;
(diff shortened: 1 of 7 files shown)


alex-taxiera:
Does the typings file also need to be updated to reflect this change? Docs too?

LJNeon:
`TextChannel.guild.shard.client` would be the equivalent of this and is also undocumented, so I choose to keep it the same, as for typings I can correct that now.

abalabahaha (approved):

PR #624, opened 2020-01-18
Remove presence duplicate check

apacheli:


Brayzure:
This used to be an issue, does this no longer happen?

abalabahaha:
It does, but the # of properties to check over has multiplied. Would probably be nicer to leave it to users to check for duplicates over properties they care about (some bots only care about status changes, or primary game name changes), and free up some CPU for the bots that don't care at all.

abalabahaha (approved):

PR #625, opened 2020-01-19
Adds inviteCreate and inviteDelete events

Skillz4Killz:
This PR adds the recently added invite related events. 

1. The idea behind having a collection of invites in `guild.invites` was because on `inviteDelete` event we would have no way of getting the invites data as we would only have the guild_id channel_id and code. This lead to the decision of deciding to cache invites in a collection on guilds.
  a. In order to have Invite in a Collection(requires an id prop), I decided to use the `code` as an id prop as well. Note: this means that `invite.id` and `invite.code` are essentially duplicate. Thoughts on removing `invite.code`?
  b. This also meant that we would need to fetch all events when the bot started so it could fill it up. I decided to fetch them in the `Guild` constructor so whenever a Guild is created in `guildCreate` or on startup, it will have invites filled. In order to fetch messages it was required that the bot have `manageGuild` permission so i clarified this in the jsdocs.
2. Updated the ecmaVersion from `2017` to `2018` in eslintrc as it was not allowing `...` operator to be used in objects.
3. Refactored the `this.guild` and `this.inviter`

### DIFF
--- a/lib/structures/Guild.js
+++ b/lib/structures/Guild.js
@@ -2,8 +2,7 @@
 
 const Base = require("./Base");
 const Channel = require("./Channel");
-const {CDN_URL} = require("../rest/Endpoints");
-const {ImageFormats, ImageSizeBoundaries} = require("../Constants");
+const Endpoints = require("../rest/Endpoints");
 const Collection = require("../util/Collection");
 const GuildChannel = require("./GuildChannel");
 const Member = require("./Member");
@@ -204,7 +203,7 @@ class Guild extends Base {
     }
 
     get iconURL() {
-        return this.icon ? `${CDN_URL}/icons/${this.id}/${this.icon}.${this.icon.startsWith("a_") ? "gif" : this._client.options.defaultImageFormat}?size=${this._client.options.defaultImageSize}` : null;
+        return this.icon ? this._client._formatImage(Endpoints.GUILD_ICON(this.id, this.icon)) : null;
     }
 
     /**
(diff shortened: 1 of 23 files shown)


lib/structures/Guild.js
```diff
@@ -59,6 +61,7 @@ class Guild extends Base {
         this.members = new Collection(Member);
         this.memberCount = data.member_count;
         this.roles = new Collection(Role);
+        this.invites = new Collection(Invite);
```
abalabahaha: Bad idea

lib/structures/Guild.js
```diff
@@ -119,9 +122,18 @@ class Guild extends Base {
...
         }
+
+        this._fetchAllInvites();
         this.update(data);
     }
 
+    _fetchAllInvites() {
+        const botMember = this.members.get(this._client.user.id);
+        if(botMember && botMember.permission.has("manageGuild")) {
+            return this.getInvites().then((invites) => invites.forEach((invite) => this.invites.add(invite)));
```
abalabahaha: Also a bad idea, the library shouldn't make automatic REST calls that aren't strictly necessary

lib/structures/Invite.js
```diff
@@ -29,21 +30,9 @@ class Invite extends Base {
     constructor(data, client) {
         super();
         this._client = client;
+        this.id = data.code;
         this.code = data.code;
         this.channel = data.channel;
-        if(data.guild) {
```
abalabahaha: Hides information in invites from REST from guilds that aren't cached

Skillz4Killz: Could you explain this to me? I don't understand how having this inside an if is useful.

```js
if (data.x) {
  this.guild = y
}
```
is equivalent to:
```js
this.guild = data.x ? y : undefined
```

Let's suppose the goal of this if was to hide info in invites on REST calls from UNCACHED guilds. If `data.guild` is not passed and the user does `invite.guild` they will get `undefined`. The same is true with a ternary. If a `data.guild` is not passed it will be undefined.

Am I missing something here?

Skillz4Killz: image


image


image


image


image
---
Bot is NOT on this guild:
image

image



If there is any way that i did not fetch an invite let me know and id be happy to test that way as well

abalabahaha: You replaced this block with `this.guild = data.guild ? client.guilds.get(data.guild.id) : undefined;`. If the guild is cached, that's fine. If the guild isn't cached, the new code would result in `undefined`. The old code would result in a partial guild-like object, which is better than `undefined`. Useful especially for invites from REST.

Also, having to check if it's a guild-like object or `Guild` instance is annoying. I'd prefer it to stay as-is.

Skillz4Killz: Done! Removed changes to this file.

abalabahaha (approved):

PR #626, opened 2020-01-19
Convert all Message<this> to Message for now

eritbh:
Partial revert of #621, ref https://github.com/microsoft/TypeScript/issues/36307

#621 introduced some strange behavior that nobody really expected before it was tested more thoroughly. The way Typescript handles this particular situation is inaccurate and results in the `Promse` returned by these methods being weirdly broken (`.then` not being callable because of conflicting union types).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,13 +40,13 @@ declare namespace Eris {
 
   interface Textable {
     lastMessageID: string;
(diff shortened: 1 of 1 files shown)


abalabahaha (approved):

abalabahaha:
:/

PR #631, opened 2020-01-22
Delete single reaction

Skillz4Killz:
This PR adds the functionality to delete all reactions for a single emoji.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -594,6 +594,23 @@ class Shard extends EventEmitter {
                 });
                 break;
             }
+            case "MESSAGE_REACTION_REMOVE_EMOJI": {
+                const channel = this.client.getChannel(packet.d.channel_id);
+                if(!channel) {
+                    break;
+                }
+                /**
+                * Fired when someone removes all reactions from a message for a single emoji
+                * @event Client#messageReactionRemoveAllForEmoji
+                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. No other property is guaranteed
+                * @prop {Object} emoji The emoji object with a `name` prop. If the emoji is a custom emoji it will also have an `id` prop.
+                */
+                this.emit("messageReactionRemoveAllForEmoji", channel.messages.get(packet.d.message_id) || {
+                    id: packet.d.message_id,
+                    channel: channel
+                }, packet.d.emoji);
+                break;
+            }
             case "GUILD_MEMBER_ADD": {
                 const guild = this.client.guilds.get(packet.d.guild_id);
                 if(!guild) { // Eventual Consistency™ (╯°□°）╯︵ ┻━┻
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -601,11 +601,11 @@ class Shard extends EventEmitter {
                 }
                 /**
                 * Fired when someone removes all reactions from a message for a single emoji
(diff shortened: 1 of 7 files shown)


Khaaz:
ref: discordapp/discord-api-docs#1309

lib/gateway/Shard.js
```diff
@@ -594,6 +594,23 @@ class Shard extends EventEmitter {
...
+                if(!channel) {
+                    break;
+                }
+                /**
+                * Fired when someone removes all reactions from a message for a single emoji
+                * @event Client#messageReactionRemoveAllForEmoji
+                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. No other property is guaranteed
+                * @prop {Object} emoji The emoji object with a `name` prop. If the emoji is a custom emoji it will also have an `id` prop.
+                */
+                this.emit("messageReactionRemoveAllForEmoji", channel.messages.get(packet.d.message_id) || {
```
Brayzure: Event name should probably be `messageReactionRemoveEmoji`. Not as wordy, and it matches the payload event name

lib/structures/Message.js
```diff
@@ -333,6 +333,15 @@ class Message extends Base {
         return this._client.removeMessageReactions.call(this._client, this.channel.id, this.id);
     }
 
+    /**
+    * Remove all reactions from a message for a single emoji
+    * @arg {String} reaction The reaction (Unicode string if Unicode emoji, `emojiName:emojiID` if custom emoji)
+    * @returns {Promise}
+    */
+    removeReactionsForEmoji(reaction) {
```
Brayzure: This method name should match what's present in the client.

Brayzure (changes requested):

lib/structures/Message.js
Skillz4Killz: @Brayzure This is the same name following the consistency in the other methods. The other methods also remove `Message` from their names so i had kept it that way as well.

Brayzure: Oh true, I do see that.

Brayzure (dismissed):

lib/structures/Message.js
abalabahaha: Seems to be missing typings

lib/rest/Endpoints.js
```diff
@@ -12,6 +12,7 @@ module.exports.CHANNEL_INVITES =                                        (chanID)
 module.exports.CHANNEL_MESSAGE_REACTION =              (chanID, msgID, reaction) => `/channels/${chanID}/messages/${msgID}/reactions/${reaction}`;
 module.exports.CHANNEL_MESSAGE_REACTION_USER = (chanID, msgID, reaction, userID) => `/channels/${chanID}/messages/${msgID}/reactions/${reaction}/${userID}`;
 module.exports.CHANNEL_MESSAGE_REACTIONS =                       (chanID, msgID) => `/channels/${chanID}/messages/${msgID}/reactions`;
+module.exports.CHANNEL_MESSAGE_REACTIONS_EMOJI =          (chanID, msgID, emoji) => `/channels/${chanID}/messages/${msgID}/reactions/${emoji}`;
```
abalabahaha: This duplicates line 12

abalabahaha (changes requested):

Khaaz:
Apparently you changed indentation of some files bringing this PR to 4k diff

abalabahaha (approved):

PR #632, opened 2020-01-23
remove optional

Skillz4Killz:
This PR removes the optional type from voiceMembers as it is always present as an empty collection or a collection with members.


Brayzure (approved):
Makes sense. The property is always set when the object is initialized.

PR #633, opened 2020-01-24
align activity type with docs

alex-taxiera:
- Removed RichActivity since discord does not have this
- Added/fixed props found in Discord docs
  - emoji (new)
  - secrets (new)
  - party (added to)
- ~~Added props found in practice and marked which props are undocumented~~
  - ~~session_id~~
- Removed undocumented props

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -120,11 +120,6 @@ declare namespace Eris {
     url?: string;
   }
   interface Activity extends ActivityPartial<ActivityType> {
-    // start undocumented props
-    id: string;
-    sync_id?: string;
-    session_id?: string;
-    // end undocumented props
     created_at: number;
     timestamps?: { start: number; end?: number };
     application_id?: string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -130,7 +130,7 @@ declare namespace Eris {
     application_id?: string;
     details?: string;
     state?: string;
-    emoji?: { name: string; id?: string; animated?: boolean }
+    emoji?: { name: string; id?: string; animated?: boolean };
     party?: { id?: string; size?: [number, number] };
     assets?: {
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -120,32 +120,32 @@ declare namespace Eris {
     url?: string;
   }
   interface Activity extends ActivityPartial<ActivityType> {
+    // start undocumented props
     id: string;
+    sync_id?: string;
+    session_id?: string;
+    // end undocumented props
```
eritbh: Why do these need to be in the typings if they're undocumented?

eritbh (commented):
huh, were `Activity` and `RichActivity` just never used on their own anywhere? in any case doesn't look like you did anything stupid, just curious about one thing

alex-taxiera:
Yeah the API doesn't actually indicate rich vs non, just sometimes there are more props.

index.d.ts
alex-taxiera: Good point, so far I've found id is incredibly useful when used in conjunction with created_at. When you start and stop you get a fresh id with the new timestamp. Can't speak much to the other ones.

eritbh: hm, what info do you get from `id` that you don't with `created_at`? I think in general we should probably avoid putting undocumented things in the typings.

alex-taxiera: You could *probably* use created_at as the id, but there is technically a case where two presences are added at the same time having the same created_at. We can still remove them though and for this case I can just use some custom types.

alex-taxiera: Removed in 8e1c4eb

eritbh (approved):

abalabahaha (approved):

Skillz4Killz (approved):

PR #635, opened 2020-01-24
shard status can be "resume" & explicitly defining shard statuses

DonovanDMC:


DonovanDMC:
I feel like changing `status` to `state` would make more sense, when I think of "status", I think online/idle/dnd/offline. When I think of state, I think of disconnected/resuming/ready, etc.

Brayzure:
I don't think changing `status` to `state` is the right idea here. The change is minor at best, while at the same time would be a breaking change for anyone who uses that property.

Brayzure (changes requested):
Should revert the property name change for now. It's breaking for anyone who uses the property, and it kinda goes outside of the scope of the PR.

apacheli:
@Brayzure it's been 2 months and this PR hasn't been edited. OP also opened another PR like a week ago, so clearly they don't really care about the breaking change.

DonovanDMC:
Just because something doesn't get touched for a while, doesn't mean people don't care about it

LJNeon:
When a change is requested and nothing is done for months, then work is done by the same person elsewhere, I would assume as much. Also, do you plan on making the requested changes? Your comment doesn't make that clear.

Brayzure:
I asked you about this on Discord two weeks ago, and didn't get a response. I'd rather not have PRs in limbo for two months.

DonovanDMC:
I don't see mentions from those servers as I have them shrouded in folders. As far as the "out of scope" bit goes, I agree, that was out of scope for this pull request.

abalabahaha:
Still needs to undo the `status` to `state` breaking change

Brayzure (approved):

PR #644, opened 2020-02-27
Handle more rest errors

ak-s:
- Stream errors must be handled to prevent crash. E.g. abort while receiving a response may cause "zlib methods will throw an error when decompressing truncated data."
- Catch _createMessage()_ reject while senging _errorMessage_.


abalabahaha (approved):
This could have been 2 PRs, oh well

PR #645, opened 2020-02-28
Fix: permissionCheck() verification

andromika:
permissionCheck() is async, returning a "Promise {Pending}" Object when checked. this returns a falsy positive state.


eritbh:
This was already fixed on the dev branch with #628. Also, PRs should be targeted at the dev branch rather than master.

PR #648, opened 2020-03-04
Update Client#createGuild

LJNeon:
Updated `Client#createGuild` as it had [many missing options](https://discordapp.com/developers/docs/resources/guild#create-guild). I'd like feedback on the new documentation and typings if possible.

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -1387,8 +1387,8 @@ class Client extends EventEmitter {
 
     /**
     * Create a guild
+    * @arg {String} name The name of the guild
     * @arg {Object} options The properties of the guild
-    * @arg {String} options.name The name of the guild
     * @arg {String} [options.region] The region of the guild
     * @arg {String} [options.icon] The guild icon as a base64 data URI. Note: base64 strings alone are not base64 data URI strings
     * @arg {Number} [options.verificationLevel] The guild verification level
@@ -1402,6 +1402,10 @@ class Client extends EventEmitter {
     * @returns {Promise<Guild>}
     */
     createGuild(name, options) {
+        if(this.guilds.size > 9) {
+            throw new Error("This method can't be used when in 10 or more guilds.")
+        }
+
         return this.requestHandler.request("POST", Endpoints.GUILDS, true, {
             name,
             region: options.region,
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -1387,16 +1387,33 @@ class Client extends EventEmitter {
 
     /**
     * Create a guild
-    * @arg {String} name The name of the guild
-    * @arg {String} region The region of the guild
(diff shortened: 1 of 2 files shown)


lib/Client.js
```diff
@@ -1387,16 +1387,33 @@ class Client extends EventEmitter {
 
     /**
     * Create a guild
-    * @arg {String} name The name of the guild
-    * @arg {String} region The region of the guild
-    * @arg {String} [icon] The guild icon as a base64 data URI. Note: base64 strings alone are not base64 data URI strings
+    * @arg {Object} options The properties of the guild
+    * @arg {String} options.name The name of the guild
```
Khaaz: You documented name as a part of options object, while you used name as the first param

Khaaz (changes requested):

lib/Client.js
```diff
@@ -1387,16 +1387,33 @@ class Client extends EventEmitter {
...
+    * @arg {String} [options.region] The region of the guild
+    * @arg {String} [options.icon] The guild icon as a base64 data URI. Note: base64 strings alone are not base64 data URI strings
+    * @arg {Number} [options.verificationLevel] The guild verification level
+    * @arg {Number} [options.defaultNotifications] The default notification settings for the guild. 0 is "All Messages", 1 is "Only @mentions".
+    * @arg {Number} [options.explicitContentFilter] The level of the explicit content filter for messages/images in the guild. 0 disables message scanning, 1 enables scanning the messages of members without roles, 2 enables scanning for all messages.
+    * @arg {String} [options.systemChannelID] The ID of the system channel
+    * @arg {String} [options.afkChannelID] The ID of the AFK voice channel
+    * @arg {Number} [options.afkTimeout] The AFK timeout in seconds
+    * @arg {Array<Object>} [options.roles] The new roles of the guild, the first one is the @everyone role. IDs are placeholders which allow channel overwrites.
```
Khaaz, line 1400: Maybe we could document the structure of Object? At least a simple explanation in the parameter description

lib/Client.js
```diff
@@ -1387,16 +1387,33 @@ class Client extends EventEmitter {
 
     /**
     * Create a guild
```
Khaaz, line 1389: I'd also suggest to specify that this method is only usable if the bot is in least than a certain amount of guilds

lib/Client.js
LJNeon: Documenting the structure would be equivalent to documenting entire role/channel objects, which seems excessive. I did add the full structure in typings at least.

Khaaz: That's fair, I just thought about mentioning that the structure of the Object is type of Channel or Role.

Khaaz (approved):

Khaaz (approved):

abalabahaha (approved):

PR #649, opened 2020-03-04
Implement Allowed Mentions

LJNeon:
Implemented [Allowed Mentions](https://discordapp.com/developers/docs/resources/channel#allowed-mentions-object). Since this functionality entirely covers the use cases of `disableEveryone`, I replaced it entirely. Feedback and testing for this PR would be appreciated.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -372,7 +372,7 @@ declare namespace Eris {
     avatarURL?: string;
     tts?: boolean;
     wait?: boolean;
-    disableEveryone?: boolean;
+    allowedMentions?: AllowedMentions;
   }
 
   interface EmbedAuthorOptions {
@@ -545,11 +545,16 @@ declare namespace Eris {
     [key: string]: {};
   }
 
+  interface AllowedMentions {
+    everyone?: boolean;
+    roles?: boolean | string[];
+    users?: boolean | string[];
+  }
   type MessageContent = string | AdvancedMessageContent;
   type AdvancedMessageContent = {
     content?: string;
     tts?: boolean;
-    disableEveryone?: boolean;
+    allowedMentions?: AllowedMentions;
     embed?: EmbedOptions;
     flags?: number;
   }
@@ -682,7 +687,7 @@ declare namespace Eris {
     compress?: boolean;
     connectionTimeout?: number;
     disableEvents?: { [s: string]: boolean };
-    disableEveryone?: boolean;
+    allowedMentions?: AllowedMentions;
     firstShardID?: number;
     getAllUsers?: boolean;
     guildCreateTimeout?: number;
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -112,7 +112,10 @@ class Client extends EventEmitter {
             defaultImageFormat: "jpg",
             defaultImageSize: 128,
             disableEvents: {},
-            allowedMentions: undefined,
+            allowedMentions: {
+                users: true,
+                roles: true
+            },
             firstShardID: 0,
             getAllUsers: false,
             guildCreateTimeout: 2000,
@@ -131,6 +134,7 @@ class Client extends EventEmitter {
             maxReconnectAttempts: Infinity,
             reconnectDelay: (lastDelay, attempts) => Math.pow(attempts + 1, 0.7) * 20000
         }, options);
+        this.options.allowedMentions = this._formatAllowedMentions(this.options.allowedMentions);
         if(this.options.lastShardID === undefined && this.options.maxShards !== "auto") {
             this.options.lastShardID = this.options.maxShards - 1;
         }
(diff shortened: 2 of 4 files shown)


index.d.ts
```diff
@@ -545,11 +545,16 @@ declare namespace Eris {
...
+    everyone?: boolean;
+    roles?: boolean | string[];
+    users?: boolean | string[];
+  }
   type MessageContent = string | AdvancedMessageContent;
   type AdvancedMessageContent = {
     content?: string;
     tts?: boolean;
-    disableEveryone?: boolean;
+    allowedMentions?: AllowedMentions;
```
Skillz4Killz, line 557: I think it would be nice to continue using this option by implementing this option to disable everyone and here mentions but allow user and role mentions by default. I do understand that using allowedMentions is more granular but it makes this into a breaking change.

lib/Client.js
```diff
@@ -623,22 +626,42 @@ class Client extends EventEmitter {
...
     * @returns {Promise<Message?>}
     */
     executeWebhook(webhookID, token, options) {
         if(!options.content && !options.file && !options.embeds) {
             return Promise.reject(new Error("No content, file, or embeds"));
         }
-        if(options.content && (options.disableEveryone !== undefined ? options.disableEveryone : this.options.disableEveryone)) {
-            options.content = options.content.replace(/@everyone/g, "@\u200beveryone").replace(/@here/g, "@\u200bhere");
+        if(options.allowedMentions || this.options.allowedMentions) {
+            const allowed = options.allowedMentions ? options.allowedMentions : this.options.allowedMentions;
```
Skillz4Killz: Move this line outside the if statement and then the if can be as simple as `if (allowed)` making it slightly less code.

lib/Client.js
```diff
@@ -623,22 +626,42 @@ class Client extends EventEmitter {
...
+            options.allowed_mentions = {
+                parse: []
+            };
+            if(allowed.everyone) {
+                options.allowed_mentions.parse.push("everyone");
+            }
+            if(allowed.roles === true) {
+                options.allowed_mentions.parse.push("roles");
+            }else if(Array.isArray(allowed.roles)) {
+                options.allowed_mentions.roles = allowed.roles;
```
Skillz4Killz: Maybe consider forcing this length to max of 100

lib/Client.js
```diff
@@ -623,22 +626,42 @@ class Client extends EventEmitter {
...
+            }
+            if(allowed.users === true) {
+                options.allowed_mentions.parse.push("users");
+            }else if(Array.isArray(allowed.users)) {
+                options.allowed_mentions.users = allowed.users;
```
Skillz4Killz: Limit this to 100 users

lib/Client.js
```diff
@@ -623,22 +626,42 @@ class Client extends EventEmitter {
...
+    * @arg {Boolean | Array<String>} [options.allowedMentions.users] Whether or not to allow all user mentions, or an array of specific user mentions to allow.
     * @returns {Promise<Message?>}
     */
     executeWebhook(webhookID, token, options) {
         if(!options.content && !options.file && !options.embeds) {
             return Promise.reject(new Error("No content, file, or embeds"));
         }
-        if(options.content && (options.disableEveryone !== undefined ? options.disableEveryone : this.options.disableEveryone)) {
-            options.content = options.content.replace(/@everyone/g, "@\u200beveryone").replace(/@here/g, "@\u200bhere");
+        if(options.allowedMentions || this.options.allowedMentions) {
```
Skillz4Killz: Consider having this if run once for the `this.options.allowedMentions` as having this run every single time we send this is not ideal. Just set it up once and save it somewhere and reuse.

Skillz4Killz (changes requested):

lib/Client.js
Skillz4Killz: Also i don't think you need a ternary here. It could just be `alloweMentions || xxxx`

lib/Client.js
Skillz4Killz: Note: these also apply to the message method not just the execute webhook one

index.d.ts
LJNeon: To clarify, you're suggesting default behavior be kept, not keeping the disableEveryone option itself?

lib/Client.js
LJNeon: Forcing the length to a max of 100 would silently cause unexpected behavior when users attempt to use more than 100. I'd prefer it to error so they realize the problem immediately, forcing the length seems harmful to me.

index.d.ts
Skillz4Killz: if disableEveryone is true default allowedMention => 
```js
{
  parse: ["roles", "users"]
}
```

lib/Client.js
Skillz4Killz: Error is fine as well

lib/Client.js
```diff
@@ -2071,6 +2076,36 @@ class Client extends EventEmitter {
         }));
     }
 
+    _formatAllowedMentions(allowed) {
+        if(allowed) {
```
Skillz4Killz: You could remove nesting by making this reverse
`if (!allowed) return this.options.allowedMentions`

lib/Client.js
```diff
@@ -2071,6 +2076,36 @@ class Client extends EventEmitter {
...
+                result.parse.push("users");
+            } else if(Array.isArray(allowed.users)) {
+                if(allowed.roles.length > 100) {
+                    throw new Error("Allowed user mentions cannot exceed 100.");
+                }
+                result.users = allowed.users;
+            }
+            return result;
+        } else if(this.options.allowedMentions) {
+            return this.options.allowedMentions;
```
Skillz4Killz: I think this might break if im not mistaken since you default `roles: true` and `users: `true` instead of `parse: ['roles', 'users']`

LJNeon: You're mistaken, when they're `true` they're added to the `parse` array.

lib/Client.js
LJNeon: I don't see any benefit in doing this, this seems like a stylistic suggestion, and not one that matches the style eris uses.

Khaaz (approved):
Looking good! That's also how I would have done it. 
As this feature is due to a breaking change in Discord API I'm also inclined to entirely remove disableEveryone.

However if we want to still support legacy behavior, we could manually add `"everyone"` if they still pass disableEveryone and it is explicitely set to false.
I personally don't think we should do this as, at worst it will just suppress everyone mention for `everyone` and this feature looks like a big change in DAPI and can probably also be for users wanting to use last eris version.

Khaaz:
Also, this ref discordapp/discord-api-docs#1396

abalabahaha (approved):

PR #651, opened 2020-03-04
Correctly document all USER ACCOUNT methods

Khaaz:
This PR correctly document user account methods. Some method were missing the flag in the documentation.  
To make it even clearer I changed it by adding `[USER ACCOUNT]` at the beginning of the method description directly.

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -879,7 +879,7 @@ class Client extends EventEmitter {
     }
 
     /**
-    * Accept an invite (not for bot accounts)
+    * [USER ACCOUNT] Accept an invite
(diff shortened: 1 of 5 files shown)


Skillz4Killz:
I still am not sure why something that is confirmed as only for user accounts is available. All of these should be removed imo.

Khaaz:
I think it has already been discussed that removing user support woud be too much work considering the lib design. Deleting these methods alone wouldn't be useful anyway
This PR just has for purpose to uniformise user-account methods and making it explicit.

abalabahaha (approved):

PR #653, opened 2020-03-04
Remove unusuable methods from Private channel (removeMessageReactionEmoji and removeMessageReactions)

Khaaz:
This PR removes `removeMessageReactionEmoji` and `removeMessageReactions` from PrivateChannel as they seem to not be callable in PrivateChannel at all.
Trying to use these methods results in a direct error from the API: 
` DiscordRESTError [50003]: Cannot execute action on a DM channel`

As for the typings, deleting these methods directly seems to not be possible because `PrivateChannel` wouldn't be implementing `TextableChannel` entirely.
I'd like review on the best approach to do this.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -59,8 +59,6 @@ declare namespace Eris {
     ): Promise<User[]>;
     addMessageReaction(messageID: string, reaction: string, userID?: string): Promise<void>;
     removeMessageReaction(messageID: string, reaction: string, userID?: string): Promise<void>;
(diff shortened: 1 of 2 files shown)


Khaaz:
Edit: Updated the typings (https://discordapp.com/channels/81384788765712384/381890266052689920/685623685968691243)
Removed from `Textable` and implemented in `GuildTextable` instead

alex-taxiera (approved):

abalabahaha (approved):

PR #654, opened 2020-03-06
Adds @types/node to dev dependencies

Khaaz:
This PR just adds \@types/node to dev dependency for `index.d.ts`.


LJNeon:
+1 on this, I've noticed parsing errors for objects like Buffers without it.

abalabahaha (approved):

PR #656, opened 2020-03-11
Prefix _token to emphasize it is private

Khaaz:
This PR emphasize that token should be private by prefixing it as `Client._token`.

**Goal**
The main purpose is to emphasize that token is private and read only. It is assigned at Client instantiation and should never be accessible for the user later.
This is mostly sematic, but in my opinion token should be private by nature.

**Side effects**
- Since token is private, it should be Client responsability to prefix the token with `"Bot "`. This is enforced by adding a `_validateToken` method in Client that will do that if necessary.
Before, both Shards and Client would prefix the token with `"Bot"`. Now it's Client responsability only.
- Since the token is directly used to identify to the gateway or for REST request, I added a getter. This makes token readOnly via `Client.token`.

**Additional benefits**
This will also hide the token from logging via `util.inspect` since Client automatically hide prefixed properties (https://github.com/abalabahaha/eris/blob/master/lib/structures/Base.js#L42)
While this is not the main goal of this PR this is a cool addition.

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -187,6 +187,12 @@ class Client extends EventEmitter {
         return this.startTime ? Date.now() - this.startTime : 0;
     }
 
+    _validateToken() {
+        if(!this._token.startsWith("Bot ")) {
+            this._token = "Bot " + this._token;
+        }
+    }
+
     /**
     * Tells all shards to connect.
     * @returns {Promise} Resolves when all shards are initialized
@@ -247,9 +253,7 @@ class Client extends EventEmitter {
     * @returns {Promise<Object>} Resolves with an object containing gateway connection info
     */
     getBotGateway() {
-        if(!this._token.startsWith("Bot ")) {
-            this._token = "Bot " + this._token;
-        }
+        this._validateToken();
         return this.requestHandler.request("GET", Endpoints.GATEWAY_BOT, true);
     }
 
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -47,7 +47,7 @@ const sleep = (ms) => new Promise((res) => setTimeout(res, ms));
 /**
 * Represents the main Eris client
 * @extends EventEmitter
-* @prop {String} token The bot user token
+* @prop {String} _token The bot user token
 * @prop {Boolean?} bot Whether the bot user belongs to an OAuth2 application
 * @prop {Object} options Eris options
 * @prop {Object} channelGuildMap Object mapping channel IDs to guild IDs
@@ -147,7 +147,7 @@ class Client extends EventEmitter {
             this.options.ws.agent = this.options.agent;
         }
 
-        this.token = token;
+        this._token = token;
 
         this.requestHandler = new RequestHandler(this);
 
@@ -177,6 +177,10 @@ class Client extends EventEmitter {
         this.connect = this.connect.bind(this);
         this.lastReconnectDelay = 0;
         this.reconnectAttempts = 0;
+    }
+
+    get token() {
+        return this._token;
     }
 
     get uptime() {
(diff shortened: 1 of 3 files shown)


LJNeon:
If the mentality here is that the token should be private, I'd think it should be removed from documentation and typings as well.

Khaaz:
> If the mentality here is that the token should be private, I'd think it should be removed from documentation and typings as well.  

I kept it there because of the getter. But that makes sense, removed it from typings and Client documentation as well!

LJNeon (commented):
```js
if(!this.client.token.startsWith("Bot ")) {
    this.client.token = "Bot " + this.client.token;
}
```
I would think you could simply move this if statement to the constructor, instead of using a function to avoid repetition.

lib/Client.js
```diff
@@ -179,10 +178,20 @@ class Client extends EventEmitter {
         this.reconnectAttempts = 0;
     }
 
+    get token() {
```
LJNeon, line 210: This getter seems entirely useless outside of backwards compatibility, but I see no real use for that here.

Khaaz: The conversation we had about that: <https://discordapp.com/channels/81384788765712384/381890266052689920/687247603381043210>

I have personally no issue with using `client._token` in Shard and RequestHandler instead of the getter, but maybe other point of view would be nice before changing the implementation?
I'll end up doing what Eljay suggested to simplify the code unless other poeple vote against that.

apacheli:
Maybe you could use `Object.defineProperty` to hide it by setting `enumerable: false`

Khaaz:
> Maybe you could use `Object.defineProperty` to hide it by setting `enumerable: false`

It looks like a valid idea, however seems over the top considering this would make `_token` the only property to be set this way.
I'd let this as it is, and if we really want to do what you suggested, I'd do it for every private property in the framework then

LJNeon:
Actually I believe Eris does this in the error classes. I'm also in favor of doing this mainly because it makes this PR no longer a breaking change while achieving the same goal.

Khaaz:
The original purpose of this PR is that token should semantically be private (1). To follow the lib standard I prefixed it as `_token`.
This has for additional benefits to also hide it from `util.inspect` (2).

I don't personally think no longer having `token` available is an issue because I doubt anyone uses it. However for backward compatibility and to be able to use `client.token` outside of the class, I added a getter.
As Eljay suggested, we could also use `client._token` directly on Shard as both class could be considered internals. In that case we can even remove the getter if we don't care about backward compatibility for `client.token`.

Now, as suggested by Apacheli, we can simply make token non enumerable instead of making it entirely private. This achieve (2) but doesn't do anything for (1).
This also would be the only place outside of Error class where we do such thing.
While I think this solution is at least what we should do, I still think making the token entirely private is not meaningless.

TLDR:
Do we entirely make `client.token` semantically private as `client._token` (in this case, do we enable to use `client._token` in Shard), or do we just make `client.token` non enumerable?

lib/Client.js
```diff
@@ -179,10 +178,20 @@ class Client extends EventEmitter {
...
+        return this._token;
+    }
+
     get uptime() {
         return this.startTime ? Date.now() - this.startTime : 0;
     }
 
+    _validateToken() {
```
Brayzure: Probably should be renamed. To me, `_validateToken` implies it's checking for proper format without making any changes. Perhaps `_formatToken`?

lib/Client.js
Brayzure: Seeing as `Client#token` is documented, it should remain in some form, regardless of your beliefs on how strictly we should adhere to its private nature.

exoRift:
I believe Abal mentioned he'd like to remove automatic prefixing and leave it to the user to do.

abalabahaha:
Alternative implementation in 99f4163, reflecting further discussion in Discord

PR #657, opened 2020-03-12
Minor typing fixes

bumblecado021304:
This PR introduces 2 missing typings of properties `guildPrefixes` and `commandAliases` in the CommandClient class.
It also replaces the `Member | undefined` typing of a `member` property in a Message class with `Member | null`, as, in case of absence of `Member`, `null` is thrown, not `undefined`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1533,7 +1533,7 @@ declare namespace Eris {
     timestamp: number;
     type: number;
     author: User;
-    member?: Member;
+    member: Member | null;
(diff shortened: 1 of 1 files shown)


Khaaz:
You PR target master. You need to changes that to point to dev branch.

bumblecado021304:
> You PR target master. You need to changes that to point to dev branch.

Thank you, I will target this PR to dev branch instead.

PR #663, opened 2020-03-17
Allow Role input for createRole

LJNeon:
There's only one inconsistency preventing Role objects from being passed to `Client#createRole()`, and being able to pass a Role object is relatively useful when cloning or recreating Roles. This change also explicitly passes options to discord which could potentially prevent user errors.

For example, without this change in order to recreate a Role you would need:
```js
await oldRole.delete();
const newRole = await guild.createRole({
  name: oldRole.name,
  permissions: oldRole.permissions.allow,
  color: oldRole.color,
  hoist: oldRole.hoist,
  mentionable: oldRole.mentionable,
});
await newRole.editPosition(oldRole.position);
```
However this PR would reduce that code to:
```js
await oldRole.delete();
const newRole = await guild.createRole(oldRole);
await newRole.editPosition(oldRole.position);
```

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -746,7 +746,7 @@ class Client extends EventEmitter {
     /**
     * Create a guild role
     * @arg {String} guildID The ID of the guild to create the role in
-    * @arg {Object|Role} [options] The properties to set or a Role instance
+    * @arg {Object|Role} [options] An object or Role containing the properties to set
     * @arg {String} [options.name] The name of the role
     * @arg {Number} [options.permissions] The role permissions number
     * @arg {Number} [options.color] The hex color of the role, in number form (ex: 0x3d15b3 or 4040115)
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -746,7 +746,7 @@ class Client extends EventEmitter {
     /**
     * Create a guild role
     * @arg {String} guildID The ID of the guild to create the role in
-    * @arg {Object} [options] The properties to set
+    * @arg {Object|Role} [options] The properties to set or a Role instance
     * @arg {String} [options.name] The name of the role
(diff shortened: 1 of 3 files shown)


apacheli:
Perhaps we could change `Role#permission` into a number. This would be a breaking change, but I don't believe it would affect the use case of most end users.

LJNeon:
I would think getting role permissions is a relatively common use case. It also removes all the convenience that the `Permission` class provides which seems like a bad idea.

Skillz4Killz:
> I would think getting role permissions is a relatively common use case. It also removes all the convenience that the `Permission` class provides which seems like a bad idea.

How so?

LJNeon:
All it'd be doing is removing ease of use for practically no benefit, this doesn't seem worth it at all just because it's not one of the most common use cases.

lib/Client.js
```diff
@@ -756,8 +756,14 @@ class Client extends EventEmitter {
     * @returns {Promise<Role>}
     */
     createRole(guildID, options, reason) {
-        options.reason = reason;
-        return this.requestHandler.request("POST", Endpoints.GUILD_ROLES(guildID), true, options).then((role) => {
+        return this.requestHandler.request("POST", Endpoints.GUILD_ROLES(guildID), true, {
+            name: options.name,
+            permissions: options instanceof Role ? options.permissions.allow : options.permissions,
```
abalabahaha: Might as well just run `instanceof Permission` here

abalabahaha (changes requested):

lib/Client.js
LJNeon: Done.

Khaaz (dismissed):

abalabahaha (approved):

PR #664, opened 2020-03-18
Fix TypeError bug in Command

apacheli:
Fixes a bug where setting `Command.requirements.permissions.*` would result in an error in DM contexts.


apacheli:
With this in mind, commands with permission requirements can still be executed in DMs. This behavior could seem unnatural, so it *could* change in the future.

abalabahaha (approved):

PR #667, opened 2020-03-18
Add ms precision ratelimit

Khaaz:
This PR adds support for ms precision ratelimit which was for some reason never added in eris.
I believe it should work out of the box looking at the format discord uses, documented [here](https://discordapp.com/developers/docs/topics/rate-limits#more-precise-rate-limit-resets).

### DIFF
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -76,7 +76,8 @@ class RequestHandler {
             const actualCall = (cb) => {
                 const headers = {
                     "User-Agent": this.userAgent,
-                    "Accept-Encoding": "gzip,deflate"
+                    "Accept-Encoding": "gzip,deflate",
+                    "X-RateLimit-Precision": "millisecond"
                 };
                 let data;
                 let finalURL = url;


Skillz4Killz:
I don't personally use the rate limit headers but if someone is using it this is a breaking change. Perhaps make this a client option to use milliseconds instead of seconds

Brayzure:
>I don't personally use the rate limit headers

Yes you do, headers are consumed automatically to provide ratelimiting for the library, instead of hardcoding ratelimits.

Brayzure (dismissed):
You need to adjust/remove the override in place for reaction endpoints, and make sure they work afterwards. Eris modifies ratelimit info for this endpoint, since the ratelimit used to require precision that the API didn't provide.

Khaaz:
Actually, I am not sure what eris is doing here exactly: <https://github.com/abalabahaha/eris/blob/dev/lib/rest/RequestHandler.js#L231>

Testing with the following simple script:
```js
(async() => {
  for (let i = 0 ; i < 10 ; i++) {
    await this.bot.addMessageReaction(msg.channel.id, msg.id, '👀')
    await this.bot.removeMessageReaction(msg.channel.id, msg.id, '👀', this.bot.user.id)
  }
})()
```
It works fine with the override or without the override, with or without ms precision.

According to discord:
> Routes for controlling emojis do not follow the normal rate limit conventions. These routes are specifically limited on a per-guild basis to prevent abuse. This means that the quota returned by our APIs may be inaccurate, and you may encounter 429s.

What exactly was eris doing for emoji that would no longer need to be done if we use ms precision?

Brayzure:
As far as I can tell, Eris was manually deducing the proper reset time for the endpoint, instead of using the `x-ratelimit-reset` header. Without the override, it should still work, but may produce 429s, or be slower than 4 per second.

abalabahaha (approved):
Discord rounds the reactions endpoint reset time upward (0.25s => 1s) when using second precision, so I forced it to 250. Should work fine though.

PR #668, opened 2020-03-21
feat: add safeRequire function

ghost:
This pull-request focuses on supplementing a one-for-all function to avoid repetition, enable readability, and for future use.

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -21,29 +21,37 @@ const ShardManager = require("./gateway/ShardManager");
 const UnavailableGuild = require("./structures/UnavailableGuild");
 const User = require("./structures/User");
 const VoiceConnectionManager = require("./voice/VoiceConnectionManager");
-const safeRequire = (dependency, fallbackDependency = null) => {
+const safeRequire = (dep, fallbackDep = null) => {
     try {
-        return require(dependency);
+        return require(dep);
     } catch(err) {
-        if(fallbackDependency !== null) {
-            return require(fallbackDependency);
+        if(fallbackDep !== null) {
+            return require(fallbackDep);
         }
     }
 };
 
+const safeRequire = (dep, fallbackDep = null, webRequire = false) => {
+    try {
+      return require(dep);
+    } catch (err) {
+      if (fallbackDep !== null) {
+        if (webRequire === true) {
+          try {
+            return require(fallbackDep);
+          } catch (err) { // eslint-disable no-empty
+          }
+        } else {
+          return require(fallbackDep);
+        }
+      }
+    }
+  };
+
 const EventEmitter = safeRequire("eventemitter3", "events");
 const Erlpack = safeRequire("erlpack");
+const ZlibSync = safeRequire("zlib-sync", "pako", true);
 const sleep = (ms) => new Promise((res) => setTimeout(res, ms));
-
-let ZlibSync;
-try {
-    ZlibSync = require("zlib-sync");
-} catch(err) {
-    try {
-        ZlibSync = require("pako");
-    } catch(err) { // eslint-disable no-empty
-    }
-}
 
 /**
 * Represents the main Eris client
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -21,32 +21,7 @@ const ShardManager = require("./gateway/ShardManager");
 const UnavailableGuild = require("./structures/UnavailableGuild");
 const User = require("./structures/User");
 const VoiceConnectionManager = require("./voice/VoiceConnectionManager");
-const safeRequire = (dep, fallbackDep = null) => {
-    try {
-        return require(dep);
-    } catch(err) {
-        if(fallbackDep !== null) {
-            return require(fallbackDep);
-        }
-    }
-};
-
-const safeRequire = (dep, fallbackDep = null, webRequire = false) => {
-    try {
-      return require(dep);
-    } catch (err) {
-      if (fallbackDep !== null) {
-        if (webRequire === true) {
-          try {
-            return require(fallbackDep);
-          } catch (err) { // eslint-disable no-empty
-          }
-        } else {
-          return require(fallbackDep);
-        }
-      }
-    }
-  };
+const {safeRequire} = require("./util/utils");
 
 const EventEmitter = safeRequire("eventemitter3", "events");
 const Erlpack = safeRequire("erlpack");
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -25,7 +25,7 @@ const {safeRequire} = require("./util/utils");
 
 const EventEmitter = safeRequire("eventemitter3", "events");
 const Erlpack = safeRequire("erlpack");
-const ZlibSync = safeRequire("zlib-sync", "pako", true);
+const ZlibSync = safeRequire("zlib-sync", ["pako"]);
 const sleep = (ms) => new Promise((res) => setTimeout(res, ms));
 
 /**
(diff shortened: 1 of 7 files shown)


Khaaz:
The idea is there, but there are some flows. 
- Doing that offers no control about destructuring the require. eg: <https://github.com/abalabahaha/eris/blob/master/lib/gateway/Shard.js#L19>
That makes me wonder why it doesn;t behave the same way in Shard (and other structures) and in Client.
- You can't have multiple layer of require (like for ZLib 2 lines below)
- There are not real benefits in term of readability of doing it only in Client. If we were to implement that, we probably would like to use it every time we require something.
Such dynamic require mechanics also exists in: `gateway/Shard`, `util/Opus`, `voice/Piper`, `voice/SharedStream`,  `voice/VoiceConnection`, `voice/voiceDataStream`.

lib/Client.js
```diff
@@ -21,18 +21,20 @@ const ShardManager = require("./gateway/ShardManager");
 const UnavailableGuild = require("./structures/UnavailableGuild");
 const User = require("./structures/User");
 const VoiceConnectionManager = require("./voice/VoiceConnectionManager");
+const safeRequire = (id, subID = null) => {
```
Khaaz: I would also rename the parameter for something more obvious, like `dependency`, `fallbackDependency` or something like that

ghost: Good idea, done!

ghost:
> * Such dynamic require mechanics also exists in: `gateway/Shard`, `util/Opus`, `voice/Piper`, `voice/SharedStream`,  `voice/VoiceConnection`, `voice/voiceDataStream`.

Currently, there is no global class for convenience and utility methods in the module; thus, I am indecisive about where to group the function. Hence, we can have a class for utility methods where methods will be grouped to enable usage of the methods by requiring the "global class" in whichever file we want. I am proposing that we may have a Util class that will hold all the utility methods/

Khaaz:
Yes, I think we can have a utils file in the util folder. Export the fonction directly, no need for a class with static methods.
To solve the issue of multiple require sequentially, you can hardcode a second check in the function but that's not the most convenient, or you can make it take an infinite number of args and call the function recursively maybe?

There are some inconsistencies with the different `require` fallback so I'm not too sure how doable that is. In some case you would need to only take a property of the fallback (like for EventEmitter), other case, variable assigned in the fallback is not the same as the one assigned if the default works (VoiceConnection: sodiumNative/tweetnacl)

ghost:
> * You can't have multiple layer of require (like for ZLib 2 lines below)

We can create this functionality by creating another parameter, namely "webRequire". Certainly, this will reduce readability. 😅

ghost:
> * Doing that offers no control about destructuring the require.

In May 2013, [this pull-request](https://github.com/nodejs/node/commit/6ed861dd7feea33224d1cac8e64010b45974ae51) was submitted to the [Node.js repository](https://github.com/nodejs/node/):
> This change is 100% backward compatible.
> 
> This change will make using `EventEmitter` slightly simpler / nicer and
> adheres to the best practice set forth by sub-stack.
> 
> ```js
> var EventEmitter = require("events");
> var emitter = new EventEmitter();
> ```
> 
> The only difference is that we now have to set `EventEmitter` as a
> property of `EventEmitter` for backward compatibility like we do with
> [`Stream`][1]
> 
> We have also set the `usingDomains` property on the `EventEmitter`
> constructor itself because that aligns with its current usage of
> `require("events").usingDomains = true`
> 
> There are other internals that would benefit from this change as well
> like `StringDecoder`

This pull-request now allows a simpler syntax while maintaining backward compatibility with the older syntax.

```javascript
const EventEmitter = require('events');
const emitter = new EventEmitter();
```

lib/gateway/Shard.js
```diff
@@ -11,31 +11,17 @@ const {GATEWAY_VERSION, GatewayOPCodes, ChannelTypes} = require("../Constants");
 const ExtendedUser = require("../structures/ExtendedUser");
 const User = require("../structures/User");
 let WebSocket = typeof window !== "undefined" ? window.WebSocket : require("ws");
+const {safeRequire} = require("../util/utils");
+WebSocket = safeRequire("uws");
+const Erlpack = safeRequire("erlpack");
+const ZlibSync = safeRequire("zlib-sync", "pako", true);
 
 let EventEmitter;
 try {
```
Khaaz: You missed eventEmitter here

lib/gateway/Shard.js
```diff
@@ -11,31 +11,17 @@ const {GATEWAY_VERSION, GatewayOPCodes, ChannelTypes} = require("../Constants");
 const ExtendedUser = require("../structures/ExtendedUser");
 const User = require("../structures/User");
 let WebSocket = typeof window !== "undefined" ? window.WebSocket : require("ws");
+const {safeRequire} = require("../util/utils");
+WebSocket = safeRequire("uws");
```
Khaaz: This will cause Websocket to be undefined if `uws` is not installed instead of using the default "ws" as it used to.
```suggestion
WebSocket = safeRequire("uws") || WebSocket;
```

lib/voice/VoiceConnection.js
```diff
@@ -7,23 +7,12 @@ const Dgram = require("dgram");
 const Piper = require("./Piper");
 const VoiceDataStream = require("./VoiceDataStream");
 let WebSocket = typeof window !== "undefined" ? window.WebSocket : require("ws");
+const {safeRequire} = require("../util/utils");
+WebSocket = safeRequire("uws");
```
Khaaz: Same comment as above.
```suggestion
WebSocket = safeRequire("uws") || WebSocket;
```

Khaaz (changes requested):

Khaaz:
> This pull-request now allows a simpler syntax while maintaining backward compatibility with the older syntax.

Alright nice, fixing this inconsistency is also an improvement.
Also, your link to the actual node.js PR doesn't work

lib/gateway/Shard.js
ghost: Done!

Khaaz (approved):

ghost:
> your link to the actual node.js PR doesn't work

😅 Here you go: [pull-request concerning EventEmitter](https://github.com/nodejs/node/commit/6ed861dd7feea33224d1cac8e64010b45974ae51)

Brayzure:
If you're going to go through the trouble to make a function to support two different dependencies, you probably want to go ahead and make it support an arbitrary number. That way, if we want to support a third ZLib library (for whatever reason) or something similar, we don't have to adjust this as well.

ghost:
@Brayzure, please take a look again.

eritbh:
It seems like you've redefined the function but kept the old implementation below it, so it would just get overwritten immediately. It also doesn't look like the calls to the function have been updated for the new implementation.

At this point, I feel like we might just be better off using a package like [node-require-fallback](https://www.npmjs.com/package/node-require-fallback) for this rather than reimplementing.

LJNeon:
Adding a dependency to make the code look slightly nicer doesn't seem worth it at all, I'd suggest just closing this PR instead personally. This also breaks support with webpack, which we don't officially support but it's still nice to have, and those who use it do exist.

Khaaz:
I'm very confused on why this has been closed without been asked by a maintainer.
I don't think it was so useless to have a utility function to do all the require, and I believe we could have had a pretty simple one that supports an arbitrary number of fallback without too much overhead (I disagree about using a dependency as well though, especially when you see it's 10 lines of code).
Last commit was indeed not done correctly, but this could have been corrected and I don't really understand why asking to close the PR

LJNeon:
It would still break webpack support in exchange for slightly nicer looking code, not really something of substance. There's some more conversation in the discord if you want to search message history for it.

PR #669, opened 2020-03-22
Missing reason parameter in TextChannel.deleteMessages and purgeChannel

Sei4or:
The deleteMessages method in TextChannel does not accept a reason parameter while the deleteMessages method in the Client does accept a reason.
Also added reason to purgeChannel.

TextChannel.deleteMessages: https://github.com/abalabahaha/eris/blob/dev/lib/structures/TextChannel.js#L92
Client.deleteMessages:
https://github.com/abalabahaha/eris/blob/master/lib/Client.js#L1182

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1085,7 +1085,8 @@ declare namespace Eris {
       limit?: number,
       filter?: (m: Message) => boolean,
       before?: string,
-      after?: string
+      after?: string,
+      reason?: string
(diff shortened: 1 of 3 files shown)


Khaaz:
You need  to edit typings as well

Khaaz (approved):

abalabahaha (approved):

PR #672, opened 2020-04-02
Fix reaction events when channel not cached

macdja38:
MESSAGE_REACTION_ADD and MESSAGE_REACTION_REMOVE
relied on channels to be cached, which is not the case
with dm channels. This allows users to see reactions
added to user dms if the bot does not have the channel
cached.


Khaaz (approved):

abalabahaha (approved):

JJRcop:
This PR didn't update documentation

PR #897, opened 2020-04-13
Add publicFlags on user and the needed USER_FLAGS

Khaaz:
Adds support for last discord API changes
ref: discord/discord-api-docs#1490 (public_flags) and discord/discord-api-docs#1493 (additional user flags)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -370,6 +370,22 @@ declare namespace Eris {
       GUILD_NEWS: 5;
       GUILD_STORE: 6;
     };
+    UserFlags: {
+      NONE: 0;
+      DISCORD_EMPLOYEE: 1;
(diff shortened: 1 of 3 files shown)


Khaaz:
It appears that some users have public_flags to undefined at startup.
Fetching the user from the API directly gives the user with the actual flag.

Do we want a specific method to fetch flags? Basically what we would simply need is manually fetching the user via REST and updates it in the cache if `public_flags` is undefined.
We can also let that as it is and just document that the public_flags value is undefined and that you will need to manually fetch the user to be sure to have the value in case it's undefined.

Could I get other point of view on the best way to implement that correctly?

abalabahaha (approved):

PR #901, opened 2020-04-15
Add support for crossposting messages

bsian03:
This is not yet documented, however please see [here](https://github.com/discord/discord-api-docs/issues/1510) for more details,
> The endpoint is now accessible to bots, but there’s a few things... which I need to fix up before I’d be comfortable documenting it.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1107,7 +1107,7 @@ declare namespace Eris {
       after?: string,
       reason?: string
     ): Promise<number>;
-    crosspost(channelID: string, messageID: string): Promise<Message>;
+    crosspostMessage(channelID: string, messageID: string): Promise<Message>;
     getGuildEmbed(guildID: string): Promise<GuildEmbed>;
     getGuildIntegrations(guildID: string): Promise<GuildIntegration[]>;
     editGuildIntegration(guildID: string, integrationID: string, options: IntegrationOptions): Promise<void>;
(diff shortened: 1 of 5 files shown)


lib/structures/NewsChannel.js
```diff
@@ -26,6 +26,14 @@ class NewsChannel extends TextChannel {
         this.rateLimitPerUser = 0;
         this.update(data);
     }
+    /**
+     * Publish a message to subscribed channels
+     * @arg {String} messageID The ID of the message
+     * @returns {Promise<Message>}
+     */
+    crosspost(messageID) {
```
Khaaz: This should probably be crosspostMessage as well?

index.d.ts
```diff
@@ -1107,6 +1107,7 @@ declare namespace Eris {
       after?: string,
       reason?: string
     ): Promise<number>;
+    crosspost(channelID: string, messageID: string): Promise<Message>;
```
Khaaz: This should be crosspostMessage as mentionned by cat

lib/structures/NewsChannel.js
```diff
@@ -26,6 +26,14 @@ class NewsChannel extends TextChannel {
...
+        return this.client.crosspost.call(this.client, this.id, messageID);
```
Khaazz: Also since you changed client to be crosspostMessage, the client call should be on `crosspostMessage`. Same goes for the others

Khaaz (dismissed):

abalabahaha (changes requested):
Would prefer docs to say crosspost too, maybe "Crosspost (publish) a message"

abalabahaha (approved):

PR #904, opened 2020-04-15
Fix docs and typings for MESSAGE_REACTION_ADD and MESSAGE_REACTION_REMOVE events

Khaaz:
This correct missing typing and docs update in #672 
Looking at this again, I was also wondering why the same behaviour has not been added to `MESSAGE_REACTION_REMOVE_ALL` and `MESSAGE_REACTION_REMOVE_EMOJI`

I also don't know if I fixed typings the correct way so please, review!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -721,7 +721,7 @@ declare namespace Eris {
     frameSize?: number;
     sampleRate?: number;
   }
-  type PossiblyUncachedMessage = Message | { id: string; channel: TextableChannel };
+  type PossiblyUncachedMessage = Message | { id: string; channel: TextableChannel } | { id: string; channel: { id: string } };
   interface RawPacket {
     op: number;
     t?: string;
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -721,7 +721,7 @@ declare namespace Eris {
     frameSize?: number;
     sampleRate?: number;
   }
-  type PossiblyUncachedMessage = Message | { id: string; channel: TextableChannel };
+  type PossiblyUncachedMessage = Message | { id: string; channel: TextableChannel } | { id: string; channel: { id: string } };
```
```suggestion
  type PossiblyUncachedMessage = Message | { id: string; channel: TextableChannel | { id: string } };
```

Please double check this on VSC i am suggesting this on Github without having my morning cup of tea 🍵

abalabahaha (approved):
Feel free to open a PR for the other events, forgot about those :(

PR #905, opened 2020-04-15
Specifies type of message in channel#messages.

oathompsonjones:


bsian03:
All instances of GuildTextableChannel will need to be changed to TextChannel and NewsChannel (with overrides in NewsChannel as necessary) to line up with #901 just fyi. Obviously depends which one comes first

bsian03 (approved):

abalabahaha (approved):

Issue #898, opened 2020-04-14
Some Guild properties are not documented


acarfy:
in structures/Guild.js

bannerURL and splashURL don't in Guild props. We can't see in documentation.

bsian03:
Think this can be closed now? Not sure why it's still open

anon28nn:
in structures/GuildChannel.js

guild is in properties okey but type don't defined.

PR #910, opened 2020-04-16
Update GuildChannel.js for JSDOC

anon28nn:
for this.guild property. {Guild} type didn't defined i define this.
#898


anon28nn:
Is it true? I don't know use GitHub very well?

bsian03:
This isn't needed imo. This is the same case for every class

PR #914, opened 2020-04-21
Request handled event

Skillz4Killz:
This PR adds an event called `requestHandled`. The name is up for discussion ofcourse. It is just a placeholder.

Purpose: Recently, I had a lot of trouble trying to edit node_module folders and restarting the bot just to be able to listen to which events my bot was triggering. Everytime i wanted to make a tiny change it would cause me to have to restart the entire bot. An event would be very crucial for debugging major Rate Limit issues such as the one I had recently. This allows full customization of how you want to handle request limits. I could for example even set up an alert system to ping me and my dev team on discord when requests are being spammed or potentially hitting Global for example.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -881,6 +881,10 @@ declare namespace Eris {
       event: "relationshipUpdate",
       listener: (relationship: Relationship, oldRelationship: { type: number }) => void
     ): T;
+    (
+      event: "requestHandled",
+      listener: (method: string, url: string, auth: boolean, body: unknown, file: MessageFile, route: string, short: boolean) => void
+    ): T;
     (event: "typingStart", listener: (channel: TextableChannel, user: User) => void): T;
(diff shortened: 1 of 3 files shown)


LJNeon:
I would suggest adding this to the `debug` event instead of creating a new event personally, maybe make the log a tad more readable instead of just logging everything in a unmarked list as well. If a new event is needed, something like `rawREST` might fit better to match the existing `rawWS` event.

Skillz4Killz:
Although I agree debug is a potential route, a separate event can also be useful for creating analytics telling you which requests are going out. The issue with the `debug` imo is it is a massive amount of logs that go inside the debug event and it just adds unnecessary complexity to figure out what you want to debug.

`rawREST` is good name. Happy to change it to that as soon as i get back on VSC

Khaaz:
We're seeing changes from other PR (merged PR) in your diff. You might want to clean that to only keep the change that this PR introduce and remove the weird merge changes

PR #915, opened 2020-04-21
Github Action To Test Linter

Skillz4Killz:
Let's make use of that free GHActions CI!

This should begin testing all commits on all PRs going forward.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,6 +1,6 @@
 name: Lint
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


Khaaz:
no idea how abal feels about that, but if you want to make this work, you'll need to
1 - make it also trigger on `pull_request`
2 - make 2 separate jobs to lint js and ts

Skillz4Killz:
> no idea how abal feels about that, but if you want to make this work, you'll need to
> 1 - make it also trigger on `pull_request`
> 2 - make 2 separate jobs to lint js and ts

Done

Khaaz:
I'd still add 2 separate jobs instead of steps. That way we explicitely see which lint job is failing.
We would only see one failure with the current setup, indifferently if it's failing for js or ts

PR #922, opened 2020-04-29
Fix voice events intents

Skillz4Killz:
This PR should fix the issues related to members not being cached on VOICE_STATE_UPDATE and REACTION events which is a lot more evident and effects bots using intents. It also will help emit the voice events as well.

~~Note: There can be a slight confusion for when a user who is NOT cached switches channel it will be treated as joining the channel. Afterwards it will be fixed. I still think this is alot better then not emitting anything at all.~~ I may be wrong about this as I now think these members will be pre-cached on GUILD_CREATE through voice_states

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -526,6 +526,15 @@ class Shard extends EventEmitter {
                 let message;
                 if(channel) {
                     message = channel.messages.get(packet.d.message_id);
+                    if(channel.guild) {
+                        const member = channel.guild.members.get(packet.d.user_id);
+                        if(!member) {
+                            // Updates the member cache with this member for future events.
+                            channel.guild.memberCount++;
+                            packet.d.member.id = packet.d.user_id;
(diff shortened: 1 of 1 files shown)


lib/gateway/Shard.js
```diff
@@ -549,7 +552,7 @@ class Shard extends EventEmitter {
                 */
                 this.emit("messageReactionAdd", message || {
                     id: packet.d.message_id,
-                    channel: channel || { id: packet.d.channel_id }
+                    channel: channel || {id: packet.d.channel_id}
```
Khaaz, line 562: might not want to fix lint errors from changes that are not directly yours in this PR I guess 🤷‍♂️

Skillz4Killz: Sorry i just did fix all with eslint popup. I think once the other lint prs are merged this will automatically be removed

abalabahaha (approved):

PR #925, opened 2020-04-30
Adding tag value

userterminal:
The "tag" value makes it easier for people to write out username#discriminator without having to take all the time to type it all out.


abalabahaha:
Duplicate of #504 #325 #252

PR #926, opened 2020-05-01
Member: default data.id to data.user.id when nonexistent

Ratismal:
Today I came upon an extremely obscure edge case with getRESTGuildMember.

The error:
```
Error: Missing object id
    at Map.update (D:\JS\uno\node_modules\eris\lib\util\Collection.js:164:19)
    at Member.update (D:\JS\uno\node_modules\eris\lib\structures\Member.js:90:40)
    at new Member (D:\JS\uno\node_modules\eris\lib\structures\Member.js:60:14)
    at D:\JS\uno\node_modules\eris\lib\Client.js:1999:117
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

The requirements for reproducing this are as follows:
1. the bot doesn't have a member cached
2. the member in question is server muted
3. getRESTGuildMember is used to retrieve the member

The issue arises in the Member#update function. https://github.com/abalabahaha/eris/blob/master/lib/structures/Member.js#L82-L91

This function checks if the data packet contains a property called "mute". If so, and the following conditions are met:
1. The voice channel isn't null, or the user isn't muted, deafened, or suppressed
1. a voiceState doesn't already exist for the user

it tries to add the data packet to the guild.voiceStates collection. The problem is, the data packet when retrieved with getRESTGuildMember does not contain an id property. Here's a snapshot:
```js
{
  user: {
    id: '103347843934212096',
    username: 'stupid cat',
    avatar: 'a_13e35e43ce8a69d758300b681a2c348c',
    discriminator: '8160',
    public_flags: 131840
  },
  roles: [
    // ...
  ],
  nick: null,
  premium_since: null,
  mute: true,
  deaf: false,
  joined_at: '2016-06-19T23:30:30.499000+00:00'
}
```

I'm assuming that there's a reason a voiceState is maintained for a user who isn't in a channel but is muted, deafened, or suppressed, so I have created a default at the beginning of the Member constructor to add the user's ID to the data packet if it doesn't exist. An alternative solution would to change the line at https://github.com/abalabahaha/eris/blob/master/lib/structures/Member.js#L84
```js
if(data.channel_id === null) {
```

Thank you!

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -41,6 +41,9 @@ const VoiceState = require("./VoiceState");
 class Member extends Base {
     constructor(data, guild, client) {
         super(data.id || data.user.id);
+        if(!data.id) {
+            data.id = data.user.id;
+        }
         if((this.guild = guild)) {
             this.user = guild.shard.client.users.get(data.id);
             if(!this.user && data.user) {
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -41,7 +41,7 @@ const VoiceState = require("./VoiceState");
 class Member extends Base {
     constructor(data, guild, client) {
         super(data.id || data.user.id);
-        if(!data.id) {
+        if(!data.id && data.user) {
             data.id = data.user.id;
         }
         if((this.guild = guild)) {


abalabahaha (approved):

lib/structures/Member.js
```diff
@@ -41,6 +41,9 @@ const VoiceState = require("./VoiceState");
 class Member extends Base {
     constructor(data, guild, client) {
         super(data.id || data.user.id);
+        if(!data.id && data.user) {
+            data.id = data.user.id;
```
Khaazz, line 45: I'm sorry to comment only now a merged PR, but couldn't that have been solved by simply using `this.id` in the constructor instead?

PR #930, opened 2020-05-03
Refactor guild member requesting

TTtie:
This PR:  
- Adds `Guild#fetchMembers`, allowing to request specific members from Discord's gateway. 
- Refactors `GUILD_MEMBERS_CHUNK` processing to use the chunk count provided by Discord.

Todo:  
- [x] Check if it breaks non-intents support (should work with non-intents now, let me know if it doesn't)
- [x] Check if it still works when fetching > 1 chunk
- [x] Use a better method than `Math.random()` to generate the nonce?
- [ ] Check if it is still somewhat fast (should be)


lib/gateway/Shard.js
```diff
@@ -1260,20 +1260,34 @@ class Shard extends EventEmitter {
...
+                            member.update(presence);
                         }
+                    });
+                }
+
+                if(this.requestMembersPromise.hasOwnProperty(packet.d.nonce)) {
+                    this.requestMembersPromise[packet.d.nonce].members.push(...members);
+                }
+
+                if(packet.d.chunk_count <= (packet.d.chunk_index + 1)) {
```
abalabahaha:
```suggestion
                if(packet.d.chunk_index >= packet.d.chunk_count - 1) {
```

abalabahaha (approved):

PR #932, opened 2020-05-04
Update Discord URL

LJNeon:
Discord has changed their URL from `discordapp.com` to `discord.com`. However, this change does not apply to the CDN URL (`cdn.discordapp.com`).

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -1052,7 +1052,7 @@ class Client extends EventEmitter {
     * @arg {String} channelID The ID of the channel
     * @arg {String | Array | Object} content A string, array of strings, or object. If an object is passed:
     * @arg {String} content.content A content string
-    * @arg {Object} [content.embed] An embed object. See [the official Discord API documentation entry](https://discordapp.com/developers/docs/resources/channel#embed-object) for object structure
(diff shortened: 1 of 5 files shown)


userterminal:
Was just about to change this

alula:
What about ability to set a custom API base URL at this point for cases such as using it with Litecord/Mooncord/any similar thing?

LJNeon:
Seems like I forgot to reply, but that does seem like a good idea. There are some other use cases for it like people who want to use v6 of the API instead of v7.

exoRift:
How come they didn't change their CDN URL?

abalabahaha:
[They don't want to](https://cdn.discordapp.com/attachments/381886868708655104/706949338042990662/unknown.png)

tbnritzdoge (approved):

tbnritzdoge (approved):
is good

Vetlix (approved):

Brayzure:
>What about ability to set a custom API base URL at this point for cases such as using it with Litecord/Mooncord/any similar thing?

Interesting idea, but probably belongs in its own PR.

abalabahaha (approved):

PR #920, opened 2020-04-24
Support for withCount field for GUILD endpoint

Khaaz:
Ref discord/discord-api-docs#1528

- `withCount` for `getRESTGuild` method
- `approximateMemberCount` property on Guild
- `approximatePresenceCount` property on Guild

PR #933, opened 2020-05-04
Optional auth for execWebhook

Khaaz:
- Remove executing webhook with no webhook token but bot auth (which is no longer possible).
- Add an option to auth with your bot token when executing a webhook. It allows to use custom emojis that are not in the current server

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -642,6 +642,7 @@ class Client extends EventEmitter {
     * @arg {String} webhookID The ID of the webhook
     * @arg {String} token The token of the webhook
     * @arg {Object} options Webhook execution options
+    * @arg {Boolean} [options.auth=false] Whether to auth with the bot token (allow to use other guilds custom emotes)
     * @arg {String} [options.content=""] A content string
     * @arg {Object | Object[]} [options.file] A file object (or an Array of them)
     * @arg {Buffer} options.file.file A buffer containing file data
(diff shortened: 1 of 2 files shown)


lib/Client.js
```diff
@@ -1929,13 +1930,14 @@ class Client extends EventEmitter {
     /**
     * Get a guild's data via the REST API. REST mode is required to use this endpoint.
     * @arg {String} guildID The ID of the guild
+    * @arg {Boolean} [withCounts=false] Whether the guild object will have approximateMemberCount and approximatePresenceCount
     * @returns {Promise<Guild>}
     */
-    getRESTGuild(guildID) {
+    getRESTGuild(guildID, withCounts = false) {
```
TTtie: Uh..  this is added in from #920 by accident, isn't it?

Khaaz: oh, I messed up somewhere thanks

lib/Client.js
```diff
@@ -682,7 +683,7 @@ class Client extends EventEmitter {
     executeSlackWebhook(webhookID, token, options) {
         const wait = !!options.wait;
         options.wait = undefined;
-        return this.requestHandler.request("POST", (token ? Endpoints.WEBHOOK_TOKEN_SLACK(webhookID, token) : Endpoints.WEBHOOK_SLACK(webhookID)) + (wait ? "?wait=true" : ""), !token, options);
+        return this.requestHandler.request("POST", Endpoints.WEBHOOK_TOKEN_SLACK(webhookID, token) + (wait ? "?wait=true" : ""), false, options);
```
abalabahaha: Does this one not support auth?

Khaaz: To trigger this endpoint, you need to send slack-formatted data so I assumed auth was just not really necessary. We can probably add the option for auth anyway but I didn't think it was of any use

abalabahaha (approved):

PR #934, opened 2020-05-04
Add support for include_roles in guild prune

Khaaz:
Ref: discord/discord-api-docs#1563 

Add support for the includeRoles parameter in `getPruneCount` and `pruneMembers`.
I also fixed typings and jsdoc as all fields are optional.

⚠️  This is a breaking change as `includeRoles` is added before `reason` in pruneMembers.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1066,7 +1066,7 @@ declare namespace Eris {
     editRolePosition(guildID: string, roleID: string, position: number): Promise<void>;
     deleteRole(guildID: string, roleID: string, reason?: string): Promise<void>;
     getPruneCount(guildID: string, options?: GetPruneOptions): Promise<number>;
-    pruneMembers(guildID: string, options: PruneMemberOptions): Promise<number>;
+    pruneMembers(guildID: string, options?: PruneMemberOptions): Promise<number>;
     getVoiceRegions(guildID: string): Promise<VoiceRegion[]>;
     getInvite(inviteID: string, withCounts?: boolean): Promise<RESTInvite>;
     acceptInvite(inviteID: string): Promise<RESTInvite>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1709,7 +1709,7 @@ declare namespace Eris {
   }
 
   interface PruneMemberOptions extends GetPruneOptions {
-    reason?: string
+    reason?: string;
   }
 
   type Status = "online" | "idle" | "dnd" | "offline";
--- a/index.d.ts
+++ b/index.d.ts
@@ -1397,7 +1397,7 @@ declare namespace Eris {
     deleteEmoji(emojiID: string, reason?: string): Promise<void>;
     createRole(options: RoleOptions | Role, reason?: string): Promise<Role>;
     getPruneCount(options?: GetPruneOptions): Promise<number>;
-    pruneMembers(options: PruneMemberOptions): Promise<number>;
+    pruneMembers(options?: PruneMemberOptions): Promise<number>;
     getRESTChannels(): Promise<AnyGuildChannel[]>;
     getRESTEmojis(): Promise<Emoji[]>;
     getRESTEmoji(emojiID: string): Promise<Emoji>;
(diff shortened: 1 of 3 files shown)


lib/Client.js
```diff
@@ -869,26 +869,30 @@ class Client extends EventEmitter {
     /**
     * Get the prune count for a guild
     * @arg {String} guildID The ID of the guild
-    * @arg {Number} days The number of days of inactivity to prune for
+    * @arg {Number} [days] The number of days of inactivity to prune for. Default to 7
```
abalabahaha: `[days=7]` to show default values, not another sentence

abalabahaha (changes requested):
Would be a good time to merge `days`/`reason`/`includeRoles` into an `options` object

lib/Client.js
Khaaz: I thought we would only use those if it defaulted in the code. But I assume it's the same effect anyway.
```suggestion
    * @arg {Number} [days=7] The number of days of inactivity to prune for
```

Khaaz:
Yeah I hesitated doing that as well. Do we want to do it for both getPruneCount and pruneMembers?
I'd assume we want but that makes it a breaking change for both method instead of only one

abalabahaha:
Yes

index.d.ts
```diff
@@ -1065,8 +1065,8 @@ declare namespace Eris {
     editRole(guildID: string, roleID: string, options: RoleOptions, reason?: string): Promise<Role>; // TODO not all options are available?
     editRolePosition(guildID: string, roleID: string, position: number): Promise<void>;
     deleteRole(guildID: string, roleID: string, reason?: string): Promise<void>;
-    getPruneCount(guildID: string, days: number): Promise<number>;
-    pruneMembers(guildID: string, days: number, reason?: string): Promise<number>;
+    getPruneCount(guildID: string, options: GetPruneOptions): Promise<number>;
```
```suggestion
    getPruneCount(guildID: string, options?: GetPruneOptions): Promise<number>;
```

index.d.ts
```diff
@@ -1065,8 +1065,8 @@ declare namespace Eris {
...
+    pruneMembers(guildID: string, options: PruneMemberOptions): Promise<number>;
```
```suggestion
    pruneMembers(guildID: string, options?: PruneMemberOptions): Promise<number>;
```

index.d.ts
```diff
@@ -1396,8 +1396,8 @@ declare namespace Eris {
     editEmoji(emojiID: string, options: { name: string; roles?: string[] }, reason?: string): Promise<Emoji>;
     deleteEmoji(emojiID: string, reason?: string): Promise<void>;
     createRole(options: RoleOptions | Role, reason?: string): Promise<Role>;
-    getPruneCount(days: number): Promise<number>;
-    pruneMembers(days: number, reason?: string): Promise<number>;
+    getPruneCount(options: GetPruneOptions): Promise<number>;
```
```suggestion
    getPruneCount(options?: GetPruneOptions): Promise<number>;
```

index.d.ts
```diff
@@ -1396,8 +1396,8 @@ declare namespace Eris {
...
+    pruneMembers(options: PruneMemberOptions): Promise<number>;
```
```suggestion
    pruneMembers(options?: PruneMemberOptions): Promise<number>;
```

index.d.ts
```diff
@@ -1703,6 +1703,15 @@ declare namespace Eris {
     };
   }
 
+  interface GetPruneOptions {
+    days?: number,
```
```suggestion
    days?: number;
```

index.d.ts
```diff
@@ -1703,6 +1703,15 @@ declare namespace Eris {
...
+    includeRoles?: string[]
```
```suggestion
    includeRoles?: string[];
```

index.d.ts
```diff
@@ -1703,6 +1703,15 @@ declare namespace Eris {
...
+  }
+
+  interface PruneMemberOptions extends GetPruneOptions {
+    reason?: string
```
```suggestion
    reason?: string;
```

abalabahaha (approved):

PR #935, opened 2020-05-04
Fix MESSAGE_REACTION_REMOVE_ALL and MESSAGE_REACTION_REMOVE_EMOJI

Khaaz:
This PR is a follow up to #672 and allow the event to be fired even if the channel is not cached.

This PR also fix `MESSAGE_REACTION_REMOVE_ALL` and `MESSAGE_REACTION_REMOVE_EMOJI` behaviour. These 2 events were not updating reaction cache at all which could lead to invalid cache.

Additionally this PR makes `REACTION_REMOVE` remove the reaction object from `message.reaction` for this specific emoji if there are no emoji. (I explicitely assign undefined instead of deleting directly)

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -583,7 +583,7 @@ class Shard extends EventEmitter {
                     if(reactionObj) {
                         --reactionObj.count;
                         if(reactionObj.count === 0) {
-                            message.reactions[reaction] = undefined;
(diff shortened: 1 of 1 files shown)


Khaaz:
Switched to `delete` instead of assigning to undefined as requested.

abalabahaha (approved):

PR #936, opened 2020-05-04
Support for guild/members/search members endpoint

Khaaz:
Ref: discord/discord-api-docs#1577

This PR add support for the new guild/members/search endpoint: `GUILD_MEMBERS_SEARCH`.
It allows to get a list of guild members matching a string query (it matches against username and nickname)

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -2049,7 +2049,7 @@ class Client extends EventEmitter {
     }
 
     /**
-    * Search a member from a specific guild from the REST API via a query string to match against username and nickname.
+    * Search for guild members by partial nickname/username
     * @arg {String} guildID The ID of the guild
     * @arg {String} query The query string to match username(s) and nickname(s) against
     * @arg {Number} [limit=1] The maximum number of members you want returned, capped at 100
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -2051,8 +2051,8 @@ class Client extends EventEmitter {
     /**
     * Search a member from a specific guild from the REST API via a query string to match against username and nickname.
     * @arg {String} guildID The ID of the guild
-    * @arg {String} query The query string to match username(s) and nickname(s) against.
-    * @arg {Number} [limit] The maximum number of member you want in return. Default to 1. Maximum to 100.
+    * @arg {String} query The query string to match username(s) and nickname(s) against
+    * @arg {Number} [limit=1] The maximum number of members you want returned, capped at 100
     * @returns {Promise<Member[]>}
     */
     searchGuildMembers(guildID, query, limit) {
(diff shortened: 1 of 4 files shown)


herothedev:
This should be merged. 👍

Brayzure (commented):
Looks functional, only thing I want to bring up is due to how Member is initialized, Member#status will always be `offline`, regardless of actual status. Unsure if this is something that's acceptable.

Khaaz:
> Looks functional, only thing I want to bring up is due to how Member is initialized, Member#status will always be `offline`, regardless of actual status. Unsure if this is something that's acceptable.

Why would it be? .update() is called in Member constructor 🤔

Brayzure:
Because this endpoint doesn't appear to provide member status in the response, meaning it's stuck at the default of `offline`.

Khaaz:
Oh yeah, sorry I didn't understand it directly, haven't noticed that when testing yeah.
I'd say that's ok since it behave like that in the general case if the status is undefined?
Discord just don't want to provide presence info on this endpoint so we can just document that in the method?

Brayzure:
Perhaps, but I think it would be preferable to remove the property entirely after making the object. Shouldn't be a big deal since these aren't objects from the cache.

Khaaz:
that works for me, we can manually set it to undefined

abalabahaha:
Opposite would be better with intents - status undefined until a presence is found. That should be another PR though, I'd say leave status as-is in this one.

Khaaz:
Alright, will do 👍

Brayzure (dismissed):
Since member status changes will need to be in another PR, this looks good to me.

abalabahaha (approved):

lib/Client.js
```diff
@@ -2048,6 +2048,23 @@ class Client extends EventEmitter {
...
+    * @arg {Number} [limit=1] The maximum number of members you want returned, capped at 100
+    * @returns {Promise<Member[]>}
+    */
+    searchGuildMembers(guildID, query, limit) {
+        return this.requestHandler.request("GET", Endpoints.GUILD_MEMBERS_SEARCH(guildID), true, {
+            query,
+            limit
+        }).then((members) => {
+            const guild = this.guilds.get(guildID);
+            return members.map((member) => new Member(member, guild, this));
```
Skillz4Killz, line 2064: I would really like to see if we could add these members to cache. With the changes to intents, making sure our members cache is incredibly useful/important now. 

```ts
return members.map(member => guild.members.add(member, guild, this))
```

This way it will atleast add any members not currently cached

abalabahaha: No.

If the library doesn't modify the cache but you want it to, you can do this yourself with the members that are returned.

If the library does modify the cache but you don't want it to, that's much harder to reverse.

PR #942, opened 2020-05-08
Remove default status in Members

Khaaz:
As suggested in #936 , remove default `"offline"` status in Members.
Status won't be received at all if the presence intent is not enabled so we will now only assign `member.status` if we receive it.

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -57,7 +57,7 @@ class Member extends Base {
         } else {
             this.user = null;
         }
-        this.status = "offline";
+
         this.game = null;
         this.nick = null;
         this.update(data);
--- a/index.d.ts
+++ b/index.d.ts
@@ -107,7 +107,7 @@ declare namespace Eris {
   interface Presence {
     activities?: Activity[];
     clientStatus?: ClientStatus;
-    status: Status;
+    status?: Status;
     game?: Activity;
   }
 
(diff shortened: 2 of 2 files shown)


lib/structures/Member.js
```diff
@@ -57,7 +57,7 @@ class Member extends Base {
         } else {
             this.user = null;
         }
-        this.status = "offline";
+
```
Skillz4Killz, line 60: This might need a typings change 🤔 For status: to status?:

Khaaz: Addressed that, I had to make status optional in `Status` interface as well. Not sure if that's the correct way to do, I'll let you have a look

Skillz4Killz: I am busy this weekend so unable to give a thorough look through eris/api docs, take this with a grain of salt. 

I would have just removed the `implements Presence` on the Member tbh. I am not sure Discord would ever send a Presence which was undefined status so idk how accurate that is to mark it also as optional. I would wait for @alex-taxiera to chime in here as he has the best experience with Presence related stuff and it would affect his project most of all.

alex-taxiera: As the sole protector of presence data, I find no issue with this. All of the other presence related data is already marked as nullable anyhow, and really this interface was built to be implemented, not used as it's own data structure.

abalabahaha (approved):

PR #945, opened 2020-05-11
Expose hoistedRole property on Member

Khaaz:
Add support for the `hoistedRole` property on Member as specified in discord/discord-api-docs#1610
Only available on `message.member` through `MESSAGE_CREATE` / `MESSAGE_UPDATE`


abalabahaha:
This property should not be relied on as specified in the linked issue.

PR #946, opened 2020-05-11
Optional Timeout argument to the delete function

SonicMastr:
This just adds a timeout option where you specify the delay in ms for a message to be deleted. This makes it easier in general to set timeouts/delays without having to call setTimeout manually. This may not be the best way to implement it either, but it gets the job done

### DIFF
--- a/lib/structures/Message.js
+++ b/lib/structures/Message.js
@@ -353,9 +353,15 @@ class Message extends Base {
     /**
     * Delete the message
     * @arg {String} [reason] The reason to be displayed in audit logs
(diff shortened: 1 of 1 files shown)


knht:
This is definitely in the territory of not everyone needing it and those who do are best served implementing it themselves.

Besides that (and the PR not pointing to the dev branch of this repository), this PR lacks typings, is logically incomplete and can't be merged like that even if the change was a welcomed one.

LJNeon:
Yea gotta -1 this PR as well, even ignoring the many problems it has, this just isn't the kind of thing that fits Eris. If you want a library that babies you I'd recommend using discord.js instead.

PR #947, opened 2020-05-13
Fix broken ratelimiter

ak-s:
- timeOffset must be positive if local clock is behind (see [#L197](https://github.com/abalabahaha/eris/blob/dev/lib/rest/RequestHandler.js#L197)).
offset = serverTime - clientTime
reset = serverResetTime - offset
serverResetTime == serverTime + resetAfter
So for reactions: reset = clientTime + resetAfter (compatibility in case of disabled millisecond precision).
Reset calculations broken since 2017, how no one noticed?
- retryAfter >= 0 is true if retryAfter === null [#L257](https://github.com/abalabahaha/eris/blob/dev/lib/rest/RequestHandler.js#L257)
Just leave it as a possible NaN after parseInt.

### DIFF
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -196,7 +196,7 @@ class RequestHandler {
                         if(this.latencyRef.timeOffset - this.latencyRef.latency >= this._client.options.latencyThreshold && timeOffset - this.latencyRef.latency >= this._client.options.latencyThreshold) {
                             this._client.emit("warn", new Error(`Your clock is ${this.latencyRef.timeOffset}ms behind Discord's server clock. Please check your connection and system time.`));
                         }
-                        this.latencyRef.timeOffset = ~~(this.latencyRef.timeOffset - this.latencyRef.timeOffsets.shift() / 10 + timeOffset / 10);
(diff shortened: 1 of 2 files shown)


abalabahaha:
Why was the reactions override offset removed?

ak-s:
Because timeOffset is needed to adjust server time, not local.
clientTime + resetAfter ==> serverTime - offset + resetAfter ==> serverResetTime - offset

I think about adding half a second to headerNow in timeOffset calculation
`const timeOffset = headerNow + 500 - (this.latencyRef.lastTimeOffsetCheck = Date.now());`
to increase accuracy, what do you say?

abalabahaha:
TBH this was more of a failed experiment from a long time ago to see if we could cheat by sending requests early so that they arrive at Discord closer to the ratelimit reset time. But yeah I guess the resolution of the Date header is worse than the reaction ratelimit.

+500 does sound theoretically better, does it change things much in practice?

ak-s:
This may be significant for large bots. Just need some testing to make sure there is no 429 error.

ak-s:
At first everything worked fine, but after a day I got a few 429 errors caused by a rounding error.

abalabahaha (approved):

PR #948, opened 2020-05-13
Add support for following News Channels

bsian03:
Also part of the issue in [here](https://github.com/discord/discord-api-docs/issues/1510), see [this comment](https://github.com/discord/discord-api-docs/issues/1510#issuecomment-627911898)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -916,6 +916,11 @@ declare namespace Eris {
     (event: "resume", listener: () => void): T;
   }
 
+  interface ChannelFollow {
+    channel_id: string;
+    webhook_id: string;
+  }
+
   export class Client extends EventEmitter {
     token?: string;
     gatewayURL?: string;
@@ -1112,6 +1117,7 @@ declare namespace Eris {
       reason?: string
     ): Promise<number>;
     crosspostMessage(channelID: string, messageID: string): Promise<Message>;
+    followChannel(channelID: string, targetChannelID: string): Promise<ChannelFollow>;
     getGuildEmbed(guildID: string): Promise<GuildEmbed>;
     getGuildPreview(guildID: string): Promise<GuildPreview>;
     getGuildIntegrations(guildID: string): Promise<GuildIntegration[]>;
@@ -1594,6 +1600,7 @@ declare namespace Eris {
     rateLimitPerUser: 0;
     messages: Collection<Message<NewsChannel>>;
     crosspostMessage(messageID: string): Promise<Message<NewsChannel>>;
+    followChannel(targetChannelID: string): Promise<ChannelFollow>;
     getMessage(messageID: string): Promise<Message<NewsChannel>>;
     getMessages(limit?: number, before?: string, after?: string, around?: string): Promise<Message<NewsChannel>[]>;
     getPins(): Promise<Message<NewsChannel>[]>;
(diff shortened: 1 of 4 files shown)


bsian03:
Not able to test this yet btw, this is due to typescript build errors as a result of #905, I'm checking what the issue is with the author

lib/structures/NewsChannel.js
```diff
@@ -34,6 +34,15 @@ class NewsChannel extends TextChannel {
     crosspostMessage(messageID) {
         return this.client.crosspostMessage.call(this.client, this.id, messageID);
     }
+
+    /**
+     * Follow a NewsChannel
+     * @param {String} targetChannelID The ID of the target channel
```
Skillz4Killz: The permissions needed to run this should be listed here IMO.

lib/Client.js
```diff
@@ -1274,6 +1274,16 @@ class Client extends EventEmitter {
         return this.requestHandler.request("POST", Endpoints.CHANNEL_CROSSPOST(channelID, messageID), true).then((message) => new Message(message, this));
     }
 
+    /**
+     * Follow a NewsChannel
```
Skillz4Killz: Permissions needed to do this should be listed here as well

index.d.ts
```diff
@@ -916,6 +916,11 @@ declare namespace Eris {
     (event: "resume", listener: () => void): T;
   }
 
+  interface ChannelFollow {
+    channel_id: string;
```
Skillz4Killz, line 920: This might want to be looked at if we provide snake case anywhere else. I don't think Eris does and we should try to be consistant and modify this before getting it to the end user.

index.d.ts
```diff
@@ -916,6 +916,11 @@ declare namespace Eris {
...
+    webhook_id: string;
```
Skillz4Killz, line 921: Another snake case

lib/structures/NewsChannel.js
bsian03: Not consistent with other docs.

lib/Client.js
bsian03: See above.

index.d.ts
bsian03: Irrelevant, this is what Discord API returns and is consistent with other stuff since it isn't returning a class

index.d.ts
bsian03: See above.

index.d.ts
bsian03: Though.. thinking about it I'll ask abal later. I don't want to alter what the API returns too much

index.d.ts
ghost: aaaa

index.d.ts
ghost: bbb

lib/Client.js
ghost: ccc

lib/structures/NewsChannel.js
ghost: gggg

index.d.ts
bsian03: Actually, after having a look at it, snake_case is used, see FriendSuggestionReasons, Activity, OAuthApplicationInfo etc

bsian03:
Tested and working

abalabahaha (approved):

PR #950, opened 2020-05-14
Correct error event types

### LINKED ISSUE #944: TypeScript: "error" event specifies Error instead of ErrorEvent
TypeScript typing for Eris specifies that "error" event emits an `Error`, but actually it emits an `ErrorEvent` object.

```
    client.on("error", (err, id) => {
        logger.error("An error occurred on shard " + id);
        logger.error(err.error); // Property 'error' does not exist on type 'Error'.ts(2339)
    });
```

output of code as above but second call is `logger.error(err);`
```
  23:42:20.667  ErisClient       ERROR    An error occurred on shard 0
  23:42:20.667  ErisClient       ERROR    ErrorEvent {
  target: WebSocket {
    _events: [Object: null prototype] {
      open: [Function],
      message: [Function],
      error: [Function]
    },
    _eventsCount: 3,
    _maxListeners: undefined,
    readyState: 2,
    protocol: '',
```

bsian03:
Websocket errors do not use the error object itself, instead is used inside its own ErrorEvent object. Fired at `this.ws.onerror(event)` on L1798 in Shard.js
This fixes #944

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,6 +2,7 @@ import { EventEmitter } from "events";
 import { Readable as ReadableStream } from "stream";
 import { Agent as HTTPSAgent } from "https";
 import { IncomingMessage } from "http";
+import { ErrorEvent } from "ws";
 
 declare function Eris(token: string, options?: Eris.ClientOptions): Eris.Client;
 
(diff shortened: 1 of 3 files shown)


abalabahaha:
I'd rather have the WS emit an `Error`, which I'm drawing up

PR #956, opened 2020-05-21
Consistant Message.guildID When In A Guild

Skillz4Killz:
This PR helps add more support for message.guildID when it is missing from Discord's payloads. For example when you try and do `getMessage()` it will not provide guildID even if the message is inside a guild. This should help make it easier to use message.guildID.

### DIFF
--- a/lib/structures/Message.js
+++ b/lib/structures/Message.js
@@ -83,6 +83,10 @@ class Message extends Base {
             } else {
                 this.member = null;
             }
+
+            if(!this.guildID) {
+                this.guildID = this.channel.guild.id;
+            }
         } else {
             this.member = null;
         }


lib/structures/Message.js
```diff
@@ -83,6 +83,10 @@ class Message extends Base {
             } else {
                 this.member = null;
             }
+
+            if(!this.guildID) {
+                this.guildID = this.channel.guild.id;
```
TTtie, line 88: Please make this check whether the guild is actually present on a channel, as `getMessage` can be executed on a DM channel as well.

Something like
```suggestion
                this.guildID = this.channel.guild && this.channel.guild.id;
```
or
```suggestion
                if (this.channel.guild) {
                    this.guildID = this.channel.guild.id;
                }
```
would suffice, in my opinion.

TTtie (changes requested):

lib/structures/Message.js
bsian03: This has already been checked in L77, it is all in the same block

TTtie: Oh I'm blind, disregard...

abalabahaha (approved):

PR #957, opened 2020-05-23
Refurbish all documentation

bsian03:
So atm there are a few "issues" with the documentation of Eris, mostly with TypeScript documentation:
* TypeScript projects have no intellisense documentation whatsoever, as JSDoc in the core JS files do not get translated over. This PR adds that
* There are a couple (this may change to a load as I go on) of missing documentation for properties/methods for both TS and JSDoc. This PR adds those. Includes RequestHandler and SequentialBucket which are both added in this PR
* Some of the types in the TypeScript documentation are incorrect. I've painstakingly tracked down each and every line of code to correct this.
* The formatting of TypeScript documentation is readable, but doesn't make a lot of sense. E.g. there are a few duplicates, but why are they not referenced by a single source? I've made this a little bit cleaner.
* (something I haven't done yet, and would prefer some feedback/suggestions on this) The layout of TS Documentation is all jumbled up. i.e. interfaces are all the way at the top and at the bottom, with type declarations and classes (though some are grouped together) nested in between those interfaces. This hasn't been done yet and I'll edit it when I get to it.

I plan to fix all of these issues to the best of my ability. This will take a long time however.

**This is a WIP. If you have any feedback, please let me know and I will sort them out on the fly**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,7 +2,6 @@ import { EventEmitter } from "events";
 import { Readable as ReadableStream } from "stream";
 import { Agent as HTTPSAgent } from "https";
 import { IncomingMessage } from "http";
-import WebSocket = require("ws");
 
 declare function Eris(token: string, options?: Eris.ClientOptions): Eris.Client;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -749,13 +749,21 @@ declare namespace Eris {
     reconnectDelay?: ReconnectDelayFunction;
   }
   interface CommandClientOptions {
+    /** Whether to register the default help command or not */
     defaultHelpCommand?: boolean;
+    /** The description to show in the default help command */
     description?: string;
+    /** Whether to ignore bot accounts or not */
     ignoreBots?: boolean;
+    /** Whether to ignore the bot's own account or not */
     ignoreSelf?: boolean;
+    /** The bot name to show in the default help command */
     name?: string;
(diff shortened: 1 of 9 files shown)


index.d.ts
```diff
@@ -2,6 +2,7 @@ import { EventEmitter } from "events";
 import { Readable as ReadableStream } from "stream";
 import { Agent as HTTPSAgent } from "https";
 import { IncomingMessage } from "http";
+import WebSocket = require("ws");
```
```suggestion
```
Only doing this because i cba stashing my uncommited changes

index.d.ts
```diff
@@ -1946,87 +1986,304 @@ declare namespace Eris {
...
+    file: Buffer;
+    /** What to name the file */
+    name: string;
+  }
+
+  /**
+   * Handles API requests
+   */
+  class RequestHandler implements SimpleJSON {
+    agent: unknown; // HTTP Agent - Do we parse our preferred agent here?
```
bsian03, line 2167: cc @abalabahaha need clarification on this please
This is referenced from `Client.options.agent`

Skillz4Killz:
Small reminder as discussed on discord:

Change the @arg to @prop on the jsdocs
https://github.com/abalabahaha/eris/blob/dev/lib/rest/RequestHandler.js#L175-L185

bsian03:
I'll come back to this once I stop starting more projects

PR #958, opened 2020-05-29
Improve invite typings

eritbh:
Rather than letting invites be typed as a mess of interfaces, the `Invite` class is actually typed as a class now. This class stores metadata properties as `null` in situations where metadata is not returned by the API, which is now reflected accurately. Additional types were added for specifying private invites and guild channel invites, both with and without the additional metadata.

In situations where an invite is possibly a guild or private invite, its type can now be deduced via `invite.guild` being truthy, rather than having to use `'guild' in invite` to check. This is a side effect of the way TS handles the distinction between omitted properties and properties with `undefined` as a value.

I'm unsure of a couple relevant API details; will leave comments asking about them.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1074,8 +1074,8 @@ declare namespace Eris {
     getPruneCount(guildID: string, options?: GetPruneOptions): Promise<number>;
     pruneMembers(guildID: string, options?: PruneMemberOptions): Promise<number>;
     getVoiceRegions(guildID: string): Promise<VoiceRegion[]>;
-    getInvite(inviteID: string, withCounts?: boolean): Promise<PrivateInvite | GuildInviteWithoutMetadata>;
-    acceptInvite(inviteID: string): Promise<PrivateInvite | GuildInviteWithoutMetadata>;
+    getInvite(inviteID: string, withCounts?: boolean): Promise<AnyInviteWithoutMetadata>;
+    acceptInvite(inviteID: string): Promise<AnyInviteWithoutMetadata>;
     deleteInvite(inviteID: string, reason?: string): Promise<void>;
     getSelf(): Promise<ExtendedUser>;
     editSelf(options: { username?: string; avatar?: string }): Promise<ExtendedUser>;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1034,12 +1034,12 @@ declare namespace Eris {
...
     ): Promise<void>;
     deleteChannelPermission(channelID: string, overwriteID: string, reason?: string): Promise<void>;
-    getChannelInvites(channelID: string): Promise<ChannelInvite[]>;
+    getChannelInvites(channelID: string): Promise<GuildInviteWithMetadata[]>;
     createChannelInvite(
       channelID: string,
       options?: { maxAge?: number; maxUses?: number; temporary?: boolean; unique?: boolean },
       reason?: string
-    ): Promise<ChannelInvite>;
+    ): Promise<GuildInviteWithoutMetadata>;
```
eritbh, line 1042: Discord docs don't say that invite metadata is returned on the create invite endpoint, but the previous implementation did use an interface that included it. Is the documentation accurate, or is the metadata actually returned in this case and the Discord docs are wrong? Ref https://discord.com/developers/docs/resources/channel#create-channel-invite

(this applies to all other `createInvite` methods on other structures)

index.d.ts
```diff
@@ -1074,8 +1074,8 @@ declare namespace Eris {
     getPruneCount(guildID: string, options?: GetPruneOptions): Promise<number>;
     pruneMembers(guildID: string, options?: PruneMemberOptions): Promise<number>;
     getVoiceRegions(guildID: string): Promise<VoiceRegion[]>;
-    getInvite(inviteID: string, withCounts?: boolean): Promise<RESTInvite>;
-    acceptInvite(inviteID: string): Promise<RESTInvite>;
+    getInvite(inviteID: string, withCounts?: boolean): Promise<PrivateInvite | GuildInviteWithoutMetadata>;
+    acceptInvite(inviteID: string): Promise<PrivateInvite | GuildInviteWithoutMetadata>;
```
eritbh: as suggested by @alex-taxiera I'll make a new alias, `AnyInviteWithoutMetadata`, to use in these positions so we don't have unions in the method definitions

eritbh:
CC'ing @DonovanDMC since they were the last one to change these typings. LMK if anything looks wrong.

bsian03:
Implemented in #993 with proper class declaration after a discussion in the channel

eritbh:
#993

PR #959, opened 2020-06-02
Add a getter for message links to the Message class

exoRift:
This PR adds a getter to Messages to fetch the link to the message for users

- Added a root URL to the Endpoints file
- Added the appropriate TS changes

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1782,6 +1782,7 @@ declare namespace Eris {
     cleanContent?: string;
     roleMentions: string[];
     channelMentions: string[];
+    link: string;
     editedTimestamp?: number;
     tts: boolean;
     mentionEveryone: boolean;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1782,7 +1782,7 @@ declare namespace Eris {
     cleanContent?: string;
     roleMentions: string[];
     channelMentions: string[];
-    link: string;
+    jumpLink: string;
     editedTimestamp?: number;
     tts: boolean;
     mentionEveryone: boolean;
--- a/lib/rest/Endpoints.js
+++ b/lib/rest/Endpoints.js
@@ -3,6 +3,7 @@
 const {REST_VERSION} = require("../Constants");
 
 module.exports.BASE_URL = "/api/v" + REST_VERSION;
+module.exports.CLIENT_URL = "https://discordapp.com"
 module.exports.CDN_URL = "https://cdn.discordapp.com";
 
 module.exports.CHANNEL =                                                (chanID) => `/channels/${chanID}`;
@@ -55,6 +56,7 @@ module.exports.GUILD_WEBHOOKS =                                        (guildID)
 module.exports.GUILD_WIDGET =                                          (guildID) => `/guilds/${guildID}/widget`;
 module.exports.GUILDS =                                                             "/guilds";
 module.exports.INVITE =                                               (inviteID) => `/invite/${inviteID}`;
+module.exports.MESSAGE =                         (guildID, channelID, messageID) => `/channels/${guildID}/${channelID}/${messageID}`
 module.exports.OAUTH2_APPLICATION =                                      (appID) => `/oauth2/applications/${appID}`;
 module.exports.USER =                                                   (userID) => `/users/${userID}`;
 module.exports.USER_BILLING =                                           (userID) => `/users/${userID}/billing`;
--- a/lib/rest/Endpoints.js
+++ b/lib/rest/Endpoints.js
@@ -56,7 +56,6 @@ module.exports.GUILD_WEBHOOKS =                                        (guildID)
 module.exports.GUILD_WIDGET =                                          (guildID) => `/guilds/${guildID}/widget`;
 module.exports.GUILDS =                                                             "/guilds";
 module.exports.INVITE =                                               (inviteID) => `/invite/${inviteID}`;
-module.exports.MESSAGE =                         (guildID, channelID, messageID) => `/channels/${guildID}/${channelID}/${messageID}`;
 module.exports.OAUTH2_APPLICATION =                                      (appID) => `/oauth2/applications/${appID}`;
 module.exports.USER =                                                   (userID) => `/users/${userID}`;
 module.exports.USER_BILLING =                                           (userID) => `/users/${userID}/billing`;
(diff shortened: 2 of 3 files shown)


exoRift:
You should probably squash these

lib/rest/Endpoints.js
```diff
@@ -3,6 +3,7 @@
 const {REST_VERSION} = require("../Constants");
 
 module.exports.BASE_URL = "/api/v" + REST_VERSION;
+module.exports.CLIENT_URL = "https://discordapp.com";
```
Skillz4Killz:
```suggestion
module.exports.CLIENT_URL = "https://discord.com";
```

I would recommend just using the new format so this doesnt need to be changed later

exoRift: The domain is being changed in another PR. I like to stay consistent with the branch I'm currently in so I'd leave changing the domain up to someone else further down the line.

Skillz4Killz: The URL is going to change. Your PR is a new addition, your not modifying anything. It doesn't make sense to me to add something you know needs to be fixed in another PR.

lib/rest/Endpoints.js
```diff
@@ -55,6 +56,7 @@ module.exports.GUILD_WEBHOOKS =                                        (guildID)
 module.exports.GUILD_WIDGET =                                          (guildID) => `/guilds/${guildID}/widget`;
 module.exports.GUILDS =                                                             "/guilds";
 module.exports.INVITE =                                               (inviteID) => `/invite/${inviteID}`;
+module.exports.MESSAGE =                         (guildID, channelID, messageID) => `/channels/${guildID}/${channelID}/${messageID}`;
```
Brayzure: This isn't an endpoint we're sending a request to, so this should probably be defined elsewhere. Should either be at the bottom of the file, or otherwise specified directly in the Message class.

Brayzure (changes requested):

lib/rest/Endpoints.js
exoRift: You think it would be ok if I just hardcoded this URL into the getter?

EDIT: Ah, you mean the bottom of the Endpoints file

Brayzure (dismissed):

lib/rest/Endpoints.js
```diff
@@ -3,6 +3,7 @@
 const {REST_VERSION} = require("../Constants");
 
 module.exports.BASE_URL = "/api/v" + REST_VERSION;
+module.exports.CLIENT_URL = "https://discord.com";
```
abalabahaha: This belongs under the next line

lib/rest/Endpoints.js
```diff
@@ -87,3 +88,6 @@ module.exports.GUILD_BANNER =                             (guildID, guildBanner)
 module.exports.GUILD_ICON =                                 (guildID, guildIcon) => `/icons/${guildID}/${guildIcon}`;
 module.exports.GUILD_SPLASH =                             (guildID, guildSplash) => `/splashes/${guildID}/${guildSplash}`;
 module.exports.USER_AVATAR =                                (userID, userAvatar) => `/avatars/${userID}/${userAvatar}`;
+
+// Client Links
```
abalabahaha:
```suggestion
// Client Endpoints
```

lib/rest/Endpoints.js
```diff
@@ -87,3 +88,6 @@ module.exports.GUILD_BANNER =                             (guildID, guildBanner)
...
+module.exports.MESSAGE =                         (guildID, channelID, messageID) => `/channels/${guildID}/${channelID}/${messageID}`;
```
abalabahaha: Should be `MESSAGE_LINK`

lib/structures/Message.js
```diff
@@ -19,6 +20,7 @@ const User = require("./User");
 * @prop {String?} cleanContent Message content with mentions replaced by names, and @everyone/@here escaped
 * @prop {String[]} roleMentions Array of mentioned roles' ids
 * @prop {String[]} channelMentions Array of mentions channels' ids
+* @prop {String} link The link users can use to jump to the message
```
abalabahaha:
```suggestion
* @prop {String} link The url used by Discord clients to jump to this message
```

index.d.ts
```diff
@@ -1782,6 +1782,7 @@ declare namespace Eris {
     cleanContent?: string;
     roleMentions: string[];
     channelMentions: string[];
+    link: string;
```
abalabahaha: `jumpLink` would be a better name

abalabahaha (changes requested):

lib/rest/Endpoints.js
exoRift: Should we be calling it an endpoint if we're not making a request to it?

Brayzure: I spoke somewhat poorly. It is an endpoint, just one that Eris isn't calling itself (unlike all the others).

lib/rest/Endpoints.js
```diff
@@ -87,3 +88,6 @@ module.exports.GUILD_BANNER =                             (guildID, guildBanner)
 module.exports.GUILD_ICON =                                 (guildID, guildIcon) => `/icons/${guildID}/${guildIcon}`;
 module.exports.GUILD_SPLASH =                             (guildID, guildSplash) => `/splashes/${guildID}/${guildSplash}`;
 module.exports.USER_AVATAR =                                (userID, userAvatar) => `/avatars/${userID}/${userAvatar}`;
+
+// Client Endpoints
+module.exports.MESSAGE_LINK =                         (guildID, channelID, messageID) => `/channels/${guildID}/${channelID}/${messageID}`;
```
abalabahaha: Out of alignment :(

abalabahaha (changes requested):

lib/rest/Endpoints.js
exoRift: Aw man, I'm stupid. Sorry! 😬

lib/structures/Message.js
```diff
@@ -272,6 +274,10 @@ class Message extends Base {
         return (this._channelMentions = (this.content.match(/<#[0-9]+>/g) || []).map((mention) => mention.substring(2, mention.length - 1)));
     }
 
+    get jumpLink() {
+        return `${Endpoints.CLIENT_URL}${Endpoints.MESSAGE_LINK(this.channel.type === 1 ? "@me" : this.channel.guild.id, this.channel.id, this.id)}`;
```
bsian03, line 278:
```suggestion
        return `${Endpoints.CLIENT_URL}${Endpoints.MESSAGE_LINK(this.guildID || "@me", this.channel.id, this.id)}`;
```
Just making it a little more concise

exoRift: Ah, I like this

exoRift: Ok, so, stupid of me to test AFTER committing, but it seems this isn't a valid property. I might have to revert this commit

bsian03: It definitely is valid..

abalabahaha (approved):

PR #963, opened 2020-06-03
Member on messageReactionAdd

Skillz4Killz:
This PR adds the entire member structure when available on the messageReactionAdd event. Can help save a fetch request to the API, also very important post-intent changes when you are not relying on cached members


lib/gateway/Shard.js
```diff
@@ -578,12 +579,12 @@ class Shard extends EventEmitter {
...
                 * @prop {String?} emoji.id The emoji ID (null for non-custom emojis)
                 * @prop {String} emoji.name The emoji name
-                * @prop {String} userID The ID of the user that added the reaction
+                * @prop {Member | String} reactor The member if in a guild or the ID of the user that added the reaction
                 */
                 this.emit("messageReactionAdd", message || {
                     id: packet.d.message_id,
                     channel: channel || {id: packet.d.channel_id}
-                }, packet.d.emoji, packet.d.user_id);
+                }, packet.d.emoji, member || packet.d.user_id);
```
abalabahaha: Copy 586 instead of `| String`

abalabahaha (changes requested):
~~`messageReactionRemove` needs the same treatment~~

On second thought, that's probably not worth it

lib/gateway/Shard.js
Skillz4Killz: I'm not sure what you meant by this abal. Can you please clarify what you need to have this merged

abalabahaha (approved):

abalabahaha (approved):

Issue #662, opened 2020-03-17
Suggestion: CommandError exception with message content


TenkenNoSoujiro:
I suggest the addition of an `eris.CommandError` class that is an `Error` subclass that takes an `eris.MessageContent` as its constructor argument. `Command#process` method could catch this error and return the message content when encountered.

Currently I am using something like this in my own Discord bot, but have to jump through several hoops to get it to work:

```js
export class CommandClient extends eris.CommandClient {
    registerCommand(label, generator, options) {
        const command = super.registerCommand(label, generator, options);
        if (Object.getPrototypeOf(command) === eris.Command.prototype) {
            Object.setPrototypeOf(command, Command.prototype);
        }
        return command;
    }
}

export class Command extends eris.Command {
    registerSubcommand(label, generator, options) {
        const command = super.registerSubcommand(label, generator, options);
        if (Object.getPrototypeOf(command) === eris.Command.prototype) {
            Object.setPrototypeOf(command, Command.prototype);
        }
        return command;
    }

    async process(msg, args) {
        try {
            return await super.process(msg, args);
        }
        catch (e) {
            if (e instanceof CommandError) return e.content;
            throw e;
        }
    }
}

const isFormattedRegExp = /^#|^(?<mark>[*_`~]).*\k<mark>$/s;

export class CommandError extends Error {
    constructor(content) {
        super("An error occurred while processing a command.");
        // for trivial content, add emphasis to the message...
        if (typeof content === "string" && !isFormattedRegExp.test(content)) {
            content = `*${content}*`;
        }
        this.content = content;
    }
}
```

I use this capability in the following way:

```js
bot.registerCommand("command", async (msg, args) => {
   …
   const result = await doSomeWork();
   …
});
async function doSomeWork() {
   …
   // The following message is sent by the bot
   throw new CommandError("Could not perform the requested action");
}
```

eritbh:
The capability you're looking for is already achievable with a minimal addition to your command definition:

```ts
bot.registerCommand("command", async (msg, args) => {
   …
   let result;
   try {
      result = await doSomeWork();
   } catch (error) {
      return error.message;
   }
   …
});
async function doSomeWork() {
   …
   // The following message is sent by the bot
   throw new Error("Could not perform the requested action");
}
```

Eris's command framework is not heavily used, and it's been debated in the past that it would be better to just remove it in favor of other third-party command frameworks that exist. I doubt the additions you're asking for will be accepted.

[...]

PR #964, opened 2020-06-04
Command Client: Transport message and error to errorMessage

### LINKED ISSUE #980: errorMessage not actually passed the message like docs suggest.
https://github.com/abalabahaha/eris/blob/6d4668cd134d0af6b28a63f9ba3fa2790e28954d/lib/command/CommandClient.js#L178

just needs a small update to ```const reply = msg.command.errorMessage(msg);``` ;)

andromika:
closes #980 and #662

### DIFF
--- a/lib/command/CommandClient.js
+++ b/lib/command/CommandClient.js
@@ -175,7 +175,7 @@ class CommandClient extends Client {
                     if(msg.command.errorMessage) {
                         try {
                             if(typeof msg.command.errorMessage === "function") {
-                                const reply = msg.command.errorMessage();
+                                const reply = msg.command.errorMessage(msg, err);
                                 if(reply !== undefined) {
                                     newMsg = await this.createMessage(msg.channel.id, reply);
                                 }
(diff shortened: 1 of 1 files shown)


lib/command/CommandClient.js
```diff
@@ -175,7 +175,7 @@ class CommandClient extends Client {
                     if(msg.command.errorMessage) {
                         try {
                             if(typeof msg.command.errorMessage === "function") {
-                                const reply = msg.command.errorMessage();
+                                const reply = msg.command.errorMessage(msg,err);
```
alex-taxiera: this should probably have a space between the params... not sure how the linter doesn't pick that up

Brayzure: It should, yes. Additionally, you'll need to update the docs to show that an error is being passed as well.

Brayzure (dismissed):

HPaulson (approved):
Space & Doc both updated now, LGTM! However, I’d recommend adding “closes 980” just to clear out the issues related to this PR.

abalabahaha (approved):

PR #965, opened 2020-06-06
Fix Bucket

motoenduroboy:
Current bucket logic leads to rate limiting issues


Brayzure (approved):
Ah yeah, my mistake. I moved the logic for calculating remaining tokens outside of the while loop, and forgot to update it.

PR #967, opened 2020-06-17
Fix Not Authenticated Errors

motoenduroboy:
After receiving a few "Not Authenticated" errors and resetting my token a few times. I checked discord docs to find out it had nothing to do with tokens and was due to packets being sent before hello was received. Added some logic to stop packets from being sent before hello is received. 

Been running on prod with no errors so far but others should check it on larger bots to see if it affects/breaks anything.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -64,6 +64,8 @@ class Shard extends EventEmitter {
         this._onWSError = this._onWSError.bind(this);
         this._onWSClose = this._onWSClose.bind(this);
 
+        this.identified = false;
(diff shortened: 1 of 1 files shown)


bsian03:
image
Just for context as to why the PR was closed (ignore my nickname pls kthx)

PassTheMayo:
I'm currently having this issue, but it only occurs when I use `getAllUsers` in the client settings. Turning it off will fix the issue, but doing so will break other features within my bot.

abalabahaha:
See the linked issue

PR #970, opened 2020-06-21
Encapsulate Textable channels in relevant interfaces

bsian03:
This is very use case specific to what I'm doing but it makes more sense to do this anyways, I don't know why I didn't in my original PR to fix union types (#949) - Shouldn't affect anything functionalitywise

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -24,7 +24,7 @@ declare namespace Eris {
   // TODO there's also toJSON(): JSONCache, though, SimpleJSON should suffice
 
   type GuildTextableChannel = TextChannel | NewsChannel
-  type TextableChannel = Textable & (GuildTextableChannel | PrivateChannel);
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -24,7 +24,7 @@ declare namespace Eris {
   // TODO there's also toJSON(): JSONCache, though, SimpleJSON should suffice
 
   type GuildTextableChannel = TextChannel | NewsChannel
-  type TextableChannel = Textable & GuildTextableChannel | PrivateChannel;
+  type TextableChannel = Textable & (GuildTextableChannel | PrivateChannel);
```
```suggestion
  type TextableChannel = (GuildTextable & GuildTextableChannel) | (Textable & PrivateChannel)
```
not on my laptop rn so this is the easiest way to change it

abalabahaha (approved):

PR #971, opened 2020-06-23
use proper function signature

alex-taxiera:
Updated catch all listener signature on event listeners. Signature was taken straight from EventEmitter

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -899,7 +899,7 @@ declare namespace Eris {
     ): T;
     (event: "voiceStateUpdate", listener: (member: Member, oldState: OldVoiceState) => void): T;
     (event: "warn" | "debug", listener: (message: string, id: number) => void): T;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -899,7 +899,7 @@ declare namespace Eris {
     ): T;
     (event: "voiceStateUpdate", listener: (member: Member, oldState: OldVoiceState) => void): T;
     (event: "warn" | "debug", listener: (message: string, id: number) => void): T;
-    (event: string, listener: Function): T;
+    (event: string, listener: (...args: any[]) => void): T;
```
Skillz4Killz, line 902:
```suggestion
    (event: string, listener: (...args: unknown[]) => void): T;
```

bsian03: `any` is actually consistent with the events typings

abalabahaha (approved):

PR #973, opened 2020-06-29
Update the user object when using cached user on message create

CheweyZ:


lib/structures/Message.js
```diff
@@ -68,6 +68,9 @@ class Message extends Base {
         if(data.author) {
             if(data.author.discriminator !== "0000") {
                 this.author = client.users.add(data.author, client);
+                if(this.author!==data.author){
```
```suggestion
                if(this.author !== data.author) {
```

bsian03 (changes requested):
consistency

lib/structures/Message.js
```diff
@@ -68,6 +68,9 @@ class Message extends Base {
...
+                    this.author.update(data.author);
+                }
```
TTtie: This as a whole can be merged together into `this.author = client.users.update(...)` on line 70

Khaaz: no we don't want to perform an update in the collection item. It would lead to invalid reference in other places.
We just want to update the Object itself. We do so by calling `<user>.update`

Khaaz: edit: I'm just dumb. the whole thing (even the first .add) can indeed be replaced by `client.users.update(data.author, client)`

abalabahaha (approved):

PR #977, opened 2020-07-02
Fix toJSON for ShardManager

apacheli:
Collection#toJSON() was different

### DIFF
--- a/lib/gateway/ShardManager.js
+++ b/lib/gateway/ShardManager.js
@@ -1,5 +1,6 @@
 "use strict";
 
+const Base = require("../structures/Base");
 const Collection = require("../util/Collection");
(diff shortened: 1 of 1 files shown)


lib/gateway/ShardManager.js
```diff
@@ -119,7 +120,7 @@ class ShardManager extends Collection {
     }
 
     toJSON(props = []) {
-        return super.toJSON([
+        return Base.prototype.toJSON.call([
```
abalabahaha:
```suggestion
        return Base.prototype.toJSON.call(this, [
```

abalabahaha (changes requested):

abalabahaha (approved):

PR #978, opened 2020-07-02
Fix "removeMessageReactionEmoji" on Message

apacheli:
Fix name for consistency purposes


exoRift:
You would also need to change the name of the client method

justsharan:
Yes and the typings too

apacheli:
> You would also need to change the name of the client method

Technically no. This was fix consistency with Client to structure methods such as `addMessageReaction` -> `addReaction` and so on (but for some reason this isn't the case with `createMessage`)

abalabahaha (approved):

PR #983, opened 2020-07-17
Minor typing fixes

funlennysub:
#657


bsian03:
#981 is already open, ~~you might want to comment the commandclient stuff there instead~~ also fixed now in #993

funlennysub:
[I checked and this is already fixed](https://github.com/abalabahaha/eris/pull/981/files#diff-b52768974e6bc0faccb7d4b75b162c99R1788), so I think we can close this pr

PR #985, opened 2020-07-21
Add GuildID property to message at reaction add/remove

Exanlv:
If the channel is not cached for messageReactionAdd and messageReactionRemove the guild id isnt in the event at all.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -622,7 +622,7 @@ class Shard extends EventEmitter {
                 /**
                 * Fired when someone removes a reaction from a message
                 * @event Client#messageReactionRemove
-                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel`. If the channel is not cached, channel key will be an object with only an id. `guildID` will be present if the message was sent in a guild channel. No other property is guaranteed
+                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. If the channel is not cached, channel key will be an object with only an id. `guildID` will be present if the message was sent in a guild channel. No other property is guaranteed
                 * @prop {Object} emoji The reaction emoji object
                 * @prop {String?} emoji.id The ID of the emoji (null for non-custom emojis)
                 * @prop {String} emoji.name The emoji name
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -640,15 +640,21 @@ class Shard extends EventEmitter {
                         message.reactions = {};
                     }
                 }
+                if(!message) {
+                    message = {
+                        id: packet.d.message_id,
+                        channel: channel || {id: packet.d.channel_id}
+                    };
+                    if(packet.d.guild_id) {
+                        message.guildID = packet.d.guild_id;
+                    }
+                }
                 /**
                 * Fired when all reactions are removed from a message
                 * @event Client#messageReactionRemoveAll
-                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. If the channel is not cached, channel key will be an object with only an id. No other property is guaranteed
+                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id`, `channel`, and if inside a guild, `guildID` keys. If the channel is not cached, channel key will be an object with only an id. No other property is guaranteed
                 */
-                this.emit("messageReactionRemoveAll", message || {
-                    id: packet.d.message_id,
-                    channel: channel || {id: packet.d.channel_id}
-                });
+                this.emit("messageReactionRemoveAll", message);
                 break;
             }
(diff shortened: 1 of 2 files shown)


Exanlv:
Just realized this wont work for DMs, will fix in a bit

lib/gateway/Shard.js
```diff
@@ -570,20 +570,26 @@ class Shard extends EventEmitter {
...
+
+                    if(packet.d.guild_id) {
+                        message.guildID = packet.d.guild_id;
+                    }
                 }
                 /**
                 * Fired when someone adds a reaction to a message
                 * @event Client#messageReactionAdd
-                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. If the channel is not cached, channel key will be an object with only an id. No other property is guaranteed
+                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. If the channel is not cached, channel key will be an object with only an id. `guildID` will be present if the message was sent in a guild channel. No other property is guaranteed
```
bsian03: Probably better if the `guildID` info is in the same sentence as the message not cached instead of after

lib/gateway/Shard.js
```diff
@@ -603,20 +609,26 @@ class Shard extends EventEmitter {
...
+
+                    if(packet.d.guild_id) {
+                        message.guildID = packet.d.guild_id;
+                    }
                 }
                 /**
                 * Fired when someone removes a reaction from a message
                 * @event Client#messageReactionRemove
-                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. If the channel is not cached, channel key will be an object with only an id. No other property is guaranteed
+                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel`. If the channel is not cached, channel key will be an object with only an id. `guildID` will be present if the message was sent in a guild channel. No other property is guaranteed
```
bsian03:
```suggestion
                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. If the channel is not cached, channel key will be an object with only an id. `guildID` will be present if the message was sent in a guild channel. No other property is guaranteed
```
I'm guessing you accidentally removed this

lib/gateway/Shard.js
Exanlv: `guildID` would be optional which is why I didnt put it in the same sentence.
How about "with `id`, `channel` and optionally `guildID` keys"?

bsian03: "with `id`, `channel`, and `guildID`, if in guild, keys" probably but I don't think that's much better haha
I just think "optionally" doesn't really fit the description and what you have right now feels out of place. My opinions though

bsian03: "`id`, `channel`, and if inside a guild, `guildID` keys"
That's probably better

Exanlv: Yeah I couldnt really think of a good way to phrase it haha. Ill change it to what you said.

Exanlv:
Added guild ID  to messageReactionRemoveAll as well

abalabahaha:
Shouldn't `messageReactionRemoveEmoji` receive the same treatment?

lib/gateway/Shard.js
```diff
@@ -628,15 +640,21 @@ class Shard extends EventEmitter {
                         message.reactions = {};
                     }
                 }
+                if(!message) {
+                    message = {
+                        id: packet.d.message_id,
+                        channel: channel || {id: packet.d.channel_id}
+                    };
+                    if(packet.d.guild_id) {
+                        message.guildID = packet.d.guild_id;
```
bsian03, line 655:
```suggestion
                        message.guildID = packet.d.guild_id;
                        if(!message.channel.guild) {
                            message.channel.guild = {id: packet.d.guild_id};
                        }
```

bsian03: Not sure if this is necessary but meh (same for the other ones)

Exanlv: If the channel is cached `message.channel.guild` should still be present. This would overwrite it?

Exanlv:
> Shouldn't `messageReactionRemoveEmoji` receive the same treatment?

Wasnt aware this existed. I'll add it in a bit

lib/gateway/Shard.js
bsian03: Ah yeah it would.. mb
Probably add something to check if guild doesn't exist before doing that then (updated the suggestion)

bsian03:
Typings will need to be updated too, just search for `PossiblyUncachedMessage` in the typings file and add the props there

lib/gateway/Shard.js
Exanlv: Added for all 4 events

abalabahaha (approved):

PR #986, opened 2020-07-22
Permission property name consistency

bsian03:
Renames Member#permission to Member#permissions to match with Role#permissions and deprecated old getter

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -26,7 +26,7 @@ const VoiceState = require("./VoiceState");
 * @prop {String?} nick The server nickname of the member
 * @prop {String[]} roles An array of role IDs this member is a part of
 * @prop {User} user The user object of the member
-* @prop {Permission} permission The guild-wide permissions of the member. Using this property is deprecated and will not be supported in future versions.
+* @prop {Permission} permission [DEPRECATED] The guild-wide permissions of the member. Use Member#permissions instead
 * @prop {Permission} permissions The guild-wide permissions of the member
 * @prop {String} defaultAvatar The hash for the default avatar of a user if there is no avatar set
 * @prop {Number} createdAt Timestamp of user creation
@@ -112,7 +112,7 @@ class Member extends Base {
     }
 
     get permission() {
-        this.guild.shard.client.emit("warn", "[DEPRECATED] member.permission was used. Please use the member.permissions getter instead.");
+        this.guild.shard.client.emit("warn", "[DEPRECATED] Member#permission is deprecated. Use Member#permissions instead");
         return this.permissions;
     }
 
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -26,7 +26,8 @@ const VoiceState = require("./VoiceState");
 * @prop {String?} nick The server nickname of the member
 * @prop {String[]} roles An array of role IDs this member is a part of
 * @prop {User} user The user object of the member
-* @prop {Permission} permission The guild-wide permissions of the member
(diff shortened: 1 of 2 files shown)


lib/structures/Member.js
```diff
@@ -26,7 +26,8 @@ const VoiceState = require("./VoiceState");
 * @prop {String?} nick The server nickname of the member
 * @prop {String[]} roles An array of role IDs this member is a part of
 * @prop {User} user The user object of the member
-* @prop {Permission} permission The guild-wide permissions of the member
+* @prop {Permission} permission The guild-wide permissions of the member. Using this property is deprecated and will not be supported in future versions.
+* @prop {Permission} permissions The guild-wide permissions of the member
```
HPaulson, line 28: Both props listed here are `{Permission}`, where as the second prop (the new one) should include the plural `s`

```diff
- * @prop {Permission} permissions The guild-wide permissions of the member
+ * @prop {Permissions} permissions The guild-wide permissions of the member
```

HPaulson (changes requested):

lib/structures/Member.js
bsian03: ????????????????????????????????????????

bsian03: Instead of making incorrect corrections to something that's already correct, please look at the documentation. Also I doubt I'm the only person who thinks this but please don't leave useless comments on PRs, thanks :)

HPaulson: @bsian03 Explain how this is incorrect? You have two props that are the exact same- `{Permission}`. Isn't the **entire** purpose of this PR to change `Member#Permission` to `Member#Permissions`, therefore the added deprecated text should be on the `Member#Permission` prop for the docs, and the new one should be `Member#Permissions`, as plural.

HPaulson: As it is currently, If I'm not mistaken, you'd have `Member#Permission` on the docs twice. One with the added deprecated text, and one without. The change I'm suggesting is that you change the new, non-deprecated point to `Permissions`, making the non-deprecated point plural for the docs. Pretty sure that's the entire point of this PR.

HPaulson: Lmfao actually, ignore those comments just made- Was on mobile earlier, and after looking at the file properly I see that's just the type, not the name. Not even sure how I did this honestly, Totally sorry for the mis-read haha

HPaulson (dismissed):
Completely misread file on mobile in my previous review- all this code LGTM!

lib/structures/Member.js
```diff
@@ -111,6 +112,11 @@ class Member extends Base {
     }
 
     get permission() {
+        this.guild.shard.client.emit("warn", "[DEPRECATED] member.permission was used. Please use the member.permissions getter instead.");
```
abalabahaha:
```suggestion
        this.guild.shard.client.emit("warn", "[DEPRECATED] Member#permission is deprecated. Use Member#permissions instead");
```

lib/structures/Member.js
```diff
@@ -26,7 +26,8 @@ const VoiceState = require("./VoiceState");
 * @prop {String?} nick The server nickname of the member
 * @prop {String[]} roles An array of role IDs this member is a part of
 * @prop {User} user The user object of the member
-* @prop {Permission} permission The guild-wide permissions of the member
+* @prop {Permission} permission The guild-wide permissions of the member. Using this property is deprecated and will not be supported in future versions.
```
abalabahaha:
```suggestion
* @prop {Permission} permission [DEPRECATED] The guild-wide permissions of the member. Use Member#permissions instead
```

abalabahaha (approved):

PR #988, opened 2020-07-22
Document Role.json in the correct place

bsian03:


bsian03:
On another note `Role.json` doesn't seem to make sense without the documentation since it looks like it gets the JSON of the role object itself and not the permissions and it basically serves as an alias to `Role.permissions.json`?

HPaulson (commented):
Not too sure why this PR needs to exist... As the previous comment mentioned, `Role.json` seems to be the json structure of the permissions, same as `Role.permissions.json`, and not the Json object of the role itself. This could be super confusing, and really isn't necessary to exist since it's just an alias of sorts. Although this PR is just to update the documentation, I'd argue that this shouldn't even exist in the structure. Otherwise, this PR is fine if it's being kept in the structure.

abalabahaha (approved):

PR #989, opened 2020-07-23
Add documentation for more props in oldChannel and oldGuild

jtsshieh:
In the channelUpdate event, userLimit was added and userLimit + rateLimitPerUser were documented. Did not test.


bsian03:
Should we add GroupChannel docs as well or is there no point since bots can't join Group DMs anyways?

abalabahaha (approved):

PR #990, opened 2020-07-26
Fixed cleanContent mention exploit

TheAkio:
People figured out a new exploit to mention roles using bots, although users can be mentioned using that too.

The scenario works as follows:

1. User sets it's nickname (or username) to `&<Role-ID>` i.e. `&371335384690327564`
2. They invoke a command in a bot that users `cleanContent` to display the user input back to the user with an input mentioning their own user surrounded by a pair of `<>` i.e. `<@&371335384690327564>` (As Displayed in Discord)
3. The role gets pinged

How did this happen?

The original message received from Discord will look like this: `<<@!108638204629925888>>`

The sanitization sees the inner mention and replaces it with the users username or nickname. Which in this case is `&371335384690327564` resulting in a `cleanContent` containing `<@&371335384690327564>` and therefore pinging a role (if the bot has permission to do so).

The fix I made basically just puts a ZWS between the @ and the Username/Nickname/Rolename. This was the best thing I could come up with right now. If there's any better solution let me know.

### DIFF
--- a/lib/structures/Message.js
+++ b/lib/structures/Message.js
@@ -236,17 +236,17 @@ class Message extends Base {
                 authorName = member.nick;
             }
         }
-        cleanContent = cleanContent.replace(new RegExp(`<@!?${this.author.id}>`, "g"), "@" + authorName);
+        cleanContent = cleanContent.replace(new RegExp(`<@!?${this.author.id}>`, "g"), "@\u200b" + authorName);
 
         if(this.mentions) {
             this.mentions.forEach((mention) => {
                 if(this.channel.guild) {
                     const member = this.channel.guild.members.get(mention.id);
                     if(member && member.nick) {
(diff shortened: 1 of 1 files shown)


Skillz4Killz:
Should cleanContent not just be removed now that there is allowedMentions?

TheAkio:
In the future, yes. First time I'm hearing about that tbh (Haven't been keeping up with Discords API for a while now). There's still a bunch of legacy codebases (including ours) that sadly rely on `cleanContent` working. It would still make sense fixing it, but also marking it as deprecated so it isn't used anymore

ghost:
this behavior is intended

cleanContent purpose is to prettify discord mentions and not to sanitize them. Your problem can quite easily get fixed with the allowedMentions feature.

abalabahaha (approved):

PR #991, opened 2020-07-29
Add fallbacks for typingStart event arguments

Dragory:
This PR adds fallbacks for typingStart event arguments (channel, user) in case the channel or user is not found in the cache.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -442,8 +442,8 @@ class Shard extends EventEmitter {
                     /**
                     * Fired when a user begins typing
                     * @event Client#typingStart
-                    * @prop {PrivateChannel | TextChannel} channel The text channel the user is typing in
-                    * @prop {User} user The user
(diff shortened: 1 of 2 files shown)


lib/gateway/Shard.js
```diff
@@ -445,7 +445,7 @@ class Shard extends EventEmitter {
                     * @prop {PrivateChannel | TextChannel} channel The text channel the user is typing in
                     * @prop {User} user The user
```
bsian03:
```suggestion
                    * @prop {PrivateChannel | TextChannel} channel The text channel the user is typing in. If the channel is not cached, this will be an object with an `id` key. No other property is guaranteed
                    * @prop {User} user The user. If the user is not cached, this will be an object with an `id` key. No other property is guaranteed
```

HPaulson (approved):
LGTM!

abalabahaha (approved):

PR #993, opened 2020-08-02
Documentation tidyup/updates

### LINKED ISSUE #1003: Typings for DiscordHTTPError (and DiscordRESTError)
These are useful things to have typings for so people can properly handle errors within the library if they so choose to.

### LINKED ISSUE #1004: createMessage: docs incorrectly list "Array of strings" as parameter
_(This is for v0.13.3, the latest release on npm as of this post; btw the docs on the website say they're for 0.13.2 in the title tag)_

The `Client.createMessage` function [in the docs](https://abal.moe/Eris/docs/Client#function-createMessage) says the `content` parameter can be:

> String | Array | Object
> A string, array of strings, or object. [...]

I wondered if an array of strings might send a random one, but it actually just throws this error:

> No content, file, or embed

It doesn't look like there's anything in the code to check or handle arrays of strings here, but it's mentioned in the comments:
https://github.com/abalabahaha/eris/blob/6d4668cd134d0af6b28a63f9ba3fa2790e28954d/lib/Client.js#L1067

bsian03:
Continuation of #957 (or rather another attempt)
<details>
<summary>See here for old stuff</summary>
I haven't done JSDoc this time round, however have done the structure changes:

From inside the namespace, the order is as follows)
* Constants - Constants and VERSION
* Types - grouped together, alphabetical order of group name and in each group (i haven't included group name)
* Interfaces, also as above (Exception being Events and Constants, of which are at the bottom of their respective group)
* Classes, in alphabetical order

Each category has 2 newline separations, with each group, or each class with a single newline separation

I'll also fill in missing docs for JSDoc

Invite.guild also now gets the relevant guild
Invite.channel now gets the relevant guild channel if the guild is cached. if group channel or not cached this will still be an object

(note for abal, please don't merge this with the other PRs since there's bound to be conflicts once the other PRs get merged)

</details>

## Documentation/codebase layout changes
- Everything is now in alphabetical order, this includes all docs (including TypeScript) and methods/props.
- Added a linter to enforce the above (classes only)
- Added missing and corrected incorrect docs
- Consistency in JSDoc types
- Removed superclass props from JSDoc
- Invite#guild and Invite#channel will now fetch their respective instances. If Invite#guild is not cached, this will still be an object. If Invite#guild is not cached or Invite#channel is a GroupChannel, this will also still be an object. (this should really be in its own PR but too late)
- Deprecate supplying `userID` parameter for messageReactionAdd and messageReactionRemove for DMs (Anything other than "@me" will error)
- Update to TypeScript v4!

### [TypeScript v4](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/)
- Updated typescript eslint parser and plugin
- Slight change to types to respect TypeScript v4 rules
- Add deprecation flag to deprecated properties/getters/methods! (This also works for overloads)

Warning may be shown in the console of the CI, this can be disregarded, [see here](https://github.com/typescript-eslint/typescript-eslint/issues/2260#issuecomment-677853883)

Resolves #1003
Resolves #1004


index.d.ts
```diff
@@ -8,834 +8,369 @@ declare function Eris(token: string, options?: Eris.ClientOptions): Eris.Client;
...
-    channelID?: string | null;
+  interface OldGuildChannel {
+    bitrate?: number;
+    name: string;
+    nsfw: boolean;
+    parentID?: string;
+    permissionOverwrites: Collection<PermissionOverwrite>;
+    rateLimitPerUser?: number;
+    position: string;
+    topic?: string;
```
jtsshieh, line 377: Discussed in #js_eris, topic can be null as well

bsian03: #981

jtsshieh: okay

bsian03:
Should I reorder JSDoc?

bsian03:
Turns out some of the events typings are incorrect, I'm going through them now

Dragory:
The following properties are missing from index.d.ts for `Message`, could be included in this PR:
* `application`
* `messageReference`
* `activity`
* `flags`

Runi-c:
This could go in this PR too:
The documentation on the event `webhooksUpdate` is wrong, `channelID` and `guildID` should be given as properties of `data`, not as separate parameters. 

~Shard.js:1625

bsian03:
PermissionOverwrite#type needs to be strict typed `"role" | "member"`

index.d.ts
```diff
@@ -1831,222 +1862,324 @@ declare namespace Eris {
...
-    deny: number;
-  }
-
   export class PermissionOverwrite extends Permission {
-    id: string;
     createdAt: number;
-    type: string;
+    id: string;
+    type: PermissionType;
     constructor(data: { allow: number; deny: number });
```
jtsshieh: id: string, and type: PermissionType are also passed in `data`

```js
    constructor(data) {
        super(data.allow, data.deny);
        this.id = data.id;
        this.type = data.type;
    }
```

bsian03:
In case you miss my message (not my ping), @jtsshieh suggested using https://github.com/bryanrsmith/eslint-plugin-sort-class-members to enforce the order of which the stuff in classes comes in

jtsshieh:
TS v4 when (or else it'll basically be in like a year because no one updates the TS version)

index.d.ts
```diff
@@ -1008,1044 +1282,986 @@ declare namespace Eris {
...
+    opusFactory: () => OpusScript;
+    volumeLevel: number;
+    constructor(converterCommand: string, opusFactory: OpusScript);
+    addDataPacket(packet: unknown): void;
+    encode(source: string | Stream, options: VoiceResourceOptions): boolean;
+    getDataPacket(): Buffer;
+    reset(): void;
+    resetPackets(): void;
+    setVolume(volume: number): void;
+    stop(e: Error, source: NodeJS.WritableStream): void; // Is this correct? There are too many options this could be
```
bsian03: Need someone to verify this for me, ta

index.d.ts
```diff
@@ -1,271 +1,829 @@
+// Needed for BrowserWebSocket
+/// <reference lib="dom" />
```
alex-taxiera: Using this DOM reference will basically inject web types into every project using eris, polluting globals and causing other issues. We need to look for some alternative here.

index.d.ts
```diff
@@ -1,271 +1,829 @@
...
-  type AnyChannel = AnyGuildChannel | PrivateChannel;
-  type AnyGuildChannel = GuildTextableChannel | VoiceChannel | CategoryChannel | StoreChannel;
-
-  interface CreateInviteOptions {
+  interface CreateChannelInviteOptions {
     maxAge?: number;
     maxUses?: number;
     temporary?: boolean;
-    unique?: boolean;
+    unique?: boolean; }
```
alex-taxiera: Closing bracket should be on separate line.

index.d.ts
```diff
@@ -1,271 +1,829 @@
...
+    argsSplitter?: (str: string) => string[];
+    defaultCommandOptions?: CommandOptions;
+    defaultHelpCommand?: boolean;
+    description?: string;
+    ignoreBots?: boolean;
+    ignoreSelf?: boolean;
+    name?: string;
+    owner?: string;
+    prefix?: string | string[];
   }
```
alex-taxiera, line 198: Are we keeping a blank line between declarations or not?

index.d.ts
```diff
@@ -362,627 +1027,242 @@ declare namespace Eris {
...
-    preferredLocale?: string;
-    afkChannelID?: string;
-    afkTimeout?: number;
-    ownerID?: string;
-    splash?: string;
-    banner?: string;
-    description?: string;
-  }
-  interface OldGuild {
+    integrations: any[]; // TODO ????
```
alex-taxiera: `unknown[]` for now is better.

alex-taxiera (changes requested):
Are the types supposed to be grouped into sections? If so then we should label them with comments where the sections begin and end, simply adding a blank line between sections is not enough.

(This will also let us enforce blank line or no blank line between each declaration, as God intended)

That being said it's not a big deal, worst thing here is the triple slash reference to DOM types.

index.d.ts
jtsshieh: is it possible to have a index.dom.ts and have that as the dom typings?

bsian03: Well we'd need some sort of postinstall/prepare script which prompts the user if they're using a browser, referencing it won't work as it'll import it as full and basically replicate this. I do have some sort of idea which implements this but it's not that great of a workaround since we still need to maintain 2 sets of typings. Caalyx actually suggested to just copy the typings directly from lib.dom.d.ts instead of referencing it so that we wouldn't get global pollution. I'll have a look at this later though

index.d.ts
bsian03: Nice catch, thanks

index.d.ts
bsian03: That group is actually for the Client, hence why that was in its own group and the other command stuff in a separate group, so yes there should be a blank line here.

index.d.ts
bsian03: That didn't actually get picked up by ESLint - should we enable the rule for no explicit any?

DonovanDMC:
An update for webhooks returning Message\<GuildTextableChannel> can be shoved in, considering webhooks only exist in guilds

DonovanDMC:
aaand it hid the rest of what I said, thanks github

bsian03:
That's already been done

index.d.ts
bsian03: Ended up deciding to just ts-ignore the references instead since the types will be referenced if they are using TypeScript in a browser environment

DonovanDMC:
need `OldGuildChannel.position` to be changed to `number` from `string` (I don't know how to suggest changes on pr's, or if I even can)

jtsshieh:
@DonovanDMC pr on bsian's fork for free pr :)

DonovanDMC:
true true (done that 👍)

bsian03:
@DonovanDMC btw it's File changes (whatever the last tab is called), go to index.d.t.s and find OldGuildChannel, then click on the blue "+" and click on that small file icon - it should load the text input with a codeblock prefixed `suggestion`

abalabahaha (approved):
This better work

PR #614, opened 2020-01-15
Can we please add guild to <#message>

LandenStephenss:
Instead of having to do <#messageChannel>.guild why not do <#message>.guild?

https://bee.is-ne.at/Byjvoh.png

PR #994, opened 2020-08-04
Added <Message>.guild

haroongames-git:
Decided we need this.

I think we should add this as some people may be moving here from discord.js (like I did), and wonder why message.guild returns an error.


ghost:
duplicate of #614 #10

ghost:
Also of #425

bsian03:
[Also if you really want Message.guild](https://www.npmjs.com/package/pluris)

PR #995, opened 2020-08-07
Args splitter for CommandClient

### LINKED ISSUE #969: [idea] Allow the use of a custom args splitter for commandclient
This is just a thought but being able to supply our own arguments splitter would be useful since some people might not want to use the current one, which splits both extended spaces and newlines (particularly the newlines)
The proposal would basically add an extra prop into the command client constructor, and if supplied, will be used to split the message content into its args. Obviously string/regexp would be the valid types

**Problems**
Now the only thing I'm concerned about this is if the user has no idea what they're doing and they end up specifying something which doesn't split spaces. I don't believe there is any way to check if the supplied value is valid (it would be extremely cumbersome to do so anyways!) so unintentional stuff may happen. A warning message will probably suffice, something like the one in joinVoiceChannel options.opusOnly, but we'll see

bsian03:
Resolves #969 
Adds a function prop to command client options allowing people to split args however they want it

### DIFF
--- a/lib/command/CommandClient.js
+++ b/lib/command/CommandClient.js
@@ -17,6 +17,7 @@ class CommandClient extends Client {
     * @arg {String} token bot token
     * @arg {Object} [options] Eris options (same as Client)
     * @arg {Object} [commandOptions] Command options
+    * @arg {Function} [argsSplitter] The function used to split args. This passes a string and should return an array of strings. You should not override this unless you know what you are doing
     * @arg {Boolean} [commandOptions.defaultHelpCommand=true] Whether to register the default help command or not
     * @arg {String} [commandOptions.description="An Eris-based Discord bot"] The description to show in the default help command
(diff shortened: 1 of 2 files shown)


lib/command/CommandClient.js
```diff
@@ -17,6 +17,7 @@ class CommandClient extends Client {
     * @arg {String} token bot token
     * @arg {Object} [options] Eris options (same as Client)
     * @arg {Object} [commandOptions] Command options
+    * @arg {Function} [argsSplitter] The function used to split args. This passes a string and should return an array of strings. You should not override this unless you know what you are doing
```
abalabahaha:
```suggestion
    * @arg {Function} [argsSplitter] The function used to split args. The function is given a string with the contents of the command message (without the prefix) and should return an array of strings. By default, args are split by consecutive whitespace
```

abalabahaha (approved):

PR #996, opened 2020-08-08
4014 should not reconnect

zerefdev:
image

Also..
https://discord.com/developers/docs/topics/opcodes-and-status-codes#voice-voice-close-event-codes
image

### DIFF
--- a/lib/voice/VoiceConnection.js
+++ b/lib/voice/VoiceConnection.js
@@ -347,6 +347,8 @@ class VoiceConnection extends EventEmitter {
                     reconnecting = false;
                 } else if(code === 1000) {
                     reconnecting = false;
+                } else if(code === 4014) {
+                    reconnecting = false;
                 }
                 this.disconnect(err, reconnecting);
                 if(reconnecting) {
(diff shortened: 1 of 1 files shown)


lib/voice/VoiceConnection.js
```diff
@@ -347,6 +347,8 @@ class VoiceConnection extends EventEmitter {
                     reconnecting = false;
                 } else if(code === 1000) {
                     reconnecting = false;
+                } else if(code === 4014) {
+                    reconnecting = false;
```
bsian03, line 349: Probably makes more sense if this was put above 1000

HPaulson (dismissed):
This is totally correct, LGTM!

lib/voice/VoiceConnection.js
abalabahaha:
```suggestion
                } else if(code === 4014) {
                    reconnecting = false;
                } else if(code === 1000) {
                    reconnecting = false;
```

abalabahaha (approved):

PR #997, opened 2020-08-12
Allow lib files to be imported directly

ghost:
this pr removes gluten from all brands of eris bread

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
       },
       "./index.js"
     ],
+    "./": "./",
     "./esm": "./esm.mjs"
   },
   "typings": "./index.d.ts",


exoRift:
What does this do, exactly?

Androz2091:
nothing I think

bsian03:
> What does this do, exactly?

Does what it says on the tin. Gotta cater for those who have special dietary requirements!

eritbh:
It appears to allow consumers to import any eris file individually, rather than being limited to the listed entry points ([Reference in node docs](https://nodejs.org/api/esm.html#esm_package_entry_points )) I'm not sure why this change is necessary or what's motivating it.

Edit: the original addition of the exports list prevented people from importing anything from files at all, which can break things for people using eris in bundled environments, and this PR fixes that by allowing any file to be imported from eris (with the caveat that file extensions are required).

abalabahaha (approved):

PR #998, opened 2020-08-13
Calculate Member permissions without hitting the member cache if we already have a Member instance

alex-taxiera:


Runi-c:
have been using this without issue on Tupperbox for a few weeks, necessary for running with no member cache or for getting channel perms of members fetched from REST

abalabahaha (approved):

PR #1001, opened 2020-08-18
Fix shards in browser environments

### LINKED ISSUE #1000: Browser Websockets are not properly supported
If eris is ran in a browser then the [browser's websocket is used in shards](https://github.com/abalabahaha/eris/blob/master/lib/gateway/Shard.js#L15), however the shard still uses the websocket as if it was still using the `ws` module. This causes this to error while connecting showing this error from [here](https://github.com/abalabahaha/eris/blob/6d4668cd134d0af6b28a63f9ba3fa2790e28954d/lib/gateway/Shard.js#L1880):
```
SyntaxError: Failed to construct 'WebSocket': The subprotocol '[object Object]' is invalid.
```
The second argument is actually used for [something else](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket) in browsers, but there are many more changes needed if planned to use the browser's websocket.

Not sure if there is browser-friendly drop-in replacement, but eris in-browser won't work without some slight rewrite of `Shard.js`.

Snazzah:
This is probably the best way to add this change without rewriting `Shard.js`, since Eris in browser is rarely used but still useful.

closes #1000


lib/util/BrowserWebsocket.js
```diff
@@ -0,0 +1,62 @@
...
+        super();
+
+        if(typeof window === "undefined") {
+            new Error("BrowserWebsocket cannot be used outside of a browser environment");
+        }
+
+        this._ws = new window.WebSocket(url);
+        this._ws.onopen = () => this.emit("open");
+        this._ws.onmessage = this._onMessage.bind(this);
+        this._ws.onerror = (event) => this.emit("error", event);
```
abalabahaha: `error` event should be an actual `Error` object. I forget what info the arg has, but a generic `Error` would suffice if it doesn't have useful info

abalabahaha (changes requested):

abalabahaha (approved):

PR #1006, opened 2020-09-01
"connect" event does not have an err parameter

DonovanDMC:
The "connect" event does not have an `err` parameter as the typings suggest.
Ref: https://github.com/abalabahaha/eris/blob/dev/lib/gateway/Shard.js#L1818


bsian03:
duplicate #993

DonovanDMC:
Took 5 years of going through that PR to find it 👌

PR #1007, opened 2020-09-02
Fix editing flags in messages authored by others.

retrixe:
~~This fixes missing Message#flags in TypeScript projects~~ and also editing other people's messages to change flags e.g. if you want to edit another person's message to suppress/un-suppress the embeds in it, Eris causes a permission error since it tries to pass allowedMentions to Discord as well.

Note: I'm not entirely sure if we should pass allowedMentions to edits unless if specified in the first place, so if this needs changing, do reply and I will update the PR asap, which I think we should.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1799,6 +1799,7 @@ declare namespace Eris {
     cleanContent: string;
     roleMentions: string[];
     channelMentions: string[];
+    flags: number;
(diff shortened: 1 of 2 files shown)


bsian03:
TS fixed in #993

abalabahaha (approved):

PR #1014, opened 2020-09-15
Added missing property types (commandAliases, guildPrefixes) for CommandClient

lazukr:
I noticed that CommandClient is missing its .commandAliases and .guildPrefixes properties when trying to use them in Typescript, thus I have added them to index.d.ts.


bsian03:
duplicate #993

lazukr:
Confirmed that this is a duplicate. Closing.

PR #1020, opened 2020-09-21
Remove userID from addReaction

ghost:
This removes the userID parameter from addReaction/addMessageReaction.
* it can be confusing for users ("do I put my client user ID there?")
* this endpoint cannot be used with anything other than the constant text `@me`, so there is no need for user input here.


alex-taxiera (approved):
typings look good

bsian03:
I deprecated this in #993 already

ghost:
oh

PR #1025, opened 2020-10-02
gateway shard: check undocumented field before accessing

wh0:
`user_guild_settings` is not documented in the 'ready' event. https://discord.com/developers/docs/topics/gateway#ready

Most such fields we already check if they exist before calling a method on them or something. This adds that check for `user_guild_settings`, where we previously would call `.forEach` immediately.


alex-taxiera:
This is a deprecated and currently unsupported feature (user accounts)

abalabahaha (approved):
User stuff isn't supported, but this isn't terribly harmful either

PR #1026, opened 2020-10-02
change "reconnect" property in shards to allow auto

turret-dev:
this is so we can allow autoreconnecting


ghost:
Fairly sure #993 covers this.

bsian03:
I concur - also the property should also be optional

PR #1027, opened 2020-10-02
Inline replies and default intents

bsian03:
~~Default to v8 gateway and REST versions, as a result, also~~ add default intents.
Adds inline replies which are not live yet but referenced as per discord/discord-api-docs#2118

EDIT: @Khaaz wanted me to revert back to v7 REST and v6 gateway as he had other stuff planned for v8 which would cause a bit of a mess with this PR


Yamgirly:
Maybe I don't have a proper understanding of intents, but imo defaulting to using all intents (bar the privileged intents) kinda seems like it's undoing the work of intents entirely. I would expect most people would leave this default unless they need the privileged intents, rather than only requesting the intents that they need. As well as that, for people that don't need any intents, I would expect just leaving the parameter blank would be fine, but in this case it would be sending almost all the intents?

bsian03:
Completely get your thought here - I thought the "general" user who had no idea what intents are would benefit from this, hence why the PR exists. Of course people can choose which intents they need, that's up to their own discretion and their fault if something breaks. I just set this up with a "get started quicker and sort it out later" mindset

bsian03:
Going to double check if everything is up to date before I mark this as ready again

lib/structures/Message.js
```diff
@@ -79,6 +80,12 @@ class Message extends Base {
         } else {
             this._client.emit("error", new Error("MESSAGE_CREATE but no message author:\n" + JSON.stringify(data, null, 2)));
         }
+        if(data.referenced_message !== undefined) {
+            this.referencedMessage = this.channel.messages.get(data.referenced_message.id) || new Message(data.referenced_message, this._client);
+        }
+        if(data.referenced_message !== undefined) {
+            this.referencedMessage = this.channel.messages.get(data.referenced_message.id) || new Message(data.referenced_message, this._client);
+        }
```
Dragory, line 88: Duplicated?

bsian03:
Going to split the PR

bsian03:
PR split to the 2 above

PR #1028, opened 2020-10-02
Fix administrator permission not overriding other permissions for @everyone role

ghost:
I have no idea why anyone would ever give the @everyone role administrator permissions, but it got brought up in the dapi channel, so I guess someone is doing it?

Anyway, with this change, `member.permission.has` will correctly return a permission being allowed if the @everyone role has Administrator and the user has no other roles granting them that permission.

### DIFF
--- a/lib/structures/Guild.js
+++ b/lib/structures/Guild.js
@@ -729,7 +729,9 @@ class Guild extends Base {
             return new Permission(Permissions.all);
         } else {
             let permissions = this.roles.get(this.id).permissions.allow;
-            if(permissions & Permissions.all) return new Permission(Permissions.all);
+            if(permissions & Permissions.all) {
+                return new Permission(Permissions.all);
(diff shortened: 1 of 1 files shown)


lib/structures/Guild.js
```diff
@@ -729,6 +729,9 @@ class Guild extends Base {
             return new Permission(Permissions.all);
         } else {
             let permissions = this.roles.get(this.id).permissions.allow;
+            if(permissions & Permissions.all) {
```
bsian03:
```suggestion
            if(permissions & Permissions.administrator) {
```
~~`Permissions.all` checks if literally all permissions are enabled~~ nvm you're checking if the role has *any* permission

ghost: Oops 🙃

abalabahaha (approved):

PR #1033, opened 2020-10-04
Document discovery_splash and features in editGuild

Khaaz:
discord/discord-api-docs#1764

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -612,8 +612,8 @@ declare namespace Eris {
     systemChannelFlags: 0 | 1 | 2;
     systemChannelID?: string;
     verificationLevel?: number;
-    discoverySplash: string;
-    features: string[];
+    discoverySplash?: string;
+    features?: string[];
   }
   interface GuildTemplateOptions {
     name?: string;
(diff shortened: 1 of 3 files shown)


bsian03:
TS docs needed

cyclane:
You should also add `discoverySplash`, `discoverySplashURL` and `dynamicDiscoverySplashURL` to Guild and `discoverySplashURL` to GuildPreview.

bsian03:
ignore above comment, i didn't realise it was for something different

index.d.ts
```diff
@@ -612,6 +612,8 @@ declare namespace Eris {
     systemChannelFlags: 0 | 1 | 2;
     systemChannelID?: string;
     verificationLevel?: number;
+    discoverySplash: string;
+    features: string[];
```
bsian03: Both should be alphabetically positioned and both should also be optional
```suggestion
    discoverySplash?: string;
    features?: string[];
```

index.d.ts
```diff
@@ -612,6 +612,8 @@ declare namespace Eris {
     systemChannelFlags: 0 | 1 | 2;
```
bsian03: Not sure why this isn't optional but may as well fix it here

bsian03 (changes requested):

abalabahaha (approved):

PR #1034, opened 2020-10-04
document lockPermissions and parentID for editChannelPosition

Khaaz:
discord/discord-api-docs#1776

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1412,7 +1412,7 @@ declare namespace Eris {
       type: string,
       reason?: string
     ): Promise<void>;
-    editChannelPosition(channelID: string, position: number): Promise<void>;
+    editChannelPosition(channelID: string, position: number, options: { lockPermissions: string; parentID: string }): Promise<void>;
     editGuild(guildID: string, options: GuildOptions, reason?: string): Promise<Guild>;
     editGuildDiscovery(guildID: string, options?: DiscoveryOptions): Promise<DiscoveryMetadata>;
     editGuildEmoji(
@@ -1897,7 +1897,7 @@ declare namespace Eris {
       type: PermissionType,
       reason?: string
     ): Promise<PermissionOverwrite>;
-    editPosition(position: number): Promise<void>;
+    editPosition(position: number, options: { lockPermissions: string; parentID: string }): Promise<void>;
     getInvites(): Promise<Invite[]>;
     permissionsOf(memberID: string | Member): Permission;
   }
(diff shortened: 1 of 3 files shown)


bsian03:
TS docs needed

index.d.ts
```diff
@@ -1412,7 +1412,7 @@ declare namespace Eris {
       type: string,
       reason?: string
     ): Promise<void>;
-    editChannelPosition(channelID: string, position: number): Promise<void>;
+    editChannelPosition(channelID: string, position: number, options: { lockPermissions: string; parentID: string }): Promise<void>;
```
bsian03: Would be better if we had an interface for the options param

bsian03: Also they should be optional (as well as the other method below)

lib/Client.js
```diff
@@ -918,9 +918,12 @@ class Client extends EventEmitter {
     * Edit a guild channel's position. Note that channel position numbers are lowest on top and highest at the bottom.
     * @arg {String} channelID The ID of the channel
     * @arg {Number} position The new position of the channel
+    * @arg {Object} options Additional options when editing position
+    * @arg {Boolean} options.lockPermissions Whether to sync the permissions with the new parent if moving to a new category
+    * @arg {String} options.parentID The new parent ID (category channel) for the channel that is moved
     * @returns {Promise}
     */
-    editChannelPosition(channelID, position) {
+    editChannelPosition(channelID, position, options) {
```
bsian03: Could probably deprecate the `position` param and move it into `options` like how we do for `createChannel`

lib/Client.js
```diff
@@ -918,9 +918,12 @@ class Client extends EventEmitter {
     * Edit a guild channel's position. Note that channel position numbers are lowest on top and highest at the bottom.
     * @arg {String} channelID The ID of the channel
     * @arg {Number} position The new position of the channel
+    * @arg {Object} options Additional options when editing position
+    * @arg {Boolean} options.lockPermissions Whether to sync the permissions with the new parent if moving to a new category
+    * @arg {String} options.parentID The new parent ID (category channel) for the channel that is moved
```
bsian03: All should be optional

lib/Client.js
bsian03: Default param also needed otherwise this'll throw a typeerror

bsian03 (changes requested):

lib/Client.js
Khaaz: position being mandatory anyway, I think we are good keeping there. Avoid an unecessary breaking change as well.
Missed that default param though ty

abalabahaha (approved):

PR #1036, opened 2020-10-04
Update documentation for options.auth

Khaaz:
Webhooks now by default can use all emoji if the everyone role has the use external emojis permission. This can be safely removed

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -640,11 +640,12 @@ class Client extends EventEmitter {
     }
 
     /**
-    * Execute a webhook
+    * Execute a webhook.
+    * A webhook can use all existing custom emojis as long as the everyone role has the "use external emojis" permission.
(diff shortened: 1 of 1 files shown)


bsian03:
Tbh I'd leave this in case someone removes use external emojis but still wants the webhook to use emojis that a specific bot has access to

ghost:
Bot permissions have no effect on webhook emojis, even if the request is authorized with the bot token.

Khaaz:
Yeah as said above, it's a (potentially unintentional) discord feature that have been removed entirely

Khaaz:
as suggested by evan, we might want to actually keep this feature for larger scale bot that still want to auth with the bot token when sending request to discord.
This PR just precise the documentation.

abalabahaha (approved):

PR #1038, opened 2020-10-04
REST V8, Gateway v8

Khaaz:
Ref: discord/discord-api-docs#2097

- X-Ratelimit-Precision was removed because no longer necessary (all ratelimits are ms by default)
- `guildSubscription` is removed because obsolete since intents.
- Removed deprecated endpoint and methods (GUILD_EMBED)
- use correct invites endpoint
- Permissions will still be numbers internally and are parsed as string when create/edit role.
- Game was removed, replaced by activities[0] for backward compatbility
- DM channel are no longer sent in CHANNEL_CREATE
- `editGuildMember` / `editMember` now returns the Member (discord/discord-api-docs#2505)

BREAKING CHANGES:
- `getGuildEmbed` replaced with `getGuildWidget`
- `PermissionsOverwrite.type` is now a number instead of a String: 
  - editChannelPermission is backward compatible
- member.game was removed, but a getter was ket for backward compatibility

~~This unfortunately invalidate #1016 ?~~

Review and testing is appreciated.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -723,9 +723,9 @@ declare namespace Eris {
     summary: string;
   }
   interface IntegrationOptions {
-    enableEmoticons: string;
-    expireBehavior: string;
-    expireGracePeriod: string;
+    enableEmoticons?: string;
+    expireBehavior?: string;
+    expireGracePeriod?: string;
   }
   interface PruneMemberOptions extends GetPruneOptions {
     computePruneCount?: boolean;
@@ -2124,21 +2124,21 @@ declare namespace Eris {
     application?: IntegrationApplication;
     createdAt: number;
     enabled: boolean;
-    enableEmoticons: boolean;
-    expireBehavior: number;
-    expireGracePeriod: number;
+    enableEmoticons?: boolean;
+    expireBehavior?: number;
+    expireGracePeriod?: number;
     id: string;
     name: string;
-    revoked: boolean;
-    roleID: string;
-    subscriberCount: number;
-    syncedAt: number;
-    syncing: boolean;
+    revoked?: boolean;
+    roleID?: string;
+    subscriberCount?: number;
+    syncedAt?: number;
+    syncing?: boolean;
     type: string;
     user?: User;
     constructor(data: BaseData, guild: Guild);
     delete(): Promise<void>;
-    edit(options: { enableEmoticons: string; expireBehavior: string; expireGracePeriod: string }): Promise<void>;
+    edit(options: IntegrationOptions): Promise<void>;
     sync(): Promise<void>;
   }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -1477,7 +1477,7 @@ declare namespace Eris {
     users: Collection<User>;
     userSettings: UserSettings;
     voiceConnections: VoiceConnectionManager;
-    constructor(token: string, options?: ClientOptions);
+    constructor(token: string, options: ClientOptions);
     acceptInvite(inviteID: string): Promise<Invite<"withoutCount">>;
     addGroupRecipient(groupID: string, userID: string): Promise<void>;
     addGuildDiscoverySubcategory(guildID: string, categoryID: string, reason?: string): Promise<DiscoverySubcategoryResponse>;
@@ -1880,7 +1880,7 @@ declare namespace Eris {
     commands: { [s: string]: Command };
     guildPrefixes: { [s: string]: string | string[] };
     preReady?: true;
-    constructor(token: string, options?: ClientOptions, commandOptions?: CommandClientOptions);
+    constructor(token: string, options: ClientOptions, commandOptions?: CommandClientOptions);
     checkPrefix(msg: Message): string;
     onMessageCreate(msg: Message): Promise<void>;
     onMessageReactionEvent(msg: Message, emoji: Emoji, reactor: Member | Uncached | string): Promise<void>
--- a/index.d.ts
+++ b/index.d.ts
@@ -2218,6 +2218,7 @@ declare namespace Eris {
     defaultAvatar: string;
     defaultAvatarURL: string;
     discriminator: string;
+    game: Activity | null;
     guild: Guild;
     id: string;
     joinedAt: number;
@@ -2375,8 +2376,8 @@ declare namespace Eris {
     unsendMessage(messageID: string): Promise<void>;
   }
 
-  export class Relationship extends Base implements Presence {
-    activities?: Activity[];
+  export class Relationship extends Base implements Omit<Presence, "activities"> {
+    activities: Activity[] | null;
     clientStatus?: ClientStatus;
     id: string;
     status: Status;
--- a/index.d.ts
+++ b/index.d.ts
@@ -929,6 +929,12 @@ declare namespace Eris {
     type: T;
     url?: string;
   }
+  interface ClientPresence {
+    activities: Activity[] | null;
+    afk: boolean;
+    since: number | null;
+    status: Status;
+  }
   interface ClientStatus {
     desktop: Status;
     mobile: Status;
@@ -1455,7 +1461,7 @@ declare namespace Eris {
     lastReconnectDelay: number;
     notes: { [s: string]: string };
     options: ClientOptions;
-    presence: Presence;
+    presence: ClientPresence;
     privateChannelMap: { [s: string]: string };
     privateChannels: Collection<PrivateChannel>;
     ready: boolean;
@@ -2447,7 +2453,7 @@ declare namespace Eris {
     lastHeartbeatSent: number | null;
     latency: number;
     preReady: boolean;
-    presence: Presence;
+    presence: ClientPresence;
     presenceUpdateBucket: Bucket;
     ready: boolean;
     reconnectInterval: number;
--- a/index.d.ts
+++ b/index.d.ts
@@ -205,7 +205,7 @@ declare namespace Eris {
     firstShardID?: number;
     getAllUsers?: boolean;
     guildCreateTimeout?: number;
-    intents?: number | IntentStrings[];
+    intents: number | IntentStrings[];
     largeThreshold?: number;
     lastShardID?: number;
     /** @deprecated */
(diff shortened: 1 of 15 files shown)


lib/structures/GuildAuditLogEntry.js
```diff
@@ -71,11 +71,11 @@ class GuildAuditLogEntry extends Base {
                 this.membersRemoved = +data.options.members_removed;
             }
             if(data.options.type) {
-                if(data.options.type === "member") {
+                if(data.options.type === "1") {
```
DonovanDMC, line 79: jsdoc up there says number, but this is checking for a string
Seems a bit off

```suggestion
                if(data.options.type === 1) {
```

lib/structures/GuildAuditLogEntry.js
```diff
@@ -71,11 +71,11 @@ class GuildAuditLogEntry extends Base {
...
                     this.member = guild.members.get(data.options.id) || {
                         id: data.options.id
                     };
-                } else if(data.options.type === "role") {
+                } else if(data.options.type === "0") {
```
DonovanDMC, line 83: jsdoc up there says number, but this is checking for a string
Seems a bit off

```suggestion
                if(data.options.type === 0) {
```

DonovanDMC:
Just realized by looking back around, I was looking at the wrong thing, and going by what the Discord docs say, those are strings

index.d.ts
```diff
@@ -594,6 +593,27 @@ declare namespace Eris {
...
+    avatar_url: string;
+  }
+  interface WidgetData {
+    id: string;
+    name: string;
+    instant_invite: string;
+    channels: WidgetChannel[];
+    members: WidgetMember[];
+    presence_count: number;
+  }
```
bsian03: The whole of this (not sure why Widget is here) should be below VoiceRegion

lib/Client.js
```diff
@@ -93,7 +93,6 @@ class Client extends EventEmitter {
     * @arg {Number} [options.firstShardID=0] The ID of the first shard to run for this client
     * @arg {Boolean} [options.getAllUsers=false] Get all the users in every guild. Ready time will be severely delayed
     * @arg {Number} [options.guildCreateTimeout=2000] How long in milliseconds to wait for a GUILD_CREATE before "ready" is fired. Increase this value if you notice missing guilds
-    * @arg {Boolean} [options.guildSubscriptions=true] If false, disables some guild subscription events, including typing and presence events. This will reduce processing load, but will also result in inconsistent member caching
```
bsian03, line 96: I'm assuming this is being removed in favour of intents?

lib/Client.js
```diff
@@ -604,14 +602,14 @@ class Client extends EventEmitter {
     * @arg {Boolean} [options.hoist] Whether to hoist the role in the user list or not
     * @arg {Boolean} [options.mentionable] Whether the role is mentionable or not
     * @arg {String} [options.name] The name of the role
-    * @arg {Number} [options.permissions] The role permissions number
+    * @arg {Number | Permission} [options.permissions] The role permissions
```
bsian03: We can allow string here
```suggestion
    * @arg {Number | String | Permission} [options.permissions] The role permissions
```

lib/Client.js
```diff
@@ -901,11 +899,12 @@ class Client extends EventEmitter {
     * @arg {String} overwriteID The ID of the overwritten user or role (everyone role ID = guild ID)
     * @arg {Number} allow The permissions number for allowed permissions
     * @arg {Number} deny The permissions number for denied permissions
-    * @arg {String} type The object type of the overwrite, either "member" or "role"
+    * @arg {Number} type The object type of the overwrite, either 1 for "member" or 0 for "role"
     * @arg {String} [reason] The reason to be displayed in audit logs
     * @returns {Promise}
     */
     editChannelPermission(channelID, overwriteID, allow, deny, type, reason) {
+        type = Number(type);
```
bsian03: This seems useless

lib/Client.js
```diff
@@ -1149,12 +1148,13 @@ class Client extends EventEmitter {
     * @arg {Boolean} [options.hoist] Whether to hoist the role in the user list or not
     * @arg {Boolean} [options.mentionable] Whether the role is mentionable or not
     * @arg {String} [options.name] The name of the role
-    * @arg {Number} [options.permissions] The role permissions number
+    * @arg {Number | Permission} [options.permissions] The role permissions
```
bsian03: Can allow string
```suggestion
    * @arg {Number | String | Permission} [options.permissions] The role permissions
```

lib/Client.js
```diff
@@ -1631,14 +1622,23 @@ class Client extends EventEmitter {
     }
 
     /**
-    * Get a guild's widget object
+    * Get a guild's widget object.
```
bsian03: .
```suggestion
    * Get a guild's widget object
```

lib/structures/Guild.js
```diff
@@ -350,7 +350,7 @@ class Guild extends Base {
     * @arg {Boolean} [options.hoist] Whether to hoist the role in the user list or not
     * @arg {Boolean} [options.mentionable] Whether the role is mentionable or not
     * @arg {String} [options.name] The name of the role
-    * @arg {Number} [options.permissions] The role permissions number
+    * @arg {Number | Permission} [options.permissions] The role permissions
```
bsian03: Can accept string
```suggestion
    * @arg {Number | String | Permission} [options.permissions] The role permissions
```

lib/structures/Guild.js
```diff
@@ -559,7 +559,7 @@ class Guild extends Base {
...
```
bsian03: Can accept string
```suggestion
    * @arg {Number | String | Permission} [options.permissions] The role permissions
```

lib/structures/Member.js
```diff
@@ -159,6 +155,10 @@ class Member extends Base {
         }
     }
 
+    get game() {
+        return this.activities ? this.activities[0] : null;
```
bsian03: Small thing and it doesn't really matter but should we check if the array is empty?

index.d.ts
```diff
@@ -793,7 +812,7 @@ declare namespace Eris {
     hoist?: boolean;
     mentionable?: boolean;
     name?: string;
-    permissions?: number;
+    permissions?: number | Permission;
```
bsian03: Can accept string
```suggestion
    permissions?: number | string | Permission;
```

lib/Client.js
```diff
@@ -1631,14 +1622,23 @@ class Client extends EventEmitter {
...
     * @arg {String} guildID The ID of the guild
     * @returns {Promise<Object>} A guild widget object
     */
     getGuildWidget(guildID) {
         return this.requestHandler.request("GET", Endpoints.GUILD_WIDGET(guildID), true);
     }
 
+    /**
+    * Get a guild's widget settings object. Requires MANAGE_GUILD permission
```
bsian03, line 1839: If we're gonna put required permissions we may as well do it everywhere. Also camelCase

bsian03 (changes requested):

lib/structures/Member.js
Khaaz: it will just returns undefined. Activities is there but it's empty. 🤷🏻‍♂️

lib/Client.js
Khaaz: indeed yes

lib/Client.js
Khaaz: actually yes

Khaaz: changing it to check for "member" / "type" for backward compat

bsian03 (changes requested):
GuildIntegration needs to be updated as well, see discord/discord-api-docs#2175

lib/Client.js
```diff
@@ -953,11 +953,14 @@ class Client extends EventEmitter {
     * @arg {String} overwriteID The ID of the overwritten user or role (everyone role ID = guild ID)
     * @arg {BigInt} allow The permissions number for allowed permissions
     * @arg {BigInt} deny The permissions number for denied permissions
-    * @arg {String} type The object type of the overwrite, either "member" or "role"
+    * @arg {Number} type The object type of the overwrite, either 1 for "member" or 0 for "role"
     * @arg {String} [reason] The reason to be displayed in audit logs
     * @returns {Promise}
     */
     editChannelPermission(channelID, overwriteID, allow, deny, type, reason) {
+        if(isNaN(type)) { // backward compatbility
```
LJNeon: Wouldn't it be better to check `typeof type === "string"` instead? This may have confusing or unintended behavior as is.

LJNeon:
This PR doesn't seem to have any logic for the [detailed errors](https://discord.com/developers/docs/reference#error-messages) added in v8 of the API. I would think we could make some very useful and easy to read errors using that information, and we already have error classes to do that with.

bsian03:
Detailed error handling already exists in DiscordRESTError

LJNeon:
You're right, my bad. I missed that somehow.

index.d.ts
```diff
@@ -684,7 +680,7 @@ declare namespace Eris {
     entries: GuildAuditLogEntry[];
     integrations: GuildIntegration[];
     users: User[];
-    webhooks: Webhook[];
+    webhooks: Webhook[]
```
```suggestion
    webhooks: Webhook[];
```
I'm assuming you didn't mean to remove the semi

lib/Client.js
```diff
@@ -77,7 +77,7 @@ class Client extends EventEmitter {
     /**
     * Create a Client
     * @arg {String} token The auth token to use. Bot tokens should be prefixed with `Bot` (e.g. `Bot MTExIHlvdSAgdHJpZWQgMTEx.O5rKAA.dQw4w9WgXcQ_wpV-gGA4PSk_bm8`). Prefix-less bot tokens are [DEPRECATED]
-    * @arg {Object} [options] Eris options (all options are optional)
+    * @arg {Object} options Eris client options
```
bsian03: I'd note that all options apart from intents are optional instead

bsian03 (dismissed):

abalabahaha (approved):
YOLO

PR #1045, opened 2020-10-05
no channel cache reliance on Message* events

Khaaz:
emit message* events with a partial channel object if no channel in cache.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -519,7 +519,7 @@ class Shard extends EventEmitter {
                 /**
                 * Fired when a cached message is deleted
                 * @event Client#messageDelete
-                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. No other property is guaranteed. If the channel is not cached, channel will be an object with only `id`.
+                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. No other property is guaranteed. If the channel is not cached, channel will be an object with an `id` key.
                 */
                 this.emit("messageDelete", (channel && channel.messages.remove(packet.d)) || {
                     id: packet.d.id,
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -497,7 +497,7 @@ class Shard extends EventEmitter {
                 /**
                 * Fired when a message is updated
                 * @event Client#messageUpdate
-                * @prop {Message} message The updated message. If oldMessage is null, it is recommended to discard this event, since the message data will be very incomplete (only `id` and `channel` are guaranteed). If the channel isn't cached, `channel` will be an object with only `id`.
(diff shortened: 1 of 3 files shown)


lib/gateway/Shard.js
```diff
@@ -456,17 +457,21 @@ class Shard extends EventEmitter {
                     /**
                     * Fired when a message is created
                     * @event Client#messageCreate
-                    * @prop {Message} message The message
+                    * @prop {Message} message The message. Can be partial with only the channel_id if the channel is not cached
```
bsian03: The message itself won't be partial, the channel will, and that should be documented in the Message class

lib/gateway/Shard.js
```diff
@@ -492,7 +497,7 @@ class Shard extends EventEmitter {
                 /**
                 * Fired when a message is updated
                 * @event Client#messageUpdate
-                * @prop {Message} message The updated message. If oldMessage is null, it is recommended to discard this event, since the message data will be very incomplete (only `id` and `channel` are guaranteed)
+                * @prop {Message} message The updated message. If oldMessage is null, it is recommended to discard this event, since the message data will be very incomplete (only `id` and `channel` are guaranteed). If the channel isn't cached, `channel` will be an object with only `id`.
```
bsian03: consistency
```suggestion
                * @prop {Message} message The updated message. If oldMessage is null, it is recommended to discard this event, since the message data will be very incomplete (only `id` and `channel` are guaranteed). If the channel isn't cached, `channel` will be an object with an `id` key.
```

lib/gateway/Shard.js
```diff
@@ -510,17 +515,17 @@ class Shard extends EventEmitter {
...
                 const channel = this.client.getChannel(packet.d.channel_id);
-                if(!channel) {
-                    break;
-                }
+
                 /**
                 * Fired when a cached message is deleted
                 * @event Client#messageDelete
-                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. No other property is guaranteed
+                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. No other property is guaranteed. If the channel is not cached, channel will be an object with only `id`.
```
bsian03: consistency
```suggestion
                * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. No other property is guaranteed. If the channel is not cached, channel will be an object with an `id` key.
```

bsian03 (changes requested):

lib/structures/Message.js
```diff
@@ -12,7 +12,7 @@ const User = require("./User");
 * @prop {Object?} application The application of the activity in the message
 * @prop {Array<Object>} attachments Array of attachments
 * @prop {User} author The message author
-* @prop {PrivateChannel | TextChannel | NewsChannel} channel The channel the message is in
+* @prop {PrivateChannel | TextChannel | NewsChannel} channel The channel the message is in.Can be partial with only the id if the channel is not cached.
```
```suggestion
* @prop {PrivateChannel | TextChannel | NewsChannel} channel The channel the message is in. Can be partial with only the id if the channel is not cached.
```

bsian03 (changes requested):

abalabahaha (approved):

PR #1050, opened 2020-10-19
Fix doc for webhooksUpdate event

Snazzah:
The documentation for the `webhooksUpdate` event showed that there was data and two IDs rather than the two IDs being in the data object.


Snazzah:
Fixed in #993 [(commit)](https://github.com/bsian03/eris/commit/50e9e6551768f9b42b3596a9f71c2e475e396fce)

PR #1053, opened 2020-10-23
add: Message.createReply

arthur-rl:
Creates a reply to a message that mentions the author of the message.
```js
        message.createReply("This is a reply!")
```
img

### DIFF
--- a/lib/structures/Message.js
+++ b/lib/structures/Message.js
@@ -375,6 +375,22 @@ class Message extends Base {
         return this._client.crosspostMessage.call(this._client, this.channel.id, this.id);
     }
 
+    /**
+    * Create a reply to the message
+    * @arg {String | Object} content A string.
+    * @arg {Object} [file] A file object
+    * @arg {Buffer} file.file A buffer containing file data
+    * @arg {String} file.name What to name the file
+    * @returns {Promise<Message>}
+    */
+   createReply(content, file) {
+        if (typeof content == "string") {
+            return this._client.createMessage.call(this._client, this.channel.id, `<@${this.author.id}>, ${content}`, file);
+        } else {
+            console.log(new Error(`Content must be string, not ${typeof content}`));
+        }
+    }   
+
     toJSON(props = []) {
         return super.toJSON([
             "attachments",


ghost:
Why?

arthur-rl:
> 
> 
> Why?

why not?

ghost:
because you don't... just add random stuff to a library for no reason? if you want it in your fork, that's fine, but it has no place in the library, at least in my opinion

arthur-rl:
> 
> 
> because you don't... just add random stuff to a library for no reason? if you want it in your fork, that's fine, but it has no place in the library, at least in my opinion

I mean it's quite useful. otherwise you'll have to do 
```js
message.channel.createMessage(`<@${message.author.id}> blah`)
```
In my opinion its easier
```js
message.createReply("blah")
```

Snazzah:
There is really no need for a change that is just prepending a mention to a message. Just because D.JS does it doesn't make it needed here.

DonovanDMC:
Eris isn't meant to hold your hand with a lot of things. You can implement this yourself on your side by either editing the code, or adding it to the Message prototype.
```ts
Object.defineProperty(Eris.Message.prototype, "reply", {
	value: async function(content) {
		return this.channel.createMessage(`<@!${this.author.id}>, ${content}`);
	}
});
```

I haven't fully tested this, but it should work. You will need to do a lot more to support what the normal createMessage supports (i.e. embeds, allowMentions, files, etc...)

arthur-rl:
> 
> 
> Eris isn't meant to hold your hand with a lot of things. You can implement this yourself on your side by either editing the code, or adding it to the Message prototype.
> 
> ```ts
> Object.defineProperty(Eris.Message.prototype, "reply", {
> 	value: async function(content) {
> 		return this.channel.createMessage(`<@!${this.author.id}>, ${content}`);
> 	}
> });
> ```
> 
> I haven't fully tested this, but it should work. You will need to do a lot more to support what the normal createMessage supports (i.e. embeds, allowMentions, files, etc...)

Thanks. I will add this to my code. closing the PR now.

PR #1054, opened 2020-10-25
Redo invite to allow for non strict compilation (and other doc stuff)

bsian03:
This one is for @DonovanDMC because I was feeling nice

EDIT: also the following
- Undeprecated messageReactionRemove apart from in PrivateChannel cuz i forgot to
- Put the change in #972 back from a merge conflict which I didn't resolve properly


index.d.ts
```diff
@@ -16,7 +16,7 @@ declare namespace Eris {
   type AnyGuildChannel = GuildTextableChannel | VoiceChannel | CategoryChannel | StoreChannel;
   type ChannelTypes = Constants["ChannelTypes"][keyof Constants["ChannelTypes"]];
   type GuildTextableChannel = TextChannel | NewsChannel;
-  type InviteChannel = InvitePartialChannel | Exclude<AnyGuildChannel, CategoryChannel>;
+  type InviteChannel = InvitePartialChannel | Exclude<AnyGuildChannel, CategoryChannel> | GroupChannel;
```
DonovanDMC: @bsian03 this is where the GroupChannel bit is

DonovanDMC (dismissed):

index.d.ts
```diff
@@ -1867,19 +1845,26 @@ declare namespace Eris {
     dynamicSplashURL(format?: ImageFormat, size?: number): string;
   }
 
-  export class Invite extends Base {
-    channel: InvitePartialChannel | Exclude<AnyGuildChannel, CategoryChannel>;
+  // If CT (count) is null, it is an invite with metadata, if boolean, it is an invite without metadata
+  export class Invite<CT extends boolean | null, CH extends InviteChannel = InviteChannel> extends Base {
```
alex-taxiera: probably better idea to use a string instead of true/false/null for CT

also CT could have a default to clean up some other code

alex-taxiera (changes requested):

alex-taxiera (dismissed):

DonovanDMC (dismissed):

abalabahaha (changes requested):
Unrelated changes should be in a separate PR

abalabahaha (approved):

PR #1058, opened 2020-10-30
Fix non-api errors permanently removing remaining requests

bsian03:
If anyone wants to reproduce this issue, best way is to execute a guild ban with an invalid (easiest would be `\n`) character in the reason. Then check the relevant rate limit and notice that `remaining` will never replenish unless you restart the process or delete the rate limit. 
This will also eventually cause the bucket to hang, stopping any further requests for that particular endpoint(s). In the case of guild ban, this will be any requests after the invalid audit log reason request.


abalabahaha (changes requested):
Can't say I'm fond of this, a token should be reserved ASAP after the `if()`. Error handling would be more appropriate.

Also seems like `this.processing` should get similar treatment.

lib/rest/RequestHandler.js
```diff
@@ -60,6 +61,10 @@ class RequestHandler {
         const _stackHolder = {}; // Preserve async stack
         Error.captureStackTrace(_stackHolder);
 
+        if(body && this.invalidHeaderRegex.test(body.reason)) {
```
zerefdev, line 65: you might want to check if body.reason exists

bsian03: `undefined` always returns false, so not needed

bsian03:
I'm wanting to see if I can find a direct solution I can implement into the bucket itself in case an error comes from somewhere other than audit log reasons. If anyone has a rough idea feel free to ping me in the discord channel

lib/rest/RequestHandler.js
```diff
@@ -57,10 +58,37 @@ class RequestHandler {
...
+        return this._request(method, url, auth, body, file, route, short).then(
+            (res) => res,
+            (rej) => {
+                if(!(rej instanceof DiscordHTTPError || rej instanceof DiscordRESTError)) {
+                    ++this.ratelimits[route].remaining;
+                }
+                throw rej;
+            }
+        );
+    }
```
bsian03: I'm wondering if it's better to just chain this back into the original `request` method or if it's fine

lib/rest/RequestHandler.js
```diff
@@ -57,10 +58,37 @@ class RequestHandler {
...
+    _request(method, url, auth, body, file, _route, short) {
+        const route = _route || this.routefy(url, method);
 
         const _stackHolder = {}; // Preserve async stack
         Error.captureStackTrace(_stackHolder);
 
+        if(body && this.invalidHeaderRegex.test(body.reason)) {
+            throw new TypeError("Invalid character in audit log reason");
+        }
+
```
bsian03, line 68: Should I keep this for clarity

abalabahaha:
Re-implemented less hackily in 6083e09

PR #1060, opened 2020-11-03
Support editing/deleting webhook messages

bsian03:
https://github.com/discord/discord-api-docs/issues/1544#issuecomment-708707645


Cubox:
I was about to implement this. Thanks for doing it!

Cubox:
I tested this PR fully, works flawlessly.

Cubox (dismissed):

lib/Client.js
```diff
@@ -2233,6 +2233,34 @@ class Client extends EventEmitter {
...
     }
 
+    /**
+    * Delete a webhook message
+    * @arg {String} webhookID
+    * @arg {String} token
+    * @arg {String} messageID
+    * @returns {Promise}
+    */
+    webhookDeleteMessage(webhookID, token, messageID) {
```
apacheli: `*WebhookMessage` would make more sense right? At least grammatically. This does seem weird though since `executeWebhook` wouldn't be consistent in naming.

lib/structures/Message.js
```diff
@@ -392,6 +392,33 @@ class Message extends Base {
         return this._client.unpinMessage.call(this._client, this.channel.id, this.id);
     }
 
+    /**
+    * Delete the message as a webhook
+    * @param {String} token The token of the webhook
+    * @returns {Promise}
+    */
+    webhookDelete(token) {
```
apacheli: Nitpick, but imo it would be better to merge this into the `delete` method. A little bit of ternary magic seems nicer.

bsian03: Laziness :-)
I'll ask abal about this and see if he wants it in a single method

lib/Client.js
bsian03: Yes but then you get the case in message where the method is called `deleteWebhook` which makes it sound like you're deleting a webhook that belongs to that message (but then this point will be invalidated if `webhookDelete` merges into `delete` method)

lib/structures/Message.js
abalabahaha: User needs to check if a token should be provided anyway, better to separate functions separate

Khaaz (approved):

abalabahaha (approved):

PR #1061, opened 2020-11-04
feat: Hide client property 

Wilbert-mad:
As projects get bigger or you're ~debugging~ logging it might be hard to have to scroll over as the client is logged multiple times.
This should only affect logging it will not show the client but it will still be a  property of the class.
allowing for easy navigation during logging.

### Exampal of large data
```js
Message {
...data,
Client: {}
channel: {
 data...
 Client: {}
 }
author: {
 data...
 Client: {}
 }
guild: {
 data...
 Client: {}
 }
}
```

### DIFF
--- a/lib/structures/Guild.js
+++ b/lib/structures/Guild.js
@@ -60,7 +60,7 @@ const {Permissions} = require("../Constants");
 class Guild extends Base {
     constructor(data, client) {
         super(data.id);
-        this._client = client;
+        Object.defineProperty(this, "_client", { value: client });
         this.shard = client.shards.get(client.guildShardMap[this.id]);
         this.unavailable = !!data.unavailable;
         this.joinedAt = Date.parse(data.joined_at);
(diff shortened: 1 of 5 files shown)


ghost:
You will encounter this in anything everywhere. It’s much more preferred for each individual to tune their debugger for their own needs instead of modifying offending code.

For an individual like you it would be much better option to put more effort into learning how to use debugging tools provided to you.

bsian03:
https://nodejs.org/api/util.html#util_util_inspect_object_options

Wilbert-mad:
I am well aware. Yet I don't think a lot of people would know how to debug as they would maybe resort to console logging there arguments.

apacheli:
This could probably be done with the inspector's custom inspect used in the [Base](https://github.com/abalabahaha/eris/blob/dev/lib/structures/Base.js#L16) class.

abalabahaha:
There are a lot more back-references besides `client`; some people want some props to show up in logging, others don't. Options listed above are preferred.

For people who don't know how to debug, it's never too late to start 👀

PR #1062, opened 2020-11-05
Add basic support for stickers

Dragory:
Now updated for 0.14.0. No support for posting stickers, just sticker data in Message objects.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1945,6 +1945,7 @@ declare namespace Eris {
     prefix?: string;
     reactions: { [s: string]: unknown; count: number; me: boolean };
     roleMentions: string[];
+    stickers: Sticker[];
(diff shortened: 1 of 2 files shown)


Eight8-7020 (approved):

abalabahaha (changes requested):
Might as well make a class for these

Khaaz:
Navigating through PR right now, I see this haven't been updated with a class as asked so I'm upping it.
For that matter I don't personally think it's that necessary since one or 2 other structures could have been classes like Emoji or webhooks and are not.

abalabahaha:
This PR and #1113 both do things slightly differently, and both seem to have small bugs. I've merged them together as 9cd3a5d

PR #1065, opened 2020-11-07
Remove `[key: string]: number` from certain TS typings.

oathompsonjones:
Unless there is something I'm not seeing here, these typings are not needed and could do more harm than good. 
Using `[key: string]: number` allows properties with any key, which in this case is incorrect as all known properties are explicitly listed, and there are no unknown properties to account for.
As such, these lines should be removed to prevent any unexpected behaviour.


bsian03:
#1057

abalabahaha:
Dupes #1057

oathompsonjones:
Oops, my bad.

PR #1067, opened 2020-11-08
Add System Channel flags

jtsshieh:
https://discord.com/developers/docs/resources/guild#guild-object-system-channel-flags

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -363,6 +363,7 @@ declare namespace Eris {
     region: string;
     rulesChannelID?: string;
     splash?: string;
+    systemChannelFlags: 0 | 1 | 2;
     systemChannelID?: string;
     verificationLevel: 0 | 1 | 2 | 3 | 4;
   }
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -363,6 +363,7 @@ declare namespace Eris {
     region: string;
     rulesChannelID?: string;
     splash?: string;
+    systemChannelFlags: 0 | 1 | 2;
```
bsian03, line 366: Could probably add these as a type alias so we only need to edit 1 place if we have to

jtsshieh: Verification level, explicit content filter, and default notifications are also number unions, but only in OldGuild, not the actual Guild structure or GuildOptions. Should systemChannelFlags be changed to `number` in those places, or would it be better to change those into type aliases along with systemChannelFlags and have them strictly typed.

bsian03: If that's the case then I'd just leave it for now and have a separate PR making that change all in one go

abalabahaha (approved):

PR #1068, opened 2020-11-09
Fix types for Message.reactions

Dragory:


apacheli:
Include a fix for the JSDoc as well: https://github.com/abalabahaha/eris/blob/master/lib/structures/Message.js#L36

Dragory:
The JSDoc is still accurate. I don't think any of the other "one-off" objects like this are documented with detail either (e.g. Message.activity), so leaving it as `Object` should be fine.

apacheli:
> The JSDoc is still accurate. I don't think any of the other "one-off" objects like this are documented with detail either (e.g. Message.activity), so leaving it as `Object` should be fine.

The two properties below it are not accurate is what I was referring to. (Sorry if I caused any confusion with the link)
```
* @prop {Object} reactions An object containing the reactions on the message
* @prop {Boolean} reactions.me Whether or not the bot user did the reaction
* @prop {Number} reactions.count The number of times the reaction was used
```

Dragory:
Oh, you're right! Sorry, I completely missed that. I imagine just removing the `.me` and `.count` would be the cleanest in this case, unless there's some way to document them properly.

Dragory:
The JSDoc for Message is now fixed as well.

lib/structures/Message.js
```diff
@@ -34,8 +34,6 @@ const User = require("./User");
 * @prop {Boolean} pinned Whether the message is pinned or not
 * @prop {String?} prefix The prefix used in the Message, if any (CommandClient only)
 * @prop {Object} reactions An object containing the reactions on the message
```
bsian03: Probably best if `me` and `count` info was explained here for better clarification instead of just completely removing it

abalabahaha (approved):

PR #1070, opened 2020-11-17
Default intents

bsian03:
See #1027 
Adds default intents for when v8 is mandatory (everything apart from guildMembers and guildPresences is enabled by default)


apacheli:
I disagree with this PR. Adviath, a fairly trusted user, made a good comment about not having default intents.
https://github.com/discordjs/discord.js/issues/3924#issuecomment-656429545

PR #1076, opened 2020-11-22
Removes the "undefined" behaviour in debug event

auguwu:
This PR removes the `undefined` behaviour when using the `debug` event in **RequestHandler**, it can be in-consice when debugging.

## Before
before

## After
(with any sort of body)
after (with body)

(without any sort of body)
after (without body)

### DIFF
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -257,7 +257,7 @@ class RequestHandler {
                         }
 
                         if(resp.statusCode !== 429) {
-                            this._client.emit("debug", `${body && body.content} ${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
+                            this._client.emit("debug", `${body && body.content ? body.content + " " : ""}${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
                         }
 
                         if(resp.statusCode >= 300) {
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -257,12 +257,14 @@ class RequestHandler {
                         }
 
                         if(resp.statusCode !== 429) {
-                            this._client.emit("debug", `${body && body.content ? `${body.content} ` : ""}${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
+                            const content = typeof body === "object" && body.content ? `${body.content} ` : "";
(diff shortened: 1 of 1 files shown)


DonovanDMC (dismissed):

lib/rest/RequestHandler.js
```diff
@@ -257,7 +257,7 @@ class RequestHandler {
                         }
 
                         if(resp.statusCode !== 429) {
-                            this._client.emit("debug", `${body && body.content} ${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
+                            this._client.emit("debug", `${body && body.content ? `${body.content} ` : ""}${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
```
bsian03: Surely you'd stringify the body object, no?

bsian03 (commented):
You've also missed the bit about 5 lines below

lib/rest/RequestHandler.js
auguwu: Ah, I didn't check for that! I'll fix it in a bit. Didn't know I should of stringified the object before...

auguwu: I may of mis-interpreted what you were trying to say now I think about it, no I haven't stringified the `body` object since I don't think it was really needed to do so and could cause some faulty type errors in the class.

bsian03: Whether it's needed is debatable, for the errors, just check `typeof body === "object"` and use it in a ternary expression

lib/rest/RequestHandler.js
```diff
@@ -257,12 +257,14 @@ class RequestHandler {
                         }
 
                         if(resp.statusCode !== 429) {
-                            this._client.emit("debug", `${body && body.content} ${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
+                            const content = typeof body === "object" && body.content ? `${body.content} ` : "";
```
bsian03: Just check for `body`, not `body.content` (both ternary and actual event)

lib/rest/RequestHandler.js
```diff
@@ -257,12 +257,14 @@ class RequestHandler {
...
+                            this._client.emit("debug", `${content}${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
                         }
 
                         if(resp.statusCode >= 300) {
                             if(resp.statusCode === 429) {
-                                this._client.emit("debug", `${resp.headers["x-ratelimit-global"] ? "Global" : "Unexpected"} 429 (╯°□°）╯︵ ┻━┻: ${response}\n${body && body.content} ${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
+                                const content = typeof body === "object" && body.content ? `${body.content} ` : "";
```
bsian03: Same thing here

lib/rest/RequestHandler.js
auguwu: Gotcha! 👍

DonovanDMC (approved):

abalabahaha (approved):
TBH we should do JSON at some point, but this is OK as-is

PR #1083, opened 2020-11-23
Capture more custom inspect locations

bsian03:
For classes which don't extend Base and have props prefixed with `_` (searched with `this\._\w+ =`)

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -1,5 +1,6 @@
 "use strict";
 
+const util = require("util");
 const Base = require("../structures/Base");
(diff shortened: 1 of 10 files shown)


apacheli:
It'd be easier to just define an `inspect` method on Base and have `util.inspect.custom` call that instead. Ones that don't have `util.inspect.custom` could just call `Base.inspect` directly.

Edit: Actually, my proposal made no sense. It just seemed a bit clunky but what have right now seems to be only way to do it.

abalabahaha (approved):

PR #1085, opened 2020-11-25
Move messageReactionAdd reactor to 4th arg, fix some stuff

bsian03:
- `userID` is back in the event in the third arg - this should fix some issues with command client reaction buttons (as an example) and also reverse the breaking change in 0.14.0. `reactor` param is now in the 4th arg
- Member data will now always update when the event is received instead of adding the data if the member is not cached.
- If the reaction was not in a guild, `reactor` will now be null. Partial object data no longer exists in the event

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -557,6 +557,31 @@ declare namespace Eris {
     systemChannelID: string;
     verificationLevel?: number;
   }
+  interface DiscoveryCategory {
+    id: number;
+    name: {
+      default: string;
+      localizations?: { [lang: string]: string };
+    };
+    is_primary: boolean;
+  }
+  interface DiscoveryMetadata {
+    guild_id: string;
+    primary_category_id: number;
+    keywords: string[] | null;
+    emoji_discoverability_enabled: boolean;
+    category_ids: number[];
+  }
+  interface DiscoveryOptions {
+    primaryCategoryID?: string;
+    keywords?: string[];
+    emojiDiscoverabilityEnabled?: boolean;
+    reason?: string;
+  }
+  interface DiscoverySubcategoryResponse {
+    guild_id: string;
+    category_id: number;
+  }
   interface GetPruneOptions {
     days?: number;
     includeRoles?: string[];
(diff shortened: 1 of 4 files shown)


bsian03:
Just for context, @Skillz4Killz and I had a discussion about the implications of this PR
I'll leave him to say what his points are but here's mine:
- Consistency with `messageReactionRemove` (indirectly fixing a CC bug)
- (not my main point for the PR) Fixing aforementioned bug with CC directly will cause some weird logic to be applied from the source and further down chain, if not then we'll have duplicates
- Duplicate ID doesn't matter since I won't see a way of you need to use both member object and userID in the same use case, in which you can just ignore one of the two
- Since it's reversing a breaking change, people who couldn't update to 0.14.0 (i know a few people that couldn't for these breaking changes) will be able to update and be able to treat it as a feature addition

Skillz4Killz:
As i mentioned, the entire point of the CC bug is fixed by adding 1 line in the CommandClient.
```ts
if (typeof userID !== 'string') userID = userID.id;
```

Doing this is **NOT** a breaking change but it fixes a bug to anyone having it. Forcing a breaking change for this reason is just a bad choice. If there is a bug in CC, it should be fixed in CC without enforcing a breaking change.

- The last point of reversing a breaking change which lets people update is also invalid imo because this itself is a breaking change after months of stability. So anyone that has updated now has this same problem of not updating for this breaking change? So if i say hey this pr is the reason i won't update eris, will we reverse this again in x months so that I can update later? How long does this pattern carry on for to doing breaking changes just so people can update? 

- Consistency isn't a valid reason to make this change because Discord themself don't have consistency for these two events. Remove does not have member sent on it but add does. This is similar to like saying hey let's add a breaking change to `guildMemberAdd` where it goes from `(guild, member)` to `(guild, partial, member)` to be consistent with `guildMemberRemove`

- `(message, emoji, uselessArgHereWhichIDontWantBecauseIHaveIDInsideMember, member)` is much worse than `(message, emoji, member)`

---

I like the change for the caching improvement. But that can be done without a breaking change that encourages bad behavior.

PR #1089, opened 2020-11-27
unify error handling

zerefdev:


abalabahaha:
The error is thrown synchronously in direct response to calling `play()`. The `error` event is mainly for asynchronous errors or errors that shouldn't interrupt the current execution flow

PR #1094, opened 2020-11-29
Fix deprecations

bsian03:


lib/structures/TextChannel.js
```diff
@@ -37,7 +37,7 @@ class TextChannel extends GuildChannel {
     * Add a reaction to a message
     * @arg {String} messageID The ID of the message
     * @arg {String} reaction The reaction (Unicode string if Unicode emoji, `emojiName:emojiID` if custom emoji)
-    * @arg {String} [userID="@me"] The ID of the user to react as. Passing this parameter is deprecated and will not be supported in future versions.
+    * @arg {String} [userID="@me"] The ID of the user to react as.
```
alex-taxiera: shouldn't this param be removed? this overload in the d.ts is still marked as deprecated.

bsian03: I'm honestly not sure why I removed that warning. Unless @abalabahaha wants me to remove it completely then I don't really see a point of code/jsdoc saying one thing and typescript saying another

bsian03: I was dumm

abalabahaha (approved):

PR #1097, opened 2020-11-30
Fix some typescript mistakes

bsian03:


index.d.ts
```diff
@@ -1243,16 +1243,21 @@ declare namespace Eris {
   }
 
   export class Client extends EventEmitter {
-    bot?: boolean;
+    bot: boolean;
```
alex-taxiera, line 1249: I think there is jsdoc to update with this as well

abalabahaha (approved):

PR #1101, opened 2020-12-12
Fix onMessageReactionEvent

bsian03:


lib/command/CommandClient.js
```diff
@@ -210,7 +210,8 @@ class CommandClient extends Client {
         }
     }
 
-    async onMessageReactionEvent(msg, emoji, userID) {
+    async onMessageReactionEvent(msg, emoji, reactor) {
+        const userID = reactor.id || reactor;
```
alex-taxiera: the types indicate that `reactor` will be either `Member` or `{ id: string }`. is the `|| reactor` necessary?

bsian03: Yes, reaction remove only passes the user ID

alex-taxiera: Cool, then the types should be adjusted to reflect that :^)

alex-taxiera: `reactor` in `messageReactionAdd` is `Member | { id: string })`

but here `reactor` in `onMessageReactionEvent` is `Member | { id: string }) | string`

bsian03: Yeah and `messageReactionRemove` is `string`, so what's currently there is correct

abalabahaha (approved):

PR #1102, opened 2020-12-14
Added user attribute to oldMember

cyclane:
Added a `user` attribute to OldMember of type PartialUser since the event `guildMemberUpdate` is triggered when user attributes are updated as well, meaning that sometimes there is no difference between `member` and `oldMember`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -447,7 +447,7 @@ declare namespace Eris {
     (event: "guildMemberRemove", listener: (guild: Guild, member: Member | MemberPartial) => void): T;
     (
       event: "guildMemberUpdate",
-      listener: (guild: Guild, member: Member, oldMember: { nick?: string; premiumSince: number; roles: string[] } | null) => void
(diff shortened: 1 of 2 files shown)


Khaaz (approved):

index.d.ts
```diff
@@ -447,7 +447,7 @@ declare namespace Eris {
     (event: "guildMemberRemove", listener: (guild: Guild, member: Member | MemberPartial) => void): T;
     (
       event: "guildMemberUpdate",
-      listener: (guild: Guild, member: Member, oldMember: { nick?: string; premiumSince: number; roles: string[] } | null) => void
+      listener: (guild: Guild, member: Member, oldMember: { nick?: string; premiumSince: number; roles: string[]; user: PartialUser | null } | null) => void
```
bsian03, line 450: `oldMember` could probably have its own interface tbh

abalabahaha:
What use cases would this enable? Generally it's up to the listeners to filter update events for differences, since listeners usually look for changes in specific props, and events can be emitted without prop changes regardless of this specific case. Also not quite a fan of the additional memory churn this would create.

PR #1103, opened 2020-12-15
Expose application info on ready

bsian03:


macdja38:
Ik this is new, was it added to gateway v7 as well as v8?

bsian03:
Looks like all versions, I'm running my test bot on v6 gateway and it seems to be populated there

lib/Client.js
```diff
@@ -50,6 +50,7 @@ const sleep = (ms) => new Promise((res) => setTimeout(res, ms));
 /**
 * Represents the main Eris client
 * @extends EventEmitter
+* @prop {Object} application Object containing the bot application's ID and its public flags
```
alex-taxiera: the types indicate this should be optional

```suggestion
* @prop {Object} [application] Object containing the bot application's ID and its public flags
```

Khaaz (dismissed):

abalabahaha (approved):

PR #1106, opened 2020-12-17
Type this for instances which use itself

bsian03:
See #626 for context
Will require TypeScript 4.2 to be released before this will work


bsian03:
Targeting 23 February

bsian03:
Imagine forgetting to mark the PR as ready after making the commit

abalabahaha (approved):

PR #1107, opened 2020-12-17
Correct documentation for subcommandAliases

bsian03:


index.d.ts
```diff
@@ -1620,7 +1620,7 @@ declare namespace Eris {
     reactionButtonTimeout: number;
     requirements: CommandRequirements;
     restartCooldown: boolean;
-    subcommandAliases: { [alias: string]: Command };
+    subcommandAliases: { [alias: string]: string };
```
alex-taxiera, line 1623: could `Record<string, string>` be used instead?

bsian03: (on mobile) is it used anywhere else?

bsian03: It turns out we do a mix of both for some reason

abalabahaha (approved):

PR #1109, opened 2020-12-19
Implemented role tags

cyclane:
Implemented [role tags](https://discord.com/developers/docs/topics/permissions#role-object-role-tags-structure) into the Role class as attributes `botID`, `integrationID` and `premiumSubscriber`.
- Converted the premiumSubscriber to a boolean which makes more sense than keeping it as null type

### DIFF
--- a/lib/structures/Role.js
+++ b/lib/structures/Role.js
@@ -110,13 +110,16 @@ class Role extends Base {
 
     toJSON(props = []) {
         return super.toJSON([
+            "botID",
             "color",
             "hoist",
+            "integrationID ",
             "managed",
             "mentionable",
             "name",
             "permissions",
             "position",
+            "premiumSubscriber ",
             ...props
         ]);
     }
--- a/lib/structures/Role.js
+++ b/lib/structures/Role.js
@@ -5,10 +5,12 @@ const Permission = require("./Permission");
 
 /**
 * Represents a role
+* @prop {String?} botID The id of the bot this role belongs to
 * @prop {Number} color The hex color of the role in base 10
 * @prop {Number} createdAt Timestamp of the role's creation
 * @prop {Boolean} hoist Whether users with this role are hoisted in the user list or not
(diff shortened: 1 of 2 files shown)


lib/structures/Role.js
```diff
@@ -92,13 +110,16 @@ class Role extends Base {
 
     toJSON(props = []) {
         return super.toJSON([
+            "botID",
             "color",
             "hoist",
+            "integrationID ",
```
bsian03:
```suggestion
            "integrationID",
```

lib/structures/Role.js
```diff
@@ -92,13 +110,16 @@ class Role extends Base {
...
             "managed",
             "mentionable",
             "name",
             "permissions",
             "position",
+            "premiumSubscriber ",
```
bsian03:
```suggestion
            "premiumSubscriber",
```

bsian03 (changes requested):
Extra spaces you added here

cyclane:
Thank you for pointing those out, did not notice I added them

lib/structures/Role.js
```diff
@@ -47,6 +50,21 @@ class Role extends Base {
...
         }
+        if(data.tags !== undefined) {
+            if(data.tags.bot_id !== undefined) {
+                this.botID = data.tags.bot_id;
+            }
+            if(data.tags.integration_id !== undefined) {
+                this.integrationID = data.tags.integration_id;
+            }
+            if(data.tags.premium_subscriber !== undefined) {
+                this.premiumSubscriber = true;
```
Khaaz, line 61: any particular reason for making this a top level field instead of being consistent with discord here?

abalabahaha (changes requested):
Agreed, more of a fan of preserving the tags object for future proofing

abalabahaha:
Implemented in 824ccf7

Issue #1108, opened 2020-12-17
Add reaction message


Skeeww:
Hello,

I'm working on a bot and I want to react to a message but I can't find methods that are not deprecated like addReaction() or addMessageReaction()

![image](https://user-images.githubusercontent.com/9639992/102519101-5cb99080-4092-11eb-9879-7ae912ed9928.png)

Thanks for your help

ghost:
Those methods are not deprecated. The overload with `userID` property is deprecated. You can continue to use these methods with no issue.

Skeeww:
Ah okay thank you for your quick response

bsian03:
It seems like, for at least VSC, that the order overloads come in is literally how it's listed in the file. As you're filling in the parameters, you will be able to scroll through each of the overloads but not when intellisense is making a prediction of what you want (it selects the first in the list). It was something I hadn't considered, but something which could be improved on VSC's end also. I'll PR this at some point in the future to stop deprecated overloads appearing by default

PR #1111, opened 2020-12-20
Show non deprecated overloads by default

bsian03:
ref. #1108 
I've reordered deprecated overloads so that IDEs will show non deprecated overloads first. This should not affect anything at all


abalabahaha (dismissed):
I'll take your word for it

abalabahaha (approved):

PR #1119, opened 2020-12-29
Remove deprecated markers on non-deprecated removeReaction

cyyynthia:
According to the documentation, using removeReaction with a user ID is not deprecated everywhere (thankfully!) but the typings reflect otherwise, since all available methods are marked as deprecated.

I looked through the typings and removed the deprecated annotations on methods that aren't documented as deprecated in the source code. I left the ones actually documented as deprecated with the annotation.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1559,7 +1559,6 @@ declare namespace Eris {
     ): Promise<number>;
     removeGroupRecipient(groupID: string, userID: string): Promise<void>;
     removeGuildMemberRole(guildID: string, memberID: string, roleID: string, reason?: string): Promise<void>;
-    /** @deprecated */
     removeMessageReaction(channelID: string, messageID: string, reaction: string, userID: string): Promise<void>;
     removeMessageReaction(channelID: string, messageID: string, reaction: string): Promise<void>;
     removeMessageReactionEmoji(channelID: string, messageID: string, reaction: string): Promise<void>;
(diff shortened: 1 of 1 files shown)


ghost:
This issue is fixed by #1111.

intrnl:
> This issue is fixed by #1111.

I don't think it does, the `@deprecated` tag is still there.

[Here][1], [here][2], and [here][3]

[1]: https://github.com/abalabahaha/eris/pull/1111/commits/bc7e2a75951c3e4f49897ab7c992fe17960f15f7#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8R1563
[2]: https://github.com/abalabahaha/eris/pull/1111/commits/bc7e2a75951c3e4f49897ab7c992fe17960f15f7#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8L2144
[3]: https://github.com/abalabahaha/eris/pull/1111/commits/bc7e2a75951c3e4f49897ab7c992fe17960f15f7#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8R2322

bsian03:
If you look carefully, supplying userID is deprecated. Calling the method is not. My PR is correct.

bsian03:
Actually, wrong PR was linked. Is actually #1094 - either way, still duplicate.

cyyynthia:
My bad, PR #1094 indeed already addresses this. I gave a quick look by searching for "deprecated" to see if it wasn't already PR'd, but it didn't show up, I should've been more cautious! I'll close this as this is indeed a duplicate PR.

PR #1127, opened 2021-01-17
More accurate typings. (Message class and Emoji interface)

oathompsonjones:
Before:
```ts
let message: Message;
let member = message.member; // This is Member | null.
let guildID = message.guildID; // This is string | undefined.
if (message.channel instanceof TextChannel) {
    message = message as Message<TextChannel>; // Have to do this, but it is safe to do so.
    member = member as Member; // member is still treated as Member | null without this.
    guildID = guildID as string; // guildID is still treated as string | undefined without this.
} else {
    member = member as null; // member is still treated as Member | null without this.
    guildID = guildID as undefined; // guildID is still treated as string | undefined without this.
}
```
After:
```ts
let message: Messaage;
const member = message.member; // This is Member | null.
const guildID = message.guildID; // This is string | undefined.
if (message.channel instanceof TextChannel) {
    message = message as Message<TextChannel>; // Still have to do this, but it is still safe to do so.
    // member is now treated as Member automatically.
    // guildID is now treated as string automatically.
} else {
    // member is now treated as null automatically.
    // guildID is now treated as undefined automatically.
}
```

On an unrelated note, `Message#addReaction(reaction: string, userID: string): Promise<void>;` and `Message#removeReaction(reaction: string, userID: string): Promise<void>;` (and all equivalent methods where usedID is one of the parameters) are both marked at deprecated for all messages. Reading the previous commits, unless I've missed something, it looks like these should only be deprecated for messages in private channels, rather than for every message.
Could someone let me know if this is the case, or if I have missed something somewhere.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1934,10 +1934,10 @@ declare namespace Eris {
     editedTimestamp?: number;
     embeds: Embed[];
     flags: number;
-    guildID: T extends GuildChannel ? string : undefined;
+    guildID: T extends GuildTextable ? string : undefined;
     id: string;
     jumpLink: string;
-    member: T extends GuildChannel ? Member : null;
+    member: T extends GuildTextable ? Member : null;
     mentionEveryone: boolean;
     mentions: User[];
     messageReference: MessageReference | null;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1937,7 +1937,7 @@ declare namespace Eris {
     guildID: T extends GuildChannel ? string : undefined;
     id: string;
     jumpLink: string;
-    member: T extends GuildChannel ? Member : Member | null;
+    member: T extends GuildChannel ? Member : null;
     mentionEveryone: boolean;
     mentions: User[];
     messageReference: MessageReference | null;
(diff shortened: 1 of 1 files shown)


DonovanDMC:
You can use `instanceof GuildChannel`, or narrowing with the `type` property.
```ts
if(message.channel instanceof Eris.GuildChannel) {}
```

```ts
if (message.channel.type === Eris.Constants.ChannelTypes.GUILD_TEXT || message.channel.type === Eris.Constants.ChannelTypes.GUILD_NEWS) {}
```

oathompsonjones:
Also now includes some Emoji typings which needed correcting to be consistent with Discord's documentation.

index.d.ts
```diff
@@ -315,11 +315,12 @@ declare namespace Eris {
   // Emoji
   interface Emoji extends EmojiBase {
     animated: boolean;
+    available: boolean;
```
bsian03, line 318: This is in #1096 already

bsian03 (changes requested):
L1952 should also be covered here (`webhookID`)

abalabahaha (approved):

PR #1130, opened 2021-01-28
Incorrect Usage Embed Fix

TornadoMC2:
Incorrect usage message wasn't working with a function returning an embed so I changed it to match the other error message options

Before:
```js
if(args.length === 0) {
            if(shouldDelete) {
                msg.delete();
            }
            if(this.argsRequired) {
                if(this.hooks.postCheck) {
                    const response = await Promise.resolve(this.hooks.postCheck(msg, args, true));
                    if(response) {
                        msg = response.msg || msg;
                        args = response.args || args;
                    }
                }
                reply = typeof this.invalidUsageMessage === "function" ? this.invalidUsageMessage(msg) : this.invalidUsageMessage;
                if(reply) {
                    msg.channel.createMessage(reply.replace(/%prefix%/g, msg.prefix).replace(/%label%/g, this.fullLabel));
                }
                return;
            }
            return this.executeCommand(msg, args);
        }
```

After:
```js
if(args.length === 0) {
            if(shouldDelete) {
                msg.delete();
            }
            if(this.argsRequired) {
                if(this.hooks.postCheck) {
                    const response = await Promise.resolve(this.hooks.postCheck(msg, args, true));
                    if(response) {
                        msg = response.msg || msg;
                        args = response.args || args;
                    }
                }
                reply = typeof this.invalidUsageMessage === "function" ? this.invalidUsageMessage(msg) : this.invalidUsageMessage;
                if(reply) {
                    msg.channel.createMessage(reply);
                }
                return;
            }
            return this.executeCommand(msg, args);
        }
```


jtsshieh:
This would be a breaking change that breaks the default invalid usage message: https://github.com/abalabahaha/eris/blob/master/lib/command/CommandClient.js#L125

TornadoMC2:
> This would be a breaking change that breaks the default invalid usage message: https://github.com/abalabahaha/eris/blob/master/lib/command/CommandClient.js#L125

Oh, didn't see that, i'll fix that real quick

Before:
```js
if(args.length === 0) {
            if(shouldDelete) {
                msg.delete();
            }
            if(this.argsRequired) {
                if(this.hooks.postCheck) {
                    const response = await Promise.resolve(this.hooks.postCheck(msg, args, true));
                    if(response) {
                        msg = response.msg || msg;
                        args = response.args || args;
                    }
                }
                reply = typeof this.invalidUsageMessage === "function" ? this.invalidUsageMessage(msg) : this.invalidUsageMessage;
                if(reply) {
                    msg.channel.createMessage(reply);
                }
                return;
            }
            return this.executeCommand(msg, args);
        }
```
After:
```js
if(args.length === 0) {
            if(shouldDelete) {
                msg.delete();
            }
            if(this.argsRequired) {
                if(this.hooks.postCheck) {
                    const response = await Promise.resolve(this.hooks.postCheck(msg, args, true));
                    if(response) {
                        msg = response.msg || msg;
                        args = response.args || args;
                    }
                }
                reply = typeof this.invalidUsageMessage === "function" ? this.invalidUsageMessage(msg) : this.invalidUsageMessage.replace(/%prefix%/g, msg.prefix).replace(/%label%/g, this.fullLabel);
                if(reply) {
                    msg.channel.createMessage(reply);
                }
                return;
            }
            return this.executeCommand(msg, args);
        }
```

bsian03:
Relevant JSDoc and TypeScript documentation will require updating to reflect the change as well

TornadoMC2:
> Relevant JSDoc and TypeScript documentation will require updating to reflect the change as well

The docs already say `A string or a function that returns a string to show when a command was improperly used.  The function is passed the Message object as a parameter.` But I was getting an error when I was using a function to return an embed. Returning an embed was working for permissionMessage and errorMessage, and they have the same description in the docs. Tell me if i'm missing something here.

jtsshieh:
> returns a string to show when a command was improperly used

TornadoMC2:
> returns a string to show when a command was improperly used

Well yes, but on the other ones you can use a function that returns an embed. I feel like they should make it so that returning an embed or string is allowed

bsian03:
Which is what you're doing... so the documentation should be updated?

TornadoMC2:
updated docs to show that using a function to return an embed works

TornadoMC2:
Finished updating docs in all parts where the function for an invalid/error message can return an embed.

### Places Updated

- permissionMessage
- cooldownMessage
- errorMessage
- Subcommand#cooldownMessage
- Subcommand#errorMessage
- Subcommand#invalidUsageMessage
- Subcommand#permissionMessage

lib/command/Command.js
```diff
@@ -369,9 +369,9 @@ class Command {
                         args = response.args || args;
                     }
                 }
-                reply = typeof this.invalidUsageMessage === "function" ? this.invalidUsageMessage(msg) : this.invalidUsageMessage;
+                reply = typeof this.invalidUsageMessage === "function" ? this.invalidUsageMessage(msg) : this.invalidUsageMessage.replace(/%prefix%/g, msg.prefix).replace(/%label%/g, this.fullLabel);
                 if(reply) {
-                    msg.channel.createMessage(reply.replace(/%prefix%/g, msg.prefix).replace(/%label%/g, this.fullLabel));
+                    msg.channel.createMessage(reply);
```
bsian03: We should probably add support for replacing templates in the embed in relevant locations

bsian03 (commented):
TypeScript docs still need updating btw

abalabahaha:
Also seems to break functions that were expecting replacements in their returns. It should handle the returned message if it's a string or `return.content` if it's an object and `.content` exists

TornadoMC2:
> Also seems to break functions that were expecting replacements in their returns. It should handle the returned message if it's a string or `return.content` if it's an object and `.content` exists

Ok, I think I understand. Could I have an example of one of these functions to do some testing?

abalabahaha:
It's just being passed as the content arg of createMessage(). So `() => "foo"`, `() => ({ content: "foo" })`, `() => ({ embed: ..., content: "foo" })`. etc.

TornadoMC2:
> It's just being passed as the content arg of createMessage(). So `() => "foo"`, `() => ({ content: "foo" })`, `() => ({ embed: ..., content: "foo" })`. etc.

oooh I see what you mean

TornadoMC2:
Updated so that it will replace %prefix% and %label% in functions that return a string or an object with the content property

Before:
```js
if(args.length === 0) {
            if(shouldDelete) {
                msg.delete();
            }
            if(this.argsRequired) {
                if(this.hooks.postCheck) {
                    const response = await Promise.resolve(this.hooks.postCheck(msg, args, true));
                    if(response) {
                        msg = response.msg || msg;
                        args = response.args || args;
                    }
                }
                reply = typeof this.invalidUsageMessage === "function" ? this.invalidUsageMessage(msg) : this.invalidUsageMessage.replace(/%prefix%/g, msg.prefix).replace(/%label%/g, this.fullLabel);
                if(reply) {
                    msg.channel.createMessage(reply);
                }
                return;
            }
            return this.executeCommand(msg, args);
        }
```
After: 
```js
if(args.length === 0) {
            if(shouldDelete) {
                msg.delete();
            }
            if(this.argsRequired) {
                if(this.hooks.postCheck) {
                    const response = await Promise.resolve(this.hooks.postCheck(msg, args, true));
                    if(response) {
                        msg = response.msg || msg;
                        args = response.args || args;
                    }
                }
                if(typeof this.invalidUsageMessage === "function") {
                    if(typeof this.invalidUsageMessage(msg) === "string") {
                        reply = this.invalidUsageMessage(msg).replace(/%prefix%/g, msg.prefix).replace(/%label%/g, this.fullLabel)
                    } else {
                        reply = "content" in this.invalidUsageMessage(msg) ? this.invalidUsageMessage(msg)["content"].replace(/%prefix%/g, msg.prefix).replace(/%label%/g, this.fullLabel) : this.invalidUsageMessage(msg)
                    }
                } else {
                    reply = this.invalidUsageMessage.replace(/%prefix%/g, msg.prefix).replace(/%label%/g, this.fullLabel)
                }
                if(reply) {
                    msg.channel.createMessage(reply);
                }
                return;
            }
            return this.executeCommand(msg, args);
        }
```

abalabahaha (approved):

PR #1140, opened 2021-02-26
Check if guild exist in cache

### LINKED ISSUE #1139: TypeError: Cannot destructure property `emojis` of 'undefined' or 'null'.
Guild.emojis may be undefined in GUILD_EMOJIS_UPDATE.

`eris/lib/gateway/Shard.js in Shard.wsEvent at line 1595:45`

```js
case "GUILD_EMOJIS_UPDATE": {
  const guild = this.client.guilds.get(packet.d.guild_id);
  const {emojis: oldEmojis} = guild;
  guild.update(packet.d);
```

zerefdev:
Should fix #1139

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -1965,6 +1965,10 @@ class Shard extends EventEmitter {
             }
             case "GUILD_EMOJIS_UPDATE": {
                 const guild = this.client.guilds.get(packet.d.guild_id);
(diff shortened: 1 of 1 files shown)


Khaaz:
you rather should emit the event with partial than entirely block it.

abalabahaha:
Partial event implemented in 16c6a76

PR #1141, opened 2021-03-05
Prevent 401s if token is not prefixed

bsian03:


abalabahaha:
Breaks use of OAuth bearer tokens

bsian03:
Couldn't we just account for that as well?

abalabahaha:
And possible future prefixes.

PR #1143, opened 2021-03-12
adding support for type 20 (Interaction message)

flamableassassin:
despite the docs saying otherwise type 20 is still sent on V6


flamableassassin:
Reworking due to discord changing the format for an interaction msg

PR #1146, opened 2021-03-23
getChannel/getRESTChannel/getGuildChannels generic

DonovanDMC:


bsian03:
I feel like type asserting would be preferable instead of handling generics

DonovanDMC:
being able to do `<Client>.getChannel<Eris.TextChannel>("").createMessage()` instead of `(<Client>.getChannel("") as Eris.TextChannel).createMessage()` is easier and looks nicer, in my opinion

bsian03:
Or you can do `<TextableChannel>client.getChannel("")` which does the same thing

See [here](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) for info

DonovanDMC:
I know that one exists, but 
a.) most typescript projects (I've seen at least) enforce `as`
b.) that looks a bit ugly

DonovanDMC:
Discussed in Discord, disregarded due to lib consistency.

PR #1149, opened 2021-03-24
Add premiumType

HeadTriXz:
premiumType can be used to check what type of nitro a user has. It can either return 0, 1 or 2.

> 0: None
> 1: Nitro Classic
> 2: Nitro

### DIFF
--- a/lib/structures/User.js
+++ b/lib/structures/User.js
@@ -139,6 +139,7 @@ class User extends Base {
             "avatar",
             "bot",
             "discriminator",
+            "premiumType",
             "publicFlags",
             "system",
             "username",
--- a/lib/structures/User.js
+++ b/lib/structures/User.js
@@ -14,6 +14,7 @@ const Endpoints = require("../rest/Endpoints");
 * @prop {String} discriminator The discriminator of the user
 * @prop {String} id The ID of the user
 * @prop {String} mention A string that mentions the user
+* @prop {Number?} premiumType The type of nitro the user has
 * @prop {Number?} publicFlags Publicly vicible flags for this user
 * @prop {String} staticAvatarURL The URL of the user's avatar (always a JPG)
 * @prop {Boolean} system Whether the user is an official Discord system user (e.g. urgent messages)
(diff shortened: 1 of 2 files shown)


abalabahaha (dismissed):

lib/structures/User.js
```diff
@@ -41,6 +42,9 @@ class User extends Base {
         if(data.discriminator !== undefined) {
             this.discriminator = data.discriminator;
         }
+        if(data.premium_type !== undefined) {
+            this.premiumType = data.premium_type;
```
bsian03, line 46: Can you also add this to the `toJSON` method at the bottom of the file

bsian03 (changes requested):

lib/structures/User.js
HeadTriXz: Done!

lib/structures/User.js
```diff
@@ -14,7 +14,8 @@ const Endpoints = require("../rest/Endpoints");
 * @prop {String} discriminator The discriminator of the user
 * @prop {String} id The ID of the user
 * @prop {String} mention A string that mentions the user
-* @prop {Number?} publicFlags Publicly visible flags for this user
+* @prop {Number?} premiumType The type of nitro the user has
+* @prop {Number?} publicFlags Publicly vicible flags for this user
```
bsian03: Fix the typo here
```suggestion
* @prop {Number?} publicFlags Publicly visible flags for this user
```

bsian03 (changes requested):
Just so you know, all PRs should be made with the dev branch to the dev branch -  the typo was actually fixed in dev in a previous PR

HeadTriXz:
> Just so you know, all PRs should be made with the dev branch to the dev branch - the typo was actually fixed in dev in a previous PR

Yeah, I noticed. My bad. Would you like me to create a new PR using the dev branch or is it fine like this?

bsian03:
It's fine here, just mainly to prevent awkward conflicts from happening if you're not on the latest commit

abalabahaha (approved):

PR #1153, opened 2021-03-29
Assign default rest options within Client

conorwastakenwastaken:
The default properties for the new `rest` options get defined in the Request Handler and doesn't change the actual `client.options` object. The [requestGuildMembers](https://github.com/abalabahaha/eris/blob/dev/index.d.ts#L2295) function still tries to use this object, so that won't work properly for the user unless they manually define a timeout (it doesn't return the Array of members found).

I left the default options in the RequestHandler file because I wasn't sure if they were there for a reason, I can move them to Client.js if you think that's necessary.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -491,7 +491,7 @@ class Shard extends EventEmitter {
             timeout: setTimeout(() => {
                 res(this.requestMembersPromise[opts.nonce].members);
                 delete this.requestMembersPromise[opts.nonce];
-            }, (options && options.timeout) || (this.client.options.rest && this.client.options.rest.requestTimeout) || this.client.options.requestTimeout)
+            }, (options && options.timeout) || this.client.options.rest.requestTimeout || this.client.options.requestTimeout)
         });
     }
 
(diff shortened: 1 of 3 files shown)


lib/gateway/Shard.js
```diff
@@ -491,7 +491,7 @@ class Shard extends EventEmitter {
             timeout: setTimeout(() => {
                 res(this.requestMembersPromise[opts.nonce].members);
                 delete this.requestMembersPromise[opts.nonce];
-            }, (options && options.timeout) || this.client.options.requestTimeout)
+            }, (options && options.timeout) || (this.client.options.rest && this.client.options.rest.requestTimeout) || this.client.options.requestTimeout)
```
bsian03: Checking if `this.client.options.rest` exists isn't necessary (at least I don't believe it is)

conorwastakenwastaken: Oh yeah tru, forgot it gets defined as an object with no props by default. Changed. :)

abalabahaha:
> doesn't change the actual `client.options` object

Intentional, handling of REST options is moved to RequestHandler.

I committed a version of this without the `client.options.rest` stuff as 7a5ec43

conorwastakenwastaken:
Yeah, but the `requestGuildMembers` fn still uses the deprecated `client.options.requestTimeout` property. The deprecated values aren't set to their default values anymore, so the `timeout` options prop wouldn't be optional anymore.

abalabahaha:
I half-un-deprecated it for gateway purposes, but forgot to add the default back, thanks. af6c663

PR #1154, opened 2021-03-31
More coverage of Stage Channels

bsian03:
Only thing missing is the `REQUEST_TO_SPEAK` permission as instructed by abal since the permission is currently under development and may be removed in the future, and the fact that it is `1 << 32` which doesn't work with JS and as such breaks Eris' styling.

Adds onto 5ae9f52 3750145 because this is abal:
effort

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2349,6 +2349,7 @@ declare namespace Eris {
   }
 
   export class StageChannel extends VoiceChannel {
+    topic?: string;
     type: 13;
   }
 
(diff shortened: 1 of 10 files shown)


alex-taxiera:
since `StageChannel extends VoiceChannel`, are the `VoiceChannel | StageChannel` unions necessary? it wasn't added to `AnyGuildChannel` (which might make sense if it's not needed).

alex-taxiera:
> since `StageChannel extends VoiceChannel`, are the `VoiceChannel | StageChannel` unions necessary? it wasn't added to `AnyGuildChannel` (which might make sense if it's not needed).

looks like we did the same thing for NewsChannel so this is fine - but should probably be extracted to a standalone union type like `GuildTextableChannel`

abalabahaha (approved):

PR #1178, opened 2021-04-13
Workaround voice disconnect at switch channel

james58899:
Discord using same close code on kick and change voice server (switch channel will change voice server).
We need wait `VOICE_SERVER_UPDATE` to know it is kick or change voice server.

This will fix 4014 and 4006 cause connection disconnect but will cause `disconnect` event delay 5s.

Workaround https://github.com/discord/discord-api-docs/issues/2450

### DIFF
--- a/lib/voice/VoiceConnection.js
+++ b/lib/voice/VoiceConnection.js
@@ -115,6 +115,7 @@ class VoiceConnection extends EventEmitter {
         this.sequence = 0;
         this.timestamp = 0;
         this.ssrcUserMap = {};
+        this.disconnectTimeout;
 
         this.nonce = Buffer.alloc(24);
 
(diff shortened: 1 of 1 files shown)


lib/voice/VoiceConnection.js
```diff
@@ -115,6 +115,7 @@ class VoiceConnection extends EventEmitter {
         this.sequence = 0;
         this.timestamp = 0;
         this.ssrcUserMap = {};
+        this.disconnectTimeout;
```
bsian03: Prefix this maybe? Only nitpick I have

bsian03 (commented):
Discord gotta break stuff

lib/voice/VoiceConnection.js
abalabahaha:
```suggestion
        this.disconnectTimeout = null;
```

lib/voice/VoiceConnection.js
```diff
@@ -388,14 +389,22 @@ class VoiceConnection extends EventEmitter {
...
+                    this.channelID = null;
+                    this.updateVoiceState();
+                    /**
+                    * Fired when the voice connection disconnects
+                    * @event VoiceConnection#disconnect
+                    * @prop {Error?} error The error, if any
+                    */
+                    this.emit("disconnect", error);
+                }
+            }, 5000);
```
abalabahaha:
```suggestion
            }, this.shard.client ? this.shard.client.options.connectionTimeout : 30000);
```

abalabahaha (dismissed):

abalabahaha:
On second thought, I think it would be better to just treat this like a new connection (but keeping most state). I have some test commits on dev branch, mind stopping by on Discord?

james58899:
Tested dev branch, everything is good.

PR #1180, opened 2021-04-14
Deprecate getMessageReaction before param

bsian03:
Discord no longer supports supplying this param

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -146,6 +146,7 @@ declare namespace Eris {
       messageID: string,
       reaction: string,
       limit?: number,
+      /** @deprecated */
       before?: string,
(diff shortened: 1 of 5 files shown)


Brayzure:
If Discord no longer supports it, is deprecation the way to go as opposed to removal? I figured we'd deprecate stuff that was going to be removed/unsupported, not things that already have been.

bsian03:
Outright removal may have unintended side effects for users either using both params or only `after`. I ran this through with abal and he said a deprecation would be preferable

abalabahaha (approved):

abalabahaha:
Accidentally rebased as 3e86cd7

PR #1187, opened 2021-04-19
Move to content.messageReference

bsian03:
This deprecates `content.messageReferenceID` in favour of `content.messageReference` which is more flexible, allowing an object as well as a string

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -48,7 +48,7 @@ declare namespace Eris {
     content?: string;
     embed?: EmbedOptions;
     flags?: number;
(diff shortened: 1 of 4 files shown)


DonovanDMC (dismissed):

index.d.ts
```diff
@@ -48,6 +48,8 @@ declare namespace Eris {
     content?: string;
     embed?: EmbedOptions;
     flags?: number;
+    messageReference?: string | MessageReferenceReply;
```
```suggestion
    messageReference?: MessageReferenceReply;
```

abalabahaha (approved):

PR #1189, opened 2021-04-22
Expose components

bsian03:
discord/discord-api-docs#3007 discord/discord-api-docs#3200 https://discord.com/developers/docs/interactions/message-components

Adds support for components.  Type 2 Style 5 opens an external URL, the rest of them sends an interaction request.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -83,7 +83,7 @@ declare namespace Eris {
   type ConverterCommand = "./ffmpeg" | "./avconv" | "ffmpeg" | "avconv";
 
   // Webhook
-  type MessageWebhookContent = Pick<WebhookPayload, "content" | "embeds" | "file" | "allowedMentions">;
+  type MessageWebhookContent = Pick<WebhookPayload, "content" | "embeds" | "file" | "allowedMentions" | "components">;
 
   // INTERFACES
   // Internals
@@ -1032,6 +1032,7 @@ declare namespace Eris {
     allowedMentions?: AllowedMentions;
     auth?: boolean;
     avatarURL?: string;
+    components?: ActionRow[];
     content?: string;
     embeds?: EmbedOptions[];
     file?: MessageFile | MessageFile[];
--- a/index.d.ts
+++ b/index.d.ts
@@ -62,8 +62,9 @@ declare namespace Eris {
     messageReferenceID?: string;
     tts?: boolean;
   };
+  type ActionRowComponents = Button | Dropdown;
   type Button = InteractionButton | URLButton;
-  type Component = ActionRow | Button;
+  type Component = ActionRow | ActionRowComponents;
   type ImageFormat = "jpg" | "jpeg" | "png" | "gif" | "webp";
   type MessageContent = string | AdvancedMessageContent;
   type MFALevel = 0 | 1;
@@ -818,7 +819,7 @@ declare namespace Eris {
 
   // Message
   interface ActionRow {
-    components: Button[];
+    components: ActionRowComponents[];
     type: 1;
   }
   interface ActiveMessages {
@@ -847,6 +848,21 @@ declare namespace Eris {
     emoji?: Partial<PartialEmoji>;
     label?: string;
     type: 2;
+  }
+  interface Dropdown {
+    custom_id: string;
+    max_values?: number;
+    min_values?: number;
+    options: DropdownOptions;
+    placeholder?: string;
+    type: 3;
+  }
+  interface DropdownOptions {
+    default?: boolean;
+    description?: string;
+    emoji?: Partial<PartialEmoji>;
+    label: string;
+    value: number | string;
   }
   interface GetMessageReactionOptions {
     after?: string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -62,7 +62,7 @@ declare namespace Eris {
     messageReferenceID?: string;
     tts?: boolean;
   };
-  type Button = InteractionButtonComponent | URLButtonComponent;
+  type Button = InteractionButton | URLButton;
   type Component = ActionRow | Button;
   type ImageFormat = "jpg" | "jpeg" | "png" | "gif" | "webp";
   type MessageContent = string | AdvancedMessageContent;
@@ -817,6 +817,10 @@ declare namespace Eris {
   }
 
   // Message
+  interface ActionRow {
+    components: Button[];
+    type: 1;
+  }
   interface ActiveMessages {
     args: string[];
     command: Command;
@@ -838,15 +842,11 @@ declare namespace Eris {
     url: string;
     width?: number;
   }
-  interface ButtonComponentBase {
+  interface ButtonBase {
     disabled?: boolean;
     emoji?: PartialEmoji;
     label?: string;
     type: 2;
-  }
-  interface ActionRow {
-    type: 1;
-    components: Button[];
   }
   interface GetMessageReactionOptions {
     after?: string;
@@ -854,7 +854,7 @@ declare namespace Eris {
     before?: string;
     limit?: number;
   }
-  interface InteractionButtonComponent extends ButtonComponentBase {
+  interface InteractionButton extends ButtonBase {
     custom_id: string;
     style: 1 | 2 | 3 | 4;
   }
@@ -901,7 +901,7 @@ declare namespace Eris {
     pack_id: string;
     tags?: string;
   }
-  interface URLButtonComponent extends ButtonComponentBase {
+  interface URLButton extends ButtonBase {
     style: 5;
     url: string;
   }
--- a/index.d.ts
+++ b/index.d.ts
@@ -844,7 +844,7 @@ declare namespace Eris {
   }
   interface ButtonBase {
     disabled?: boolean;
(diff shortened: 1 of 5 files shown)


HeadTriXz:
Replace componenets with components.

index.d.ts
```diff
@@ -53,6 +53,7 @@ declare namespace Eris {
   // Message
   type AdvancedMessageContent = {
     allowedMentions?: AllowedMentions;
+    component?: CollectionComponent[];
```
HeadTriXz:
```suggestion
    components?: ActionRow[];
```

index.d.ts
```diff
@@ -61,6 +62,8 @@ declare namespace Eris {
     messageReferenceID?: string;
     tts?: boolean;
   };
+  type ButtonComponent = InteractionButtonComponent | URLButtonComponent;
+  type Component = CollectionComponent | ButtonComponent;
```
HeadTriXz:
```suggestion
  type Component = ActionRow | ButtonComponent;
```

index.d.ts
```diff
@@ -835,12 +838,24 @@ declare namespace Eris {
     url: string;
     width?: number;
   }
+  interface ButtonComponentBase {
+    label: string;
+    type: 2;
+  }
```
HeadTriXz:
```suggestion
  interface ButtonComponentBase {
    disabled?: boolean;
    emoji?: PartialEmoji;
    label?: string;
    type: 2;
  }
```

index.d.ts
```diff
@@ -835,12 +838,24 @@ declare namespace Eris {
...
+  interface CollectionComponent {
+    type: 1;
+    components: Component[];
+  }
```
HeadTriXz:
```suggestion
  interface ActionRow {
    type: 1;
    components: ButtonComponent[];
  }
```
ActionRow may not contain another ActionRow.

index.d.ts
```diff
@@ -2229,6 +2248,7 @@ declare namespace Eris {
     /** @deprecated */
     cleanContent: string;
     command?: Command;
+    components: CollectionComponent[];
```
HeadTriXz:
```suggestion
    components?: ActionRow[];
```

HeadTriXz (changes requested):
Docs are finally officially released, I suggested some changes and make sure to update the JSDocs.

index.d.ts
bsian03: Going to rename ButtonComponent as well

```suggestion
  type Button = InteractionButtonComponent | URLButtonComponent;
  type Component = ActionRow | Button;
```

index.d.ts
bsian03:
```suggestion
  interface ActionRow {
    type: 1;
    components: Button[];
  }
```

DonovanDMC (dismissed):
Worked completely when I tested in my code

R3dlessX (dismissed):
Good Job !

Linker-123:
What about components option in `Client#messageCreate` ?

index.d.ts
```diff
@@ -814,6 +817,10 @@ declare namespace Eris {
   }
 
   // Message
+  interface ActionRow {
+    components: Component[];
```
HeadTriXz, line 821:
```suggestion
    components: Exclude<Component, ActionRow>[];
```

The above is future proof, or simply just use `Button[]`. ActionRows may not contain another ActionRow.

HeadTriXz (changes requested):

index.d.ts
HeadTriXz:
```suggestion
    components: Exclude<Component, ActionRow>[];
```

This is future proof, you could also simply use Button[] for now though. ActionRows may not include another ActionRow.

HeadTriXz (changes requested):

index.d.ts
bsian03: I'll just revert it, Discord needs to do their documentation better

index.d.ts
```diff
@@ -835,12 +843,37 @@ declare namespace Eris {
...
+    disabled?: boolean;
+    emoji?: Partial<PartialEmoji>;
+    label?: string;
+    type: 2;
+  }
+  interface Dropdown {
+    custom_id: string;
+    max_values?: number;
+    min_values?: number;
+    options: DropdownOptions;
```
HeadTriXz:
```suggestion
    options: DropdownOptions[];
```

HeadTriXz (changes requested):

JustCat80 (changes requested):
`editWebhookMessage(webhookID, token, messageID, options) {
        if(!options.content && !options.embeds && !options.file) {
            return Promise.reject(new Error("No content, embeds or file"));
        }}`
        
        In client should be changed to 
`editWebhookMessage(webhookID, token, messageID, options) {
        if(!options.content && !options.embeds && !options.components && !options.file) {
            return Promise.reject(new Error("No content, embeds, components, or file"));
        }}`

bsian03:
No

JustCat80:
> No

Why not?

bsian03:
Because you still do need one of the 3 existing to send a component

JustCat80:
Not when editing a message

bsian03:
You should use the relevant interaction edit message endpoint for that instead

JustCat80:
Hm? could you elaborate?

HeadTriXz:
> Hm? could you elaborate?
> 

https://discord.com/developers/docs/interactions/slash-commands#responding-to-an-interaction

JustCat80:
> > Hm? could you elaborate?
> 
> https://discord.com/developers/docs/interactions/slash-commands#responding-to-an-interaction

Yea i realize now, that only works for the initial response, after that you use webhooks to do anything else

andromika:
> You should use the relevant interaction edit message endpoint for that instead

Is this enforced at API-level? if not it makes no sense to not support it, the error message is still misleading and sort of falsy. There are cases for adding components in non-interaction webhooks and this won't stop people from copying over the `content` field just to fool the lib

bsian03:
See bsian03/eris#13 and bsian03/eris#14 for these relevant PRs

DonovanDMC (dismissed):

curtisf (dismissed):
Checked your structures against the docs and ran locally, LGTM

JustCat80 (dismissed):

DonovanDMC (approved):

JustCat80 (approved):

curtisf (approved):
Matches docs (unless I'm absolutely blind)

bsian03:
My eyesight has worsened by double from 3 years ago so lol

conorwastakenwastaken (approved):

lib/Client.js
```diff
@@ -1242,8 +1276,8 @@ class Client extends EventEmitter {
                 };
             } else if(content.content !== undefined && typeof content.content !== "string") {
                 content.content = "" + content.content;
-            } else if(content.content === undefined && !content.embed && !content.embeds && content.flags === undefined) {
-                return Promise.reject(new Error("No content, embeds or flags"));
+            } else if(content.content === undefined && !content.embed && !content.embeds && !content.components && content.flags === undefined) {
```
DonovanDMC, line 1282:
```suggestion
            } else if(content.content === undefined && !content.embed && (!content.embeds || content.embeds.length === 0) && (!content.components || content.components.length === 0) && content.flags === undefined) {
```

This should be added to avoid issues when people provide only an empty array.

lib/Client.js
```diff
@@ -1491,8 +1542,8 @@ class Client extends EventEmitter {
     * @returns {Promise<Message>}
     */
     editWebhookMessage(webhookID, token, messageID, options) {
-        if(!options.content && !options.embeds && !options.file) {
-            return Promise.reject(new Error("No content, embeds or file"));
+        if(!options.content && !options.embeds && !options.components && !options.file) {
+            return Promise.reject(new Error("No content, embed, components, or file"));
```
DonovanDMC, line 1551:
```suggestion
        if(!options.content && (!options.embeds || options.embeds.length === 0) && (!options.components || options.components.length === 0) && !options.file) {
            return Promise.reject(new Error("No content, embeds, components, or file"));
```

DonovanDMC (changes requested):

DonovanDMC:
isn't `editMessage` also missing components?

lib/Client.js
abalabahaha: I think we can remove the check, there's an API error anyway and it's getting rather long

bsian03: The whole thing including for editMessage?

abalabahaha: Yeah, both create and edit. Nowadays this error is more likely to throw because of a new Discord feature than actually missing content.

DonovanDMC: in my opinion, the error this provides looks and sounds a lot better than what Discord says, which is just extremely generic
> Cannot send an empty message

bsian03: Bring that up with discord then :-)

bsian03:
Will remove the message checks in another PR if someone else doesn't get to it

PR #1190, opened 2021-04-28
Fix types, docs, lint

bsian03:


index.d.ts
```diff
@@ -923,11 +924,14 @@ declare namespace Eris {
     suppress?: boolean;
   }
   interface VoiceStreamCurrent {
+    buffer: unknown | null; // TODO check
```
bsian03: .

index.d.ts
```diff
@@ -923,11 +924,14 @@ declare namespace Eris {
...
+    bufferingTicks: number;
     options: VoiceResourceOptions;
     pausedTime?: number;
     pausedTimestamp?: number;
     playTime: number;
     startTime: number;
+    timeout: unknown | null; // TODO check
```
bsian03: .

index.d.ts
```diff
@@ -2524,14 +2525,46 @@ declare namespace Eris {
...
+    current?: VoiceStreamCurrent | null;
+    ended?: boolean;
+    endpoint: URL;
+    frameDuration: number;
+    frameSize: number;
+    heartbeatInterval: NodeJS.Timeout | null;
     id: string;
+    mode?: string;
+    modes?: string;
+    opus: { [s: string]: unknown }; // TODO
```
bsian03: .

index.d.ts
```diff
@@ -2524,14 +2525,46 @@ declare namespace Eris {
...
+    seret: Buffer;
+    shard: Shard | Record<string, never>;
+    shared: boolean;
+    speaking: boolean;
+    ssrc?: number;
+    ssrcUserMap: { [s: number]: string };
+    timestamp: number;
+    udpIP?: string;
+    udpPort?: number;
+    udpSocket: unknown | null; // TODO Dgram.Socket
```
bsian03: Install typings?

index.d.ts
```diff
@@ -2524,14 +2525,46 @@ declare namespace Eris {
...
     volume: number;
+    ws: BrowserWebSocket | unknown | null; // TODO No WS typings, might need npm i?
```
bsian03: Think this is `@types/ws`, install?

bsian03 (commented):
@abalabahaha these are the TODOs mentioned

index.d.ts
```diff
@@ -2524,14 +2525,46 @@ declare namespace Eris {
...
     playing: boolean;
     ready: boolean;
+    receiveStreamOpus?: VoiceDataStream | null;
+    receiveStreamPCM?: VoiceDataStream | null;
+    reconnecting: boolean;
+    samplingRate: number;
+    sendBuffer: Buffer;
+    sendNonce: Buffer;
+    sequence: number;
+    seret: Buffer;
```
```suggestion
    secret: Buffer;
    sendBuffer: Buffer;
    sendNonce: Buffer;
    sequence: number;
```

bsian03 (commented):
Can't spell secret

index.d.ts
abalabahaha: `dgram` is stdlib, `@types/node` is already a dep

index.d.ts
abalabahaha:
```suggestion
    buffer: Buffer | null;
```

lib/structures/Guild.js
```diff
@@ -623,13 +623,12 @@ class Guild extends Base {
     * @param {String} [userID="@me"] The user ID of the user to update
```
HeadTriXz, line 623:
```suggestion
     * @param {String} userID The user ID of the user to update
```

index.d.ts
abalabahaha:
```suggestion
    timeout: NodeJS.Timeout | null;
```

index.d.ts
abalabahaha: `userID`:  `OpusScript` from `opusscript` | `OpusEncoder` from `@discordjs/opus`

Both are optional packages so not sure if this can be typed? I'm perfectly fine with leaving this untyped/`unknown` anyway

index.d.ts
abalabahaha: `ws`: `WebSocket` from `@types.ws` | `BrowserWebSocket` | `null`

HeadTriXz (changes requested):
Nvm

HeadTriXz (changes requested):
Change `editVanity(code: string): Promise<GuildVanity>;` to `editVanity(code: string | null): Promise<GuildVanity>;` on line 1957 in index.d.ts

index.d.ts
bsian03: Leaving as unknown seems to be the best option without requiring a hard install then

index.d.ts
bsian03: Ah didn't know that

lib/structures/Guild.js
bsian03, line 623: My change is valid

abalabahaha (approved):

PR #1193, opened 2021-05-01
Allow configuring RequestHandler to use HTTP instead of HTTPS

molenzwiebel:
My use case was to tunnel all Eris traffic through a central rate limiter (Twilight's [http-proxy](https://github.com/twilight-rs/http-proxy)), which uses HTTP instead of HTTPS. Given that we can already configure the domain and base URL, being able to configure whether to use HTTPS or HTTP seemed logical.

I've named it `https` and enabled it by default, to ensure concious opt-out by doing `https: false`. A name like `useHttps` might be more appropriate.

Could also consider storing the request method (`https` or `http` module) as an instance variable on RequestHandler instead of doing the ternary on each request, but I figured this'd be fine.

JSDoc and TypeScript typings should be good.


tbnritzdoge:
Should we not just take the protocol from the rest URL provided?

molenzwiebel:
The domain doesn't have a protocol prefixed. Using it would probably be neater, but it would break backwards compat.

lib/Client.js
```diff
@@ -109,11 +109,12 @@ class Client extends EventEmitter {
     * @arg {Function} [options.reconnectDelay] A function which returns how long the bot should wait until reconnecting to Discord.
     * @arg {Number} [options.requestTimeout=15000] A number of milliseconds before requests are considered timed out. This option will stop affecting REST in a future release; that behavior is [DEPRECATED] and replaced by `options.rest.requestTimeout`
     * @arg {Object} [options.rest] Options for the REST request handler
-    * @arg {Object} [options.rest.agent] A HTTPS Agent used to proxy requests
+    * @arg {Object} [options.rest.agent] A HTTPS Agent (if https: true, default) or an HTTP agent (if https: false) used to proxy requests
```
```suggestion
    * @arg {Object} [options.rest.agent] An HTTP Agent used to proxy requests
```
this would probably be better said as `HTTP Agent`, noting that difference doesn't seem that important (and if you want to argue for it, `HTTP/S` could be used instead of those distinctions)

```suggestion
    * @arg {Object} [options.rest.agent] A HTTP(S) Agent used to proxy requests
```

flav-code:
tried but not working and no error

bsian03:
Request from @HcgRandon - add option to override port for proxy related stuff. Not urgent

bsian03:
@HcgRandon I added port functionality into this PR, is this essentially what you're currently using to connect to nirn?

PR #1199, opened 2021-05-07
Change hyperlink text

bsian03:


Vetlix (commented):
Maybe specify that it's a discord server?

coolcalcacol:
Maybe specify that the discord server is NOT a cult, instead only providing support for the library? May also be good to identify that it is the support server for the Eris discord library, some people may get confused and think it's the goddess being referred to?

curtisf (approved):

PR #1200, opened 2021-05-07
Document initial client presence

bsian03:
Happy #1200


abalabahaha:
`options.presence` isn't being used anywhere. `Client#editStatus()` already does what this intends to do.

bsian03:
Also forgot we're not on gateway v8

PR #1203, opened 2021-05-16
Fix types and docs

HeadTriXz:


index.d.ts
```diff
@@ -1668,7 +1682,7 @@ declare namespace Eris {
     executeWebhook(webhookID: string, token: string, options: WebhookPayload): Promise<void>;
     followChannel(channelID: string, webhookChannelID: string): Promise<ChannelFollow>;
     getBotGateway(): Promise<{ session_start_limit: { max_concurrency: number; remaining: number; reset_after: number; total: number }; shards: number; url: string }>;
-    getChannel(channelID: string): AnyChannel;
+    getChannel<T extends Channel = AnyChannel>(channelID: string): T;
```
bsian03, line 1720: We discussed this a couple times, concluded that type assertions are more appropriate

Dragory:
There are a few small discrepancies in the types for `TextChannel`, `GuildChannel`, and `Channel` (vs. the actual JS code) that could be included in this PR:

* `TextChannel` constructor should have the signature `(data: BaseData, client: Client, messageLimit: number)` (second parameter changed from `Guild` to `Client`).
* `GuildChannel` constructor should have the signature `(data: BaseData, client: Client)` (second parameter changed from `Guild` to `Client`).
* `Channel` constructor should have the signature `(data: BaseData, client: Client)` (added second parameter)

PR #1204, opened 2021-05-18
add certified moderator flag

untuned:
https://github.com/Discord-Datamining/Discord-Datamining/commit/6258cfefc7dde95d7c340397e9704c30da9a6280#commitcomment-50961513

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1266,6 +1266,7 @@ declare namespace Eris {
       BUG_HUNTER_LEVEL_2: 16384;
       VERIFIED_BOT: 65536;
       VERIFIED_BOT_DEVELOPER: 131072;
+      CERTIFIED_MODERATOR: 262144;
     };
     VoiceOPCodes: {
       IDENTIFY: 0;
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -252,7 +252,8 @@ module.exports.UserFlags = {
     SYSTEM:                 1 << 12,
     BUG_HUNTER_LEVEL_2:     1 << 14,
     VERIFIED_BOT:           1 << 16,
-    VERIFIED_BOT_DEVELOPER: 1 << 17
+    VERIFIED_BOT_DEVELOPER: 1 << 17,
+    CERTIFIED_MODERATOR:    1 << 18
 };
 
 


bsian03:
Is this actually exposed in v6/7?

DonovanDMC:
I don't think they've given it out yet, but I don't see why it wouldn't be
it's not near high enough to cause any issues, and new flags appearing out of thin air (usually) won't break anything

untuned:
> I don't think they've given it out yet, but I don't see why it wouldn't be
> it's not near high enough to cause any issues, and new flags appearing out of thin air (usually) won't break anything

It was recently added to canary, and supposedly it's for all members in the Discord Moderator Discord.

DonovanDMC:
While we're on flags, can we talk about the naming of `VERIFIED_BOT_DEVELOPER`? That name is outdated, as [the docs](https://discord.com/developers/docs/resources/user#user-object-user-flags) call it `EARLY_VERIFIED_BOT_DEVELOPER`, because only early devs with the badge have the flag, so its old naming is a bit misleading

untuned:
> While we're on flags, can we talk about the naming of `VERIFIED_BOT_DEVELOPER`? That name is outdated, as [the docs](https://discord.com/developers/docs/resources/user#user-object-user-flags) call it `EARLY_VERIFIED_BOT_DEVELOPER`, because only early devs with the badge have the flag, so its old naming is a bit misleading

Is that what the string is named?

coolcalcacol:
> > While we're on flags, can we talk about the naming of `VERIFIED_BOT_DEVELOPER`? That name is outdated, as [the docs](https://discord.com/developers/docs/resources/user#user-object-user-flags) call it `EARLY_VERIFIED_BOT_DEVELOPER`, because only early devs with the badge have the flag, so its old naming is a bit misleading
> 
> Is that what the string is named?

Yup, at least according to this: https://discord.com/developers/docs/resources/user#user-object-user-flags

HeadTriXz:
Also, `DISCORD_PARTNER` is no longer used either. It got replaced with `PARTNERED_SERVER_OWNER`.

DonovanDMC:
> Also, `DISCORD_PARTNER` is no longer used either. It got replaced with `PARTNERED_SERVER_OWNER`.

I didn't even notice that one

untuned:
With the [official announcement](https://blog.discord.com/announcing-our-latest-profile-badge-the-certified-discord-moderator-908182193af8) on their blog, I believe this should be merged. It has also been added to the [user flags list](https://discord.com/developers/docs/resources/user#user-object-user-flags) in the Discord developer documentation, in commit https://github.com/discord/discord-api-docs/commit/55cd3d71082e9dc31cbd007a6e74b1cd6a82beeb.

listed on a user profile

bsian03 (approved):
Name changes can be in its own PR (it's also in the v9 PR but if someone wants to do it before then, feel free)

PR #1206, opened 2021-05-20
off function on client now declared for TS

AlekEagle:
typedef for off function on client declared now.


HeadTriXz (changes requested):
You will need to update the lint rules in `.eslintrc.ts.yml` and you are missing other classes extending EventEmitter.

AlekEagle:
I think I fixed it

PR #1216, opened 2021-06-05
Discord API v9

bsian03:
### Note
Previous versions of this description contained commit hashes, but due to high maintenance, I have removed them.

## Breaking
- Guild member joinedAt is nullable
- Guild#region has been removed, replaced by GuildChannel#rtcRegion

## Additions
- Add thread channel type constants
- Add HAS_THREAD message flag
- Add thread related message types
- Add thread channel permissions
- Add thread channel URLs
- NewsThreadChannel, PrivateThreadChannel, PublicThreadChannel classes extending ThreadChannel
- Client methods
- ThreadMember class
- `threads` property in Guild
- Thread gateway events
- Threads support in Audit Logs
- Support adding/editing attachments on editMessage
- Stage Instance class and relevant methods
- Stage Instance caching
- Stage Instance gateway events
- Stage Instance support in Audit Logs
- Message#applicationID
- TextChannel#defaultAutoArchiveDuration
- Invite#stageInstance

## Changes
- Gateway/REST versions 8 => 9
- Minor documentation updates
- Client#getChannel checks for stage channels as well
- Channel.from now supports Thread channels
- Option to send webhook message to a thread channel (querystring logic has been changed)
- Channel#permissionsOf case for thread channel (incorrect commit message)
- Message#guildID/member/webhookID case for thread channel
- Other misc

Note for self: UTD docs discord/discord-api-docs#3672 tree 0d954e67b4b32703bf895d8d9c1d8853184652d1


curtisf:
Needs support for \<ThreadChannel\>.createMessage and \<ThreadChannel\>.permissionsOf. Newly created public/private channels still throw an unknown channel type warning because ChannelTypes.GUILD_PUBLIC/PRIVATE_THREAD isn't handled in Channel.js. Might be worth it to add \<Message\>.channel support for thread messages too, since current Message.js code doesn't give a channel object but instead just the ID in an object.

curtisf:
[This commit](https://github.com/curtisf/eris/commit/3e3ea256e6788f1f4303723a009457afb1e2d71a) is my first naive thought to fix these problems ^^

lib/Client.js
```diff
@@ -1588,6 +1722,9 @@ class Client extends EventEmitter {
         if(this.channelGuildMap[channelID] && this.guilds.get(this.channelGuildMap[channelID])) {
             return this.guilds.get(this.channelGuildMap[channelID]).channels.get(channelID);
         }
+        if(this.threadGuildMap[channelID] && this.guilds.get(this.threadGuildMap[channelID])) {
+            return this.guilds.get(this.threadGuildMap[channelID]).channels.get(channelID);
```
curtisf: Threads are not included in `<Guild>.channels`, this needs to be `<Guild>.threads.get(channelID);`

bsian03:
> Needs support for... <ThreadChannel>.permissionsOf

ThreadChannel already extends GuildChannel so this is covered already

Other than that, the rest of it should be covered

lib/gateway/Shard.js
```diff
@@ -2070,6 +2053,204 @@ class Shard extends EventEmitter {
...
+                if(!guild) {
+                    this.emit("debug", `Missing guild ${packet.d.guild_id} in STAGE_INSTANCE_CREATE`);
+                    break;
+                }
+                /**
+                * Fired when a stage instance is created
+                * @event Client#stageInstanceCreate
+                * @prop {StageInstance} stageInstance The stage instance
+                */
+                this.emit("stageInstanceCreate", guild.stageInstances.add(packet.d));
```
ghost: StageInstance requires a client property which is not provided when adding a new stage instance to the guild. Which causes an error: `TypeError: Cannot read property 'getChannel' of undefined at new StageInstance` [here](https://github.com/bsian03/eris/blob/d3043f36708fd48ef512faf473ad8768c45d2efb/lib/structures/StageInstance.js#L18)

bsian03: sad

lib/structures/StageInstance.js
```diff
@@ -0,0 +1,55 @@
...
+* @prop {Boolean} discoverableDisabled Whether or not stage discovery is disabled
+* @prop {Guild} guild The guild of the associated stage channel
+* @prop {String} id The ID of the stage instance
+* @prop {Number} privacyLevel The privacy level of the stage instance. 1 is public, 2 is guild only
+* @prop {String} topic The stage instance topic
+*/
+class StageInstance extends Base {
+    constructor(data, client) {
+        super(data.id);
+        this.client = client;
```
curtisf: This should be this._client (with updated methods) so your console logs don't get drowned by a printed Client.

curtisf (changes requested):

lib/Client.js
```diff
@@ -1579,6 +1678,34 @@ class Client extends EventEmitter {
         return this.requestHandler.request("POST", Endpoints.CHANNEL_FOLLOW(channelID), true, {webhook_channel_id: webhookChannelID});
     }
 
+    /**
+    * Get all active threads in a channel
+    * @arg {String} channelID The ID of the channel
+    * @returns {Promise<Array<NewsThreadChannel | PrivateThreadChannel | PublicThreadChannel>>}
+    */
+    getActiveThreads(channelID) {
+        return this.requestHandler.request("GET", Endpoints.THREADS_ACTIVE(channelID), true).then((channels) => channels.map((c) => Channel.from(c, this)));
```
Nico105:
```suggestion
        return this.requestHandler.request("GET", Endpoints.THREADS_ACTIVE(channelID), true).then((response) => response.threads.map((c) => Channel.from(c, this)));
```
the request returns an object, like for `getArchivedThreads`

Nico105:
```js
(node:20260) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'get' of undefined
    at PublicThreadChannel.permissionsOf (C:\Privat\Coding\Discord\Bots\eris-giveaways-bot\node_modules\eris\lib\structures\GuildChannel.js:130:51)
```
I couldn't really find the process for where `permissionOverwrites` gets set _besides update(data)_ so I will just leave this here.

DonovanDMC:
Pretty sure that's because everything before threads had permission overwrites, but threads don't
They have no concept of permissions, so the field is absent

The [docs](https://discord.com/developers/docs/resources/channel#channel-object-example-thread-channel) show `permission_overwrites` being completely absent on thread channels

Nico105:
> Pretty sure that's because everything before threads had permission overwrites, but threads don't
> They have no concept of permissions, so the field is absent
> 
> The [docs](https://discord.com/developers/docs/resources/channel#channel-object-example-thread-channel) show `permission_overwrites` being completely absent on thread channels

So should then just the `let permission = this.guild.permissionsOf(member).allow` get returned?

DonovanDMC:
the permissions of thread channels is basically the permissions of the channel the thread was created on, they have the `parent_id` property iirc

might be good to look into either setting `permissionOverwrites` for thread channels to empty, or to mirror its parent channel

andromika:
Not entirely sure if it is the appropriate way to go by but [this PR](https://github.com/bsian03/eris/pull/17) just assigns the parent channel `permissionOverwrites` whenever the channel types are 10 and 11.

So far this has seamlessly emulated the behaviour of a regular channel regarding permissions.

Runi-c:
There seems to be some typing issues introduced by `ThreadChannel`s derived from them not implementing `GuildTextable`, which it seems like they really should be since they are in fact textable guild channels. Probably means moving interface members from `GuildTextable` to `TextChannel`.

- `Message<ThreadChannel>` doesn't get `msg.member` or `msg.guildID`
- `Message<TextableChannel>` (the default) can't be narrowed to `Message<GuildTextableChannel>` because `ThreadChannel` isn't assignable to `GuildTextableChannel`

bsian03:
I'll get to these issues and relevant PRs on my fork within the next couple days. Other than that the PR is working and is currently being used by Dyno on production and working with no issues.

lib/structures/Message.js
```diff
@@ -220,12 +220,18 @@ class Message extends Base {
                 data.content = "This server has failed Discovery activity requirements for 3 weeks in a row. If this server fails for 1 more week, it will be removed from Discovery.";
                 break;
             }
+            case MessageTypes.THREAD_CREATED: {
+                break;
```
Runi-c, line 228:
```suggestion
                data.content = `${this.author.mention} started a thread: ${data.content}. See all threads.`;
                break;
```

bsian03: @abalabahaha leaving this up to you as I prefer not replacing the message content with a fixed localised string for a message that has no content in the first place

index.d.ts
```diff
@@ -2574,12 +2720,18 @@ declare namespace Eris {
     addMessageReaction(messageID: string, reaction: string, userID: string): Promise<void>;
     createInvite(options?: CreateInviteOptions, reason?: string): Promise<Invite<"withMetadata", TextChannel>>;
     createMessage(content: MessageContent, file?: MessageFile | MessageFile[]): Promise<Message<TextChannel>>;
+    createThreadWithoutMessage(options: CreateThreadWithoutMessageOptions): Promise<PrivateThreadChannel>;
+    createThreadWithMessage(messageID: string, options: CreateThreadOptions): Promise<PublicThreadChannel>;
     createWebhook(options: { name: string; avatar?: string | null}, reason?: string): Promise<Webhook>;
```
conorwastakenwastaken:
```suggestion
    createWebhook(options: { name: string; avatar?: string | null }, reason?: string): Promise<Webhook>;
```

conorwastakenwastaken (changes requested):

lib/structures/Invite.js
```diff
@@ -48,6 +49,20 @@ class Invite extends Base {
         this._createdAt = data.created_at !== undefined ? data.created_at : null;
         this.presenceCount = data.approximate_presence_count !== undefined ? data.approximate_presence_count : null;
         this.memberCount = data.approximate_member_count !== undefined ? data.approximate_member_count : null;
+        if(data.stage_instance !== undefined) {
+            data.stage_instance.members = data.stage_instance.members.map((m) => {
+                m.id = m.user.id;
+                return m;
+            });
+            this.stageInstance = {
+                members: this.guild.members.update(data.stage_instance.members, this.guild),
```
conorwastakenwastaken: This will throw [an error](https://github.com/abalabahaha/eris/blob/dev/lib/util/Collection.js#L31) since you're trying to send an array
```suggestion
                members: data.stage_instance.members,
```

lib/structures/Invite.js
```diff
@@ -48,6 +49,20 @@ class Invite extends Base {
         this._createdAt = data.created_at !== undefined ? data.created_at : null;
         this.presenceCount = data.approximate_presence_count !== undefined ? data.approximate_presence_count : null;
         this.memberCount = data.approximate_member_count !== undefined ? data.approximate_member_count : null;
+        if(data.stage_instance !== undefined) {
+            data.stage_instance.members = data.stage_instance.members.map((m) => {
+                m.id = m.user.id;
+                return m;
```
conorwastakenwastaken, line 55: If the bot isn't in the guild that the invite originated from, the `update` function would run into another error when trying to add the user, because `guild.shard` doesn't exist. In the event of the bot not being inside of a guild, I made it just return the member.
```suggestion
                return client.guilds.has(this.guild.id) ? this.guild.members.update(m, this.guild) : m;
```

conorwastakenwastaken (changes requested):

bsian03:
~~@Nico105 CNR permissions error~~ nvm

lib/Client.js
bsian03: This isn't the correct suggestion. We don't want to just return the channels. I'll push a commit soon

lib/structures/Invite.js
bsian03: Wrong suggestion, see just pushed commit

lib/structures/Invite.js
bsian03: That shouldn't matter though should it? Since Discord sends the guild structure if there is an associated stage instance (unless I'm mistaken)

lib/structures/GuildChannel.js
```diff
@@ -13,7 +14,7 @@ const PermissionOverwrite = require("./PermissionOverwrite");
 * @prop {String} id The ID of the channel
 * @prop {String} name The name of the channel
 * @prop {Boolean} nsfw Whether the channel is an NSFW channel or not
-* @prop {String?} parentID The ID of the category this channel belongs to
+* @prop {String?} parentID The ID of the category this channel belongs to or the message ID where the thread originated from (thread channels only)
```
Nico105:
```suggestion
* @prop {String?} parentID The ID of the category this channel belongs to or the channel ID where the thread originated from (thread channels only)
```
right?

lib/structures/GuildChannel.js
```diff
@@ -71,17 +72,21 @@ class GuildChannel extends Channel {
...
     * @arg {Number} [options.videoQualityMode] The camera video quality mode of the channel (guild voice channels only). `1` is auto, `2` is 720p
-    * @arg {Number} [options.rateLimitPerUser] The time in seconds a user has to wait before sending another message (does not affect bots or users with manageMessages/manageChannel permissions) (guild text channels only)
+    * @arg {Number} [options.rateLimitPerUser] The time in seconds a user has to wait before sending another message (does not affect bots or users with manageMessages/manageChannel permissions) (guild text and thread channels only)
     * @arg {String?} [options.rtcRegion] The RTC region ID of the channel (automatic if `null`) (guild voice channels only)
     * @arg {Boolean} [options.nsfw] The nsfw status of the channel
-    * @arg {Number?} [options.parentID] The ID of the parent channel category for this channel (guild text/voice channels only)
+    * @arg {Number?} [options.parentID] The ID of the parent channel category for this channel (guild text/voice channels only) or the message ID where the thread originated from (thread channels only)
```
Nico105:
```suggestion
    * @arg {Number?} [options.parentID] The ID of the parent channel category for this channel (guild text/voice channels only) or the channel ID where the thread originated from (thread channels only)
```

Nico105 (changes requested):

lib/structures/GuildChannel.js
bsian03: No

Nico105: https://discord.com/developers/docs/resources/channel#channel-object
Then why does it say it here?
and in code you also use `.channels.get(.parentID)` ?

DonovanDMC: [docs](https://discord.com/developers/docs/resources/channel#channel-object-channel-structure) say parent_id is the id of the channel the thread was made on
>  for threads: id of the text channel this thread was created

bsian03: Ah sorry, mixing up that with the channel ID

lib/gateway/Shard.js
```diff
@@ -2069,6 +2069,204 @@ class Shard extends EventEmitter {
...
+                * @prop {NewsThreadChannel | PrivateThreadChannel | PublicThreadChannel} channel The channel
+                */
+                this.emit("threadCreate", channel);
+                break;
+            }
+            case "THREAD_UPDATE": {
+                const channel = this.client.getChannel(packet.d.id);
+                if(!channel) {
+                    break;
+                }
```
Runi-c: Archived threads on startup that are then un-archived aren't announced with a THREAD_CREATE, only a THREAD_UPDATE - should probably cache those here

lib/structures/Invite.js
conorwastakenwastaken: Yeah but said guild structure doesn't include a `shard` property, so it results in an error. This is just what I've been experiencing tbh.

lib/structures/Invite.js
conorwastakenwastaken: Well there's a [reason](https://github.com/abalabahaha/eris/pull/1216#discussion_r680408886) I didn't do it the same way as you in your pushed commit lol. Are you not able to repro that error referenced in that comment?

lib/structures/Invite.js
bsian03: Ah I see what you mean now

conorwastakenwastaken: Yeah, that's why I tried to make it so it doesn't try and update the members if the guild isn't cached, but maybe there's a better way of doing it

lib/gateway/Shard.js
```diff
@@ -2069,6 +2069,220 @@ class Shard extends EventEmitter {
...
+                channel.update(packet.d);
+                let removedMembers;
+                let addedMembers;
+                if(packet.d.removed_member_ids) {
+                    removedMembers = packet.d.removed_member_ids.map((id) => channel.members.remove({id}) || {id});
+                }
+                if(packet.d.added_members) {
+                    addedMembers = packet.d.added_members.map((m) => {
+                        m.presence.id = m.presence.user.id;
```
ghost: `m.presence` can be null sometimes.

bsian03: Discord at it again with the poor documentation

lib/Constants.js
```diff
@@ -67,7 +67,12 @@ const Permissions = {
     manageEmojisAndStickers: 1n << 30n, manageEmojis: 1n << 30n, // [DEPRECATED]
     useApplicationCommands:  1n << 31n, useSlashCommands: 1n << 31n, // [DEPRECATED]
     voiceRequestToSpeak:     1n << 32n,
-    useExternalStickers:     1n << 37n
+
+    manageThreads:           1n << 34n,
+    createPublicThreads:     1n << 35n,
+    createPrivateThreads:    1n << 36n,
+    useExternalStickers:     1n << 37n,
+    sendMessagesinThreads:   1n << 38n
```
DonovanDMC:
```suggestion
    sendMessagesInThreads:   1n << 38n
```
noob

index.d.ts
```diff
@@ -1028,7 +1078,41 @@ declare namespace Eris {
...
+    threads: T[];
+  }
+  interface PrivateThreadMetadata extends ThreadMetadata {
+    invitable: boolean;
+  }
+  interface ThreadMetadata {
+    archiveTimestamp: number;
+    archived: boolean;
+    autoArchiveDuration: AutoArchiveDuration;
+    locked?: boolean;
```
DonovanDMC: Unless I'm looking in the wrong place, the [docs](https://discord.com/developers/docs/resources/channel#thread-metadata-object) don't list `locked` as optional

bsian03: Missed that when adding invitable (same doc commit), thanks

lib/gateway/Shard.js
```diff
@@ -2071,6 +2071,222 @@ class Shard extends EventEmitter {
...
+                    });
+                }
+                /**
+                * Fired when anyone is added or removed from a thread. If the `guildMembers` intent is not specified, this will only apply for the current user
+                * @event Client#threadMembersUpdate
+                * @prop {NewsThreadChannel | PrivateThreadChannel | PublicThreadChannel} channel The thread channel
+                * @prop {Array<ThreadMember | Object>} removedMembers An array of members that were removed from the thread channel. If a member is uncached, it will be an object with an `id` key. No other property is guaranteed
+                * @prop {Array<ThreadMember>} addedMembers An array of members that were added to the thread channel
+                */
+                this.emit("threadMembersUpdate", channel, removedMembers, addedMembers);
```
DonovanDMC: Either this or the types need to be changed, types say both `addedMembers` and `removedMembers` are always present, and the current behavior makes them able to be undefined.

DonovanDMC (changes requested):

index.d.ts
```diff
@@ -1686,6 +1791,9 @@ declare namespace Eris {
     createGuildTemplate(guildID: string, name: string, description?: string | null): Promise<GuildTemplate>;
     createMessage(channelID: string, content: MessageContent, file?: MessageFile | MessageFile[]): Promise<Message>;
     createRole(guildID: string, options?: RoleOptions | Role, reason?: string): Promise<Role>;
+    createStageInstance(channelID: string, options: StageInstanceOptions): Promise<StageInstance>;
+    createThreadWithMessage(channelID: string, options: CreateThreadOptions): Promise<NewsThreadChannel | PublicThreadChannel>;
```
cillynder:
```suggestion
    createThreadWithMessage(channelID: string, messageID: string, options: CreateThreadOptions): Promise<NewsThreadChannel | PublicThreadChannel>;
```
There should be a `messageID` parameter here

cillynder (changes requested):

index.d.ts
bsian03: Is this the only location where that error appears? I'm unable to check at the mo

lib/gateway/Shard.js
bsian03: Will reconsider how to structure this event

lib/structures/Invite.js
bsian03: Should be fixed

bsian03:
Going to create another PR for this, there won't be any changes.
Hacktoberfest & Comment cleanup

PR #1217, opened 2021-06-05
Fix possibly undefined member in voiceStateUpdate

bsian03:
This does mean there is a breaking change, see diff for info

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -739,7 +739,7 @@ class Shard extends EventEmitter {
                                 mute: packet.d.mute,
                                 selfDeaf: packet.d.self_deaf,
                                 selfMute: packet.d.self_mute,
-                                selfStream: packet.d.selfStream || false,
+                                selfStream: packet.d.self_stream || false,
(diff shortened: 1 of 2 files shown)


lib/gateway/Shard.js
```diff
@@ -795,8 +809,8 @@ class Shard extends EventEmitter {
                     /**
                     * Fired when a guild member's voice state changes
                     * @event Client#voiceStateUpdate
-                    * @prop {Member} member The member
-                    * @prop {Object} oldState The old voice state
+                    * @prop {Member | Object} member The member. If the member is not cached and Discord doesn't send a member payload, this will be an object with `id` and `voiceState` keys. No other property is guaranteed
+                    * @prop {Object?} oldState The old voice state If the member. If the above caveat applies, this will be null
```
eritbh: Am I interpreting this right?

```suggestion
                    * @prop {Object?} oldState The old voice state of the member. If the above caveat applies, this will be null
```

curtisf (approved):
Tested locally, manually deleted member + packet.d.member and this checks out

PR #1218, opened 2021-06-05
Added ClientEvents constant enum to index.d.ts

hellodhruvin:
This won't necessarily find use in the lib (since this is an enum and js doesn't have enums), but will surely be useful to people using typescript.


bsian03:
We don't use enums in typings at all just fyi

hellodhruvin:
Alright then, I guess i should close it.

PR #1219, opened 2021-06-05
Support per-guild avatars

bsian03:
Adds support for guild specific avatars 

## Notes:
- Possibly breaking change: Member#avatar will now return the avatar hash of the guild member or null if there isn't one, instead of getting the relevant hash from the User class. 
- This feature is still in private beta, only some users will be able to see and use it.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -455,10 +455,11 @@ declare namespace Eris {
     videoQualityMode: VideoQualityMode;
   }
   interface OldMember {
-    roles: string[];
+    avatar: string | null;
     nick: string | null;
-    premiumSince: number;
     pending?: boolean;
+    premiumSince: number;
+    roles: string[];
   }
   interface OldMessage {
     attachments: Attachment[];
@@ -789,7 +790,7 @@ declare namespace Eris {
     channelID?: string | null;
     deaf?: boolean;
     mute?: boolean;
-    nick?: string;
+    nick?: string | null;
     roles?: string[];
   }
   interface MemberPartial {
@@ -1634,6 +1635,7 @@ declare namespace Eris {
     editGuildWelcomeScreen(guildID: string, options: WelcomeScreenOptions): Promise<WelcomeScreen>;
     editGuildWidget(guildID: string, options: Widget): Promise<Widget>;
     editMessage(channelID: string, messageID: string, content: MessageContent): Promise<Message>;
+    /** @deprecated */
     editNickname(guildID: string, nick: string, reason?: string): Promise<void>;
     editRole(guildID: string, roleID: string, options: RoleOptions, reason?: string): Promise<Role>; // TODO not all options are available?
     editRolePosition(guildID: string, roleID: string, position: number): Promise<void>;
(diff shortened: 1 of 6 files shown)


lib/Client.js
```diff
@@ -1245,7 +1245,7 @@ class Client extends EventEmitter {
     }
 
     /**
-    * Edit the bot's nickname in a guild
+    * [DEPRECATED] Edit the bot's nickname in a guild
```
Dramex, line 1262: is it? i don't think it's DEPRECATED as per discord docs
https://discord.com/developers/docs/resources/guild#modify-guild-member

DonovanDMC: Pretty sure he's just deprecating `Client#editNickname` (in favor of `Member#edit`),  not editing nicknames as a whole

bsian03: That's because the PR for it hasn't been merged yet. Same reason why the documentation for per guild avatars also doesn't exist.

Dramex: @DonovanDMC then i'll recommend mentioning that to avoid confusing users.

bsian03: It's fine as it is

DonovanDMC: > @DonovanDMC then i'll recommend mentioning that to avoid confusing users.

I'd argue most people won't assume nickname editing as a whole isn't getting deprecated (which is obvious by the code, since this uses a different endpoint than `Member#edit`)

You also linked the wrong docs section. The one this references is one below that,
[https://discord.com/developers/docs/resources/guild#modify-current-user-nick](https://discord.com/developers/docs/resources/guild#modify-current-user-nick)

index.d.ts
```diff
@@ -455,10 +455,11 @@ declare namespace Eris {
     videoQualityMode: VideoQualityMode;
   }
   interface OldMember {
-    roles: string[];
+    avatar: string | null;
     nick: string | null;
-    premiumSince: number;
     pending?: boolean;
+    premiumSince: number;
```
curtisf, line 465: Premium since can be null for old members

lib/gateway/Shard.js
```diff
@@ -1137,11 +1138,12 @@ class Shard extends EventEmitter {
                 }
                 member = guild.members.update(packet.d, guild);
                 /**
-                * Fired when a member's roles or nickname are updated or they start boosting a server
+                * Fired when a member's guild avatar, roles or nickname are updated or they start boosting a server
```
curtisf, line 1148: Missing `when a member passes the rules gate for a server`?

curtisf (commented):
I like the idea to get rid of both of the extraneous editNickname methods in Client and Guild.

index.d.ts
bsian03: That can go in its own PR, that addition isn't really an addition, it's a relocation

lib/gateway/Shard.js
bsian03: Can also go in its own PR

lib/Client.js
abalabahaha: For future reference, https://github.com/discord/discord-api-docs/pull/3081/files#diff-6a43fc10e10de61f31b90ab22421ef0612d0ea75e3cd53891c6393a6fb84872dR772

abalabahaha (approved):

PR #1221, opened 2021-06-06
Wrong property for client

notsapinho:


Zorotic:
The property is prefixed with an underscore for a reason, therefore it isn't "wrong"

notsapinho:
yeah, but if u take a look at `PrivateChannel.js` and `GroupChannel.js`

none of the files has the `_client` property

`PrivateChannel.js`
```js
class Channel extends Base {
            constructor(data, client) {
                super(data.id);
                this.type = data.type;
                this.client = client;
```

`GroupChannel.js`
```js
class GroupChannel extends PrivateChannel {
    constructor(data, client) {
        super(data, client);
        this.recipients = new Collection(User);
        data.recipients.forEach((recipient) => {
            this.recipients.add(client.options.restMode ? new User(recipient, client) : client.users.add(recipient, client));
        });
        this.update(data);
    }
```

Zorotic:
its _client cause _client is a private property.

LJNeon:
I think it would be better to open a PR to remove user account code entirely, but in the end that's up to abal. Not much point in fixing code that will never be run.

notsapinho:
> its _client cause _client is a private property.

I was testing with my selfbot, everytime u add the bot to a group, a error shows up saying that _client is undefined, u can test by yourself

Zorotic:
> > its _client cause _client is a private property.
> 
> I was testing with my selfbot, everytime u add the bot to a group, a error shows up saying that _client is undefined, u can test by yourself

Selfbots are against ToS so I will not be testing this. With that the only way this code you changed in the PR will be accessed is when using a selfbot/userbot

bsian03:
Ignoring the fact that you're using a selfbot, which is against discord TOS and which Eris doesn't support as a result, renaming properties will cause breaking changes in other locations, which is why they are discouraged where not necessary, even if they are only affecting certain types of accounts. Please also refer to the contributing guidelines, it specifically states to discuss any changes you want to make. And that PRs should be to the Dev branch.

bsian03:
Also ignoring the fact that you're using a selfbot, which is against discord TOS and which Eris doesn't support as a result,

> a error shows up saying that _client is undefined

Changing something just to remove an underscore literally makes no difference

PR #1224, opened 2021-06-09
Allow passing a plain object to permissionsOf

bsian03:
Added in case we're dealing with plain objects, avoids throwing an error relating to member being uncached if we pass an object.

### DIFF
--- a/lib/structures/Guild.js
+++ b/lib/structures/Guild.js
@@ -898,7 +898,7 @@ class Guild extends Base {
     * @returns {Permission}
     */
     permissionsOf(memberID) {
-        const member = typeof memberID === "string" ? this.guild.members.get(memberID) : memberID;
+        const member = typeof memberID === "string" ? this.members.get(memberID) : memberID;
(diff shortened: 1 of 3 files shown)


lib/structures/Guild.js
```diff
@@ -894,11 +894,11 @@ class Guild extends Base {
 
     /**
     * Get the guild permissions of a member
-    * @arg {String | Member} memberID The ID of the member or a Member instance
+    * @arg {String | Member | Object} memberID The ID of the member or a Member object
     * @returns {Permission}
     */
     permissionsOf(memberID) {
-        const member = memberID instanceof Member ? memberID : this.members.get(memberID);
+        const member = typeof memberID === "string" ? this.guild.members.get(memberID) : memberID;
```
curtisf: this.guild causes an error because you're already in a Guild instance, need to change to .members only

curtisf (changes requested):

curtisf (approved):
Concerns addressed and works locally

PR #1227, opened 2021-06-25
Standardized parsing errors

LJNeon:
As far as I know, the main cases where this code throws are Discord outages and internet problems. In both cases (and I presume in any case really), the actual syntax error is irrelevant. Adding a standard prefix to the error message is already done for zlib errors a few lines above, and makes it easier to determine the actual error.


bsian03 (commented):
The return behaviour L2186/2190 is inconsistent, doesn't matter but would be nice if both do the "same thing"

LJNeon:
The goal was to make L2186 consistent with the zlib parsing error above. I would say making those "both do the same thing" as you put it makes more sense to me considering they're very similar errors, compared to a different return statement at the bottom.

abalabahaha:
Did we figure out the performance impact of this?

LJNeon:
> Did we figure out the performance impact of this?

Ran a quick test and it seems like the impact is around ~30% for just the parsing code. Doesn't seem to make a noticeable difference to the overall function's performance.

What I did for the test was add some `process.hrtime.bigint()` calls to the function and calculate averages after bringing a small test bot online for a few minutes and giving it events. Probably not perfect but I figure that's a real-world-ish test.

PR #1228, opened 2021-06-26
Remove extra whitespace

J-Human:
Unless this was intentionally placed, this removes the extra whitespace in the ESLint config file.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -88,4 +88,3 @@ rules:
         custom-inspect-method:
         - name: "[util.inspect.custom]"
           type: method
-


HeadTriXz:
The extra whitespace has a reason to be there.

J-Human:
> The extra whitespace has a reason to be there.

If that is so, my bad. But what exactly is that "reason" supposed to be then?

DonovanDMC:
Files typically end in a newline for compatibility, and some other things I can't think of

I know all of my files end the same way, it's automatic in most cases

I Googled it, it's for knowing when the file ends
> The empty line in the end of file appears so that standard reading from the input stream will know when to terminate the read, usually returns EOF to indicate that you have reached the end.

eritbh:
Unlike most text editors, Github doesn't show final newlines as their own lines; instead, it shows a 🚫 symbol at the end of files which do not include a final newline. This file formerly ended with two newlines, and this PR changes it to only end with one. I've confirmed this by checking out the file in my local editor.

I think this change is valid--at least, as valid as any one-line style PR, depending on how you feel about that. This and #1229 might be merged into a single style PR, along with any other changes that could be addressed at the same time; otherwise it may be smart to just handle this the next time the file needs to be edited for another reason.

abalabahaha:
Merged into #1229

PR #1229, opened 2021-06-26
Use "const" over "var" in README

J-Human:
This prefers the use of constants over the "var" keyword for the bot sample located in the ReadMe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,12 @@ Ping Pong Example
 ```js
 const Eris = require("eris");
 
-const bot = new Eris("Bot TOKEN");
 // Replace TOKEN with your bot account's token
+const bot = new Eris("Bot TOKEN", {
+    intents: [
+        "guildMessages"
+    ];
+});
 
 bot.on("ready", () => { // When the bot is ready
     console.log("Ready!"); // Log "Ready!"
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Ping Pong Example
 ```js
 const Eris = require("eris");
 
-var bot = new Eris("Bot TOKEN");
+const bot = new Eris("Bot TOKEN");
 // Replace TOKEN with your bot account's token
 
 bot.on("ready", () => { // When the bot is ready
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -20,7 +20,7 @@ Ping Pong Example
 ```js
 const Eris = require("eris");
 
-var bot = new Eris("Bot TOKEN");
+const bot = new Eris("Bot TOKEN");
```
reinacchi: Due to PR #1038 was merged, intents are now required for bots to operate.

reinacchi (changes requested):

ghost:
You specified the intents, and that seems fine, but changing `var bot` to `const bot` does not actually matter much as nobody will be stupid enough to actually reassign the `bot` into something else.

bsian03:
Which is also the reason why the PR hasn't been merged and why it has the farming label applied to it. Kinda obvious

J-Human:
I don't have any bad intention regarding this repository but showing this kind of tolerance seems a bit too much, honestly speaking. It's quite simple:
* What's the exclusive "farming" tag for? It simply shows you can't handle situations well and instead shaming the author. Maybe renaming it into "invalid" would do. Furthermore, this isn't farming, I wasn't doing anything a lot of times.
* If this PR is too low for your standards, why don't you close it? 

What's the purpose for not minding this as if it's useless? I get avergateftw's point but I feel having a good code convention in your sample will help. It just shows that you have good habits.

bsian03:
The PR is valid, it's just low effort. Those kinds of PRs should have the farming label applied. I'm not discriminating your PRs, I've had to apply the same label on my own PRs updating the readme

bsian03:
My wording earlier might not've been the best (I did literally just wake up) so if it came across the wrong way, apologies

README.md
```diff
@@ -20,8 +20,12 @@ Ping Pong Example
 ```js
 const Eris = require("eris");
 
-var bot = new Eris("Bot TOKEN");
 // Replace TOKEN with your bot account's token
+const bot = new Eris("Bot TOKEN", {
+    intents: [
+        "guildMessages"
+    ];
```
abalabahaha: Syntax error

abalabahaha (approved):

PR #1019, opened 2020-09-20
Add tag getter to User class

er2off:

PR #1231, opened 2021-07-01
Update Member.js

GodyFromDiscord:
Simple function addition to the Member class, which allows you to perform `message.author.tag` etc, simply returns their username and discriminator; Gody#0378 instead of having to do ${message.author.username}#${message.author.discriminator} in your code everywhere you wanna put their user and discriminator together. Although this can be done easily it just makes more sense to be added as a function itself.

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -152,6 +152,12 @@ class Member extends Base {
     get username() {
         return this.user.username;
     }
+    
+    get tag() {
+        return `${this.user.username}#${this.user.discriminator}`
+    }
+    
+    
 
     get voiceState() {
         if(this.guild && this.guild.voiceStates.has(this.id)) {


Catboi8:
Personally i don't see much need for this, its just extra that honestly isn't even too useful and its already easily done with \`${message.author.username}#${message.author.discriminator}\`

JustCat80 (commented):
Personally i don't see much need for this, its just extra that honestly isn't even too useful and its already easily done with `${message.author.username}#${message.author.discriminator}`

ImDarkDiamond:
This is a duplicate of #252 and a few other PRs too.

JustCat80:
[Oop yea](https://github.com/abalabahaha/eris/pull/252#issuecomment-312463946)

TakenKills:
yes, and you can also just define that yourself using `Object.defineProperty`

GodyFromDiscord:
Ah well it seems it's been requested before so it'll most likely be denied, was just a small addition and as I did say it could be easily done with the one liner, but it just makes it quicker, but seeing as there's multiple go arounds to this it can be closed

DonovanDMC:
6th pr for this 🎉
#1019 #925 #504 #325 #252

PR #1233, opened 2021-07-03
Fix voiceState.selfStream when user stops streaming

ak-s:
Discord never sends `self_stream: false` in my tests. This field is [optional](https://discord.com/developers/docs/resources/voice#voice-state-object) and does not appear in VOICE_STATE_UPDATE packet when user is not streaming.


bsian03 (dismissed):
This change will mean that every time Discord sends a `member` payload (e.g. `MESSAGE_CREATE`), `selfStream` will always be set to false, even if the member is sharing their screen. To avoid this i'd recommend checking if one of the other `self_` props exist and if it does, then put all `self_` props inside that (since member only contains `mute` and `deaf`)

If you join the server either I (or Piero) will be able to explain this better

ak-s:
I got it. But I would not like to "tie" `self_` props together. How about adding a missing `self_stream` during processing `VOICE_STATE_UPDATE` packet?
You are using a similar approach here: https://github.com/abalabahaha/eris/pull/1217/files#diff-0e7a53040939ede35855c97c923a8a2e2cc6a88afe0c106ac9a34ec262dbc604R742

bsian03:
(abal please see diff and the linked PR commit)

abalabahaha (approved):

PR #1235, opened 2021-07-03
Update user flag names

Awoocado:


bsian03 (changes requested):
Early verified bot developer as well

GoldenAngel2 (changes requested):
`CERTIFIED_MODERATOR` should be changed to: `DISCORD_CERTIFIED_MODERATOR` 
> https://discord.com/developers/docs/resources/user#user-object-user-flags

So the name lines up with what is currently documented.

untuned:
> `CERTIFIED_MODERATOR` should be changed to: `DISCORD_CERTIFIED_MODERATOR`
> 
> > https://discord.com/developers/docs/resources/user#user-object-user-flags
> 
> So the name lines up with what is currently documented.

It's internally named `CERTIFIED_MODERATOR`.

bsian03 (approved):

PR #1237, opened 2021-07-03
Update sticker properties

conorwastakenwastaken:
These changes will deprecate the old `stickers` property and use `stickerItems` instead. Since Discord no longer sends `sticker`, `sticker_items` should be used instead.

The `User` property sent with the sticker (if any) is no longer partial and will update the user cache.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -875,20 +875,21 @@ declare namespace Eris {
     messageID: string;
     failIfNotExists?: boolean;
   }
-  interface StickerItems {
-    id: string;
-    name: string;
-    format_type: Constants["StickerFormats"][keyof Constants["StickerFormats"]];
-  }
   interface Sticker extends StickerItems {
+    /** @deprecated */
     asset: "";
     available?: boolean;
     description: string;
     guild_id?: string;
     pack_id?: string;
     sort_value?: number;
-    tags?: string;
-    user?: PartialUser;
+    tags: string;
+    user?: User;
+  }
+  interface StickerItems {
+    id: string;
+    name: string;
+    format_type: Constants["StickerFormats"][keyof Constants["StickerFormats"]];
   }
 
   // Presence
@@ -2259,9 +2260,9 @@ declare namespace Eris {
     reactions: { [s: string]: { count: number; me: boolean } };
     referencedMessage?: Message | null;
     roleMentions: string[];
-    sticker_items?: Sticker[];
     /** @deprecated */
-    stickers?: Sticker[];
+    stickers?: StickerItems[];
+    stickerItems?: StickerItems[];
     timestamp: number;
     tts: boolean;
     type: number;
--- a/index.d.ts
+++ b/index.d.ts
@@ -881,6 +881,7 @@ declare namespace Eris {
     format_type: Constants["StickerFormats"][keyof Constants["StickerFormats"]];
   }
   interface Sticker extends StickerItems {
+    asset: "";
     available?: boolean;
     description: string;
     guild_id?: string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -875,13 +875,15 @@ declare namespace Eris {
     messageID: string;
     failIfNotExists?: boolean;
   }
-  interface Sticker {
-    available?: boolean;
-    description?: string;
-    format_type: Constants["StickerFormats"][keyof Constants["StickerFormats"]];
-    guild_id?: string;
+  interface StickerItems {
     id: string;
     name: string;
+    format_type: Constants["StickerFormats"][keyof Constants["StickerFormats"]];
+  }
+  interface Sticker extends StickerItems {
+    available?: boolean;
+    description: string;
+    guild_id?: string;
     pack_id?: string;
     sort_value?: number;
     tags?: string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -876,13 +876,16 @@ declare namespace Eris {
     failIfNotExists?: boolean;
   }
   interface Sticker {
-    asset: string;
-    description: string;
+    available?: boolean;
+    description?: string;
     format_type: Constants["StickerFormats"][keyof Constants["StickerFormats"]];
+    guild_id?: string;
     id: string;
     name: string;
-    pack_id: string;
+    pack_id?: string;
+    sort_value?: number;
     tags?: string;
+    user?: PartialUser;
   }
 
   // Presence
--- a/index.d.ts
+++ b/index.d.ts
@@ -2261,7 +2261,7 @@ declare namespace Eris {
     referencedMessage?: Message | null;
     roleMentions: string[];
     /** @deprecated */
-    stickers?: StickerItems[];
+    stickers?: Sticker[];
     stickerItems?: StickerItems[];
     timestamp: number;
     tts: boolean;
--- a/index.d.ts
+++ b/index.d.ts
@@ -2258,6 +2258,8 @@ declare namespace Eris {
     reactions: { [s: string]: { count: number; me: boolean } };
     referencedMessage?: Message | null;
     roleMentions: string[];
+    sticker_items?: Sticker[];
+    /** @deprecated */
     stickers?: Sticker[];
     timestamp: number;
     tts: boolean;
--- a/lib/structures/Message.js
+++ b/lib/structures/Message.js
@@ -292,6 +292,9 @@ class Message extends Base {
                 };
             });
         }
+        if(data.stickers !== undefined) {
+            this.stickers = data.stickers;
+        }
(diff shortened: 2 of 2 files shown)


lib/structures/Message.js
```diff
@@ -291,8 +291,8 @@ class Message extends Base {
                 };
             });
         }
-        if(data.stickers !== undefined) {
-            this.stickers = data.stickers;
+        if(data.sticker_items !== undefined) {
+            this.stickers = data.sticker_items;
```
bsian03: Can the old prop be deprecated and have `sticker_items` be a new prop

index.d.ts
```diff
@@ -876,13 +876,16 @@ declare namespace Eris {
...
     format_type: Constants["StickerFormats"][keyof Constants["StickerFormats"]];
+    guild_id?: string;
     id: string;
     name: string;
-    pack_id: string;
+    pack_id?: string;
+    sort_value?: number;
     tags?: string;
+    user?: PartialUser;
   }
```
bsian03: Can the Sticker interface extend a newly created `StickerItem` interface with the 3 relevant props

index.d.ts
```diff
@@ -876,13 +876,16 @@ declare namespace Eris {
     failIfNotExists?: boolean;
   }
   interface Sticker {
-    asset: string;
```
bsian03, line 879: Just mark as empty string

index.d.ts
```diff
@@ -876,13 +876,16 @@ declare namespace Eris {
...
-    description: string;
+    available?: boolean;
+    description?: string;
```
bsian03: Description is still guaranteed

bsian03 (changes requested):
I'm thinking we can also do a User cache update for `sticker.user` instead of leaving it as a plain object

index.d.ts
```diff
@@ -875,14 +875,20 @@ declare namespace Eris {
...
-    asset: string;
-    description: string;
-    format_type: Constants["StickerFormats"][keyof Constants["StickerFormats"]];
+  interface StickerItems {
     id: string;
     name: string;
-    pack_id: string;
+    format_type: Constants["StickerFormats"][keyof Constants["StickerFormats"]];
+  }
+  interface Sticker extends StickerItems {
```
bsian03, line 878: Sticker before StickerItems (alphabetical order)

index.d.ts
```diff
@@ -875,14 +875,20 @@ declare namespace Eris {
...
+    asset: "";
```
bsian03, line 880: Add deprecated flag

lib/structures/Message.js
```diff
@@ -291,8 +292,15 @@ class Message extends Base {
                 };
             });
         }
-        if(data.stickers !== undefined) {
-            this.stickers = data.stickers;
+        if(data.sticker_items !== undefined) {
+            this.sticker_items = data.sticker_items.map((sticker) => {
```
bsian03: camelCase (also update JSDoc ordering when you change this)

index.d.ts
```diff
@@ -2253,6 +2259,8 @@ declare namespace Eris {
     reactions: { [s: string]: { count: number; me: boolean } };
     referencedMessage?: Message | null;
     roleMentions: string[];
+    sticker_items?: Sticker[];
```
bsian03: StickerItems (as per discord docs)

lib/structures/Message.js
```diff
@@ -291,8 +292,15 @@ class Message extends Base {
...
+                if(sticker.user) {
+                    sticker.user = client.users.add(sticker.user, client);
```
bsian03: Update instead of add

index.d.ts
```diff
@@ -875,14 +875,20 @@ declare namespace Eris {
...
+  }
+  interface Sticker extends StickerItems {
+    asset: "";
+    available?: boolean;
+    description: string;
+    guild_id?: string;
+    pack_id?: string;
+    sort_value?: number;
     tags?: string;
+    user?: PartialUser;
```
bsian03: Should just be the user class

index.d.ts
```diff
@@ -875,14 +875,20 @@ declare namespace Eris {
...
+    format_type: Constants["StickerFormats"][keyof Constants["StickerFormats"]];
+  }
+  interface Sticker extends StickerItems {
+    asset: "";
+    available?: boolean;
+    description: string;
+    guild_id?: string;
+    pack_id?: string;
+    sort_value?: number;
     tags?: string;
```
bsian03: Guaranteed

bsian03 (changes requested):

lib/structures/Message.js
conorwastakenwastaken: Well I preferred camelCase too, but you said `sticker_items` so I didn't bother questioning it haha, thanks & will do 👍🏻

index.d.ts
conorwastakenwastaken: Not sure how I missed that, thanks.

index.d.ts
```diff
@@ -2253,7 +2260,9 @@ declare namespace Eris {
     reactions: { [s: string]: { count: number; me: boolean } };
     referencedMessage?: Message | null;
     roleMentions: string[];
-    stickers?: Sticker[];
+    /** @deprecated */
+    stickers?: StickerItems[];
```
bsian03: This is still an array of Stickers, not StickerItems

lib/structures/Message.js
```diff
@@ -291,8 +292,15 @@ class Message extends Base {
...
-            this.stickers = data.stickers;
+        if(data.sticker_items !== undefined) {
+            this.stickerItems = data.sticker_items.map((sticker) => {
+                if(sticker.user) {
+                    sticker.user = client.users.update(sticker.user, client);
+                }
+
+                return sticker;
+            });
+            this.stickers = this.stickerItems;
```
bsian03: Please still use the deprecated `data.stickers`, it shouldn't be changed completely to match another reference for backwards compatibility/preventing a breaking change

bsian03 (changes requested):

lib/structures/Message.js
```diff
@@ -506,7 +516,7 @@ class Message extends Base {
             "reactions",
             "referencedMesssage",
             "roleMentions",
-            "stickers",
```
bsian03, line 509: Still keep this (we will remove it when the whole prop gets removed in a future API version)

bsian03 (changes requested):

lib/structures/Message.js
```diff
@@ -294,6 +295,15 @@ class Message extends Base {
         if(data.stickers !== undefined) {
```
curtisf, line 296: Could you add a newline before this if to keep the newlines in Message more consistent?

lib/structures/Message.js
```diff
@@ -294,6 +295,15 @@ class Message extends Base {
...
             this.stickers = data.stickers;
         }
```
curtisf, line 298: Newline after this too please

lib/structures/Message.js
curtisf, line 296: Q: In what cases do `stickers` get sent over `stickerItems` now? Could be doing something wrong, but interestingly I couldn't see a single instance where `stickers` was present when I was listening for raw MESSAGE_CREATEs containing stickers (even though it's just marked as deprecated in the docs). I tried custom and stock stickers and neither had `stickers` present.

curtisf (commented):
Other than the comments I left, tested this and it works fine. Thanks!

lib/structures/Message.js
conorwastakenwastaken: They never do. Discord marked the property as deprecated in the docs, however in reality it was actually removed and replaced with `sticker_items`. In my original commits, I simply changed the Message#stickers property to use `sticker_items`, but Bsian felt it was necessary to deprecate the old value and add `stickerItems` to keep things consistent. 👍🏻

conorwastakenwastaken: Done, thanks!

curtisf (approved):

lib/structures/Message.js
bsian03: I did just confirm - old messages before the change still use `stickers` instead of `sticker_items` (as such, this change would break backwards compatibility with those messages)

bsian03 (approved):
Will need testing unless piero has done that already (I'm unable to do that atm), otherwise RTM
cc @curtisf

lib/structures/Message.js
conorwastakenwastaken: Ahh that makes sense, thanks for clarifying.

PR #1239, opened 2021-07-04
Support sending multiple embeds

reinacchi:
https://discord.com/developers/docs/resources/channel#create-message-jsonform-params
https://discord.com/developers/docs/resources/channel#edit-message-jsonform-params

Discord has deprecated embed when creating/editing messages and has changed to embeds. It's better to switch from the regular embed to embeds before Discord might remove them in the future API version 👍🏻

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -604,6 +604,8 @@ class Client extends EventEmitter {
                 content.content = "" + content.content;
             } else if(content.content === undefined && !content.embed && !content.embeds && !file) {
                 return Promise.reject(new Error("No content, file, or embeds"));
+            } else if(content.embed && content.embeds) {
+                content.embeds.push(content.embed);
             } else if(content.embed) {
                 this.emit("warn", "[DEPRECATED] content.embed is deprecated. Use content.embeds instead");
                 content.embeds = [content.embed];
@@ -1241,6 +1243,8 @@ class Client extends EventEmitter {
                 content.content = "" + content.content;
             } else if(content.content === undefined && !content.embed && !content.embeds && content.flags === undefined) {
                 return Promise.reject(new Error("No content, embeds or flags"));
+            } else if(content.embed && content.embeds) {
+                content.embeds.push(content.embed);
             } else if(content.embed) {
                 this.emit("warn", "[DEPRECATED] content.embed is deprecated. Use content.embeds instead");
                 content.embeds = [content.embed];
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -581,6 +581,7 @@ class Client extends EventEmitter {
     * @arg {Boolean} [content.allowedMentions.repliedUser] Whether or not to mention the author of the message being replied to.
     * @arg {String} content.content A content string
     * @arg {Object} [content.embed] An embed object. See [the official Discord API documentation entry](https://discord.com/developers/docs/resources/channel#embed-object) for object structure
+    * @arg {Array<Object>} [content.embeds] An array of embed objects. See [the official Discord API documentation entry](https://discord.com/developers/docs/resources/channel#embed-object) for object structure
     * @arg {Object} [content.messageReference] The message reference, used when replying to messages
     * @arg {String} [content.messageReference.channelID] The channel ID of the referenced message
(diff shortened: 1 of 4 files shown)


lib/Client.js
```diff
@@ -601,8 +602,10 @@ class Client extends EventEmitter {
                 };
             } else if(content.content !== undefined && typeof content.content !== "string") {
                 content.content = "" + content.content;
-            } else if(content.content === undefined && !content.embed && !file) {
-                return Promise.reject(new Error("No content, file, or embed"));
+            } else if(content.content === undefined && !content.embed && !content.embeds && !file) {
```
ghost, line 605: You should probably check if embeds length is above 0.

lib/Client.js
```diff
@@ -581,6 +581,7 @@ class Client extends EventEmitter {
     * @arg {Boolean} [content.allowedMentions.repliedUser] Whether or not to mention the author of the message being replied to.
     * @arg {String} content.content A content string
     * @arg {Object} [content.embed] An embed object. See [the official Discord API documentation entry](https://discord.com/developers/docs/resources/channel#embed-object) for object structure
```
bsian03: Mark deprecated

lib/Client.js
bsian03, line 605: I'd rather have the deprecated prop reassigned into an `embeds` array instead of still sending it through to Discord

bsian03 (changes requested):

lib/Client.js
```diff
@@ -601,8 +602,11 @@ class Client extends EventEmitter {
                 };
             } else if(content.content !== undefined && typeof content.content !== "string") {
                 content.content = "" + content.content;
-            } else if(content.content === undefined && !content.embed && !file) {
-                return Promise.reject(new Error("No content, file, or embed"));
+            } else if(content.content === undefined && !content.embed && !content.embeds && !file) {
+                return Promise.reject(new Error("No content, file, or embeds"));
+            } else if(content.embed) {
```
bsian03: Check if anything exists in `embeds` to prevent accidental overriding

bsian03 (changes requested):
This was just a quick look through, I may have other things to nitpick later

lib/Client.js
reinacchi: What'd you exactly meant by "prevent accidental overriding"??

bsian03: Right now if someone provides `embed` and `embeds`, then `embeds` will be overwritten

reinacchi: Do you want if someone provides both `embed` and `embeds`, it'll send both or only `embeds`? 😅

JustCat80: You'd want it to add embed to embeds rather then override I think, no?

bsian03: If someone provides both `embed` and `embeds`, send both and let Discord handle that
FI `embeds` will override `embed` if it is supplied

lib/Client.js
```diff
@@ -601,8 +602,13 @@ class Client extends EventEmitter {
                 };
             } else if(content.content !== undefined && typeof content.content !== "string") {
                 content.content = "" + content.content;
-            } else if(content.content === undefined && !content.embed && !file) {
-                return Promise.reject(new Error("No content, file, or embed"));
+            } else if(content.content === undefined && !content.embed && !content.embeds && !file) {
+                return Promise.reject(new Error("No content, file, or embeds"));
+            } else if(content.embed && content.embeds) {
+                content.embeds.push(content.embed);
+            } else if(content.embed) {
```
bsian03: No, just check if embed exists and if embeds doesn't exist, and if both those conditions are true, then set content.embeds to [content.embed]
I'm really not sure why you're complicating things by pushing (which will break anyways if someone puts the wrong thing in embeds)
Also deprecation will not get captured in this case, hence my request

bsian03 (changes requested):

reinacchi (commented):
Please, kindly correct me if I made another mistake again.

bsian03 (approved):

JustCat80 (approved):

PR #1240, opened 2021-07-05
Updating the README.md

GodyFromDiscord:
Would it not make sense to have this updated? Mainly seeing as the server is somewhat plugged in the DAPI server and is where a mainstream of help can be provided by other users. Along with it's vanity as well it seems more liable. 

The invite was also updated anyway so it would make sense to have the name changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Useful Links
 ------------
 
 - [The website](https://abal.moe/Eris/) has more details and documentation.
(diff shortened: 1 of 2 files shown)


GodyFromDiscord:
I have just realised this was already done on the dev branch but not yet on master

bsian03:
#1199

PR #1242, opened 2021-07-08
Better types for events

HeadTriXz:
Changed the interfaces to tuples instead, which makes it easier for people to do stuff like `keyof ClientEvents` giving you all event names as a type. 

Also added **once**, **off** and **emit**.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -490,99 +490,96 @@ declare namespace Eris {
     selfStream: boolean;
     selfVideo: boolean;
   }
-  interface EventListeners<T> {
-    (event: "ready" | "disconnect", listener: () => void): T;
-    (event: "callCreate" | "callRing" | "callDelete", listener: (call: Call) => void): T;
-    (event: "callUpdate", listener: (call: Call, oldCall: OldCall) => void): T;
-    (event: "channelCreate" | "channelDelete", listener: (channel: AnyChannel) => void): T;
-    (
-      event: "channelPinUpdate",
-      listener: (channel: TextableChannel, timestamp: number, oldTimestamp: number) => void
-    ): T;
-    (
-      event: "channelRecipientAdd" | "channelRecipientRemove",
-      listener: (channel: GroupChannel, user: User) => void
-    ): T;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -490,99 +490,96 @@ declare namespace Eris {
...
-    (event: "connect" | "end" | "ready" | "start", listener: () => void): T;
-    (event: "debug" | "warn", listener: (message: string) => void): T;
-    (event: "disconnect", listener: (err?: Error) => void): T;
-    (event: "error", listener: (err: Error) => void): T;
-    (event: "pong", listener: (latency: number) => void): T;
-    (event: "speakingStart" | "speakingStop" | "userDisconnect", listener: (userID: string) => void): T;
-    (event: "unknown", listener: (packet: unknown) => void): T;
+  interface EventListeners {
+    callCreate: [call: Call];
+    callDelte: [call: Call];
```
aetheryx: typo

DonovanDMC:
finally, events as tuples, so I can stop doing all of my custom event stuff manually

DonovanDMC (approved):

coolcalcacol:
Perhaps also add removeListener for each event?

index.d.ts
```diff
@@ -1837,7 +1833,14 @@ declare namespace Eris {
     unbanGuildMember(guildID: string, userID: string, reason?: string): Promise<void>;
     unpinMessage(channelID: string, messageID: string): Promise<void>;
     validateDiscoverySearchTerm(term: string): Promise<{ valid: boolean }>;
-    on: ClientEvents<this>;
+    emit<K extends keyof ClientEvents>(event: K, ...args: ClientEvents[K]): boolean;
+    emit(event: string, ...args: any[]): boolean;
```
bsian03, line 1837: Don't think we need `emit` to be typed as the only time it'll be used is within Eris internals

DonovanDMC: I personally use it to reemit interactions for regular command handling (causes less duplicated code, and I don't feel like reinitializing my event listeners)

And what can something like this really hurt? It still allows for the addition of arbitrary events via the last line

bsian03: I feel like for a normal use case, including this seems pretty pointless

abalabahaha: Eh seems ok

bsian03 (approved):

PR #1245, opened 2021-07-15
Fix status

reinacchi:
`"offline"` doesn't exist instead it should be `"invisible"`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -135,7 +135,9 @@ declare namespace Eris {
   type ActivityType = BotActivityType | Constants["ActivityTypes"]["CUSTOM"];
   type BotActivityType = Constants["ActivityTypes"][Exclude<keyof Constants["ActivityTypes"], "CUSTOM">];
   type FriendSuggestionReasons = { name: string; platform_type: string; type: number }[];
-  type Status = "online" | "idle" | "dnd" | "invisible" | "offline";
+  type Status = "online" | "idle" | "dnd";
+  type SelfStatus = Status | "invisible";
+  type UserStatus = Status | "offline";
 
   // Selfbot
   type ConnectionVisibilityTypes = Constants["ConnectionVisibilityTypes"][keyof Constants["ConnectionVisibilityTypes"]];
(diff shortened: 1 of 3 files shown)


Dramex:
According to discord api docs it's `offline` not `invisable`

source: 
https://discord.dev/topics/gateway#presence-update-presence-update-event-fields

DonovanDMC:
As far as the gateway is concerned, offline and invisible are the same thing
docs say `offline`, so it should be `offline`

reinacchi:
`Client.editStatus("offline")` won't change bots status to fully offline, this is why I made this pr since it works when `offline` is replaced with `invisible` and that's what is written down in the [Eris Docs](https://abal.moe/Eris/docs/Client#method-editStatus)

I had already checked in the Discord API docs if the gateway expects `offline` and not `invisible` but as said, I'm making this pr to fix `Client#editStatus` as `offline` doesn't affect bots status, it'll still remain online.

bsian03:
I'll allow the extra type for `invisible`, but not completely replace `offline`. Setting a status to `offline` and it not working should be reported as a bug/outdated documentation to the API repo.

I'll also mark PR as draft until this is clarified.

DonovanDMC:
According to the docs, [both invisible and offline exists](https://discord.com/developers/docs/topics/gateway#update-presence-status-types) (mistake on my part), so invisible should be added, not replacing offline

bsian03:
> According to discord api docs it's `offline` not `invisable`

You're looking at the wrong location, that's for gateway presence update event, not Update presence command

Dramex:
> > According to discord api docs it's `offline` not `invisable`
> 
> You're looking at the wrong location, that's for gateway presence update event, not Update presence command

Apologize thoguht this PR for presence update🤦‍♂️, then in this case yes invisible replaces offline.

reinacchi:
> According to the docs, [both invisible and offline exists](https://discord.com/developers/docs/topics/gateway#update-presence-status-types) (mistake on my part), so invisible should be added, not replacing offline

You're right. Just checked on that recently. But since afaik that `offline` can only be received, and `invisible` can only be sent through the gateway, does `invisible` still need to be added or replaced fully by `offline`?

bsian03:
Refer to my earlier comment.

bsian03:
Can we separate invisble and offline types and use them in their relevant references

bsian03 (approved):

PR #1248, opened 2021-07-20
fix some jsdoc for editRole & createRole 

DonovanDMC:
createRole is missing Bigint, and there's a difference in the types (missing Number/String) between editRole & createRole


DonovanDMC:
reeee

bsian03 (approved):

PR #1251, opened 2021-07-28
Added "thread" to MessageTypes & MessageFlags

ray-1337:
return a new error "Unhandled MESSAGE_CREATE type" recently, should be fixed

image


Dramex:
there is already opened PR #1216

ray-1337:
[already opened]

PR #1252, opened 2021-07-28
Sticker Sending

DonovanDMC:
Through inspecting client network requests, and testing with bots, sending `sticker_ids` allows bots to send stickers (multiple as well).

[postman request](https://butts-are.cool/DiscordCanary_07-28-2021_05-32-41.png)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -53,20 +53,6 @@ declare namespace Eris {
   type VerificationLevel = 0 | 1 | 2 | 3 | 4;
 
   // Message
-  type AdvancedMessageContent = {
-    allowedMentions?: AllowedMentions;
-    components?: ActionRow[];
-    content?: string;
-    /** @deprecated */
-    embed?: EmbedOptions;
-    embeds?: EmbedOptions[];
-    flags?: number;
-    messageReference?: MessageReferenceReply;
-    /** @deprecated */
-    messageReferenceID?: string;
-    stickerIDs?: string[];
-    tts?: boolean;
-  };
   type ActionRowComponents = Button | SelectMenu;
   type Button = InteractionButton | URLButton;
   type Component = ActionRow | ActionRowComponents;
@@ -854,6 +840,21 @@ declare namespace Eris {
     args: string[];
     command: Command;
     timeout: NodeJS.Timer;
+  }
+  
+  interface AdvancedMessageContent {
+    allowedMentions?: AllowedMentions;
+    components?: ActionRow[];
+    content?: string;
+    /** @deprecated */
+    embed?: EmbedOptions;
+    embeds?: EmbedOptions[];
+    flags?: number;
+    messageReference?: MessageReferenceReply;
+    /** @deprecated */
+    messageReferenceID?: string;
+    stickerIDs?: string[];
+    tts?: boolean;
   }
   interface AllowedMentions {
     everyone?: boolean;
--- a/index.d.ts
+++ b/index.d.ts
@@ -2304,7 +2304,6 @@ declare namespace Eris {
     stickerItems?: StickerItems[];
     /** @deprecated */
     stickers?: Sticker[];
-
     timestamp: number;
     tts: boolean;
     type: number;
--- a/index.d.ts
+++ b/index.d.ts
@@ -62,7 +62,7 @@ declare namespace Eris {
     messageReference?: MessageReferenceReply;
     /** @deprecated */
     messageReferenceID?: string;
-    stickerIDS?: string[];
+    stickerIDs?: string[];
     tts?: boolean;
   };
   type ImageFormat = "jpg" | "jpeg" | "png" | "gif" | "webp";
--- a/index.d.ts
+++ b/index.d.ts
@@ -839,7 +839,6 @@ declare namespace Eris {
     command: Command;
     timeout: NodeJS.Timer;
   }
-  
   interface AdvancedMessageContent {
     allowedMentions?: AllowedMentions;
     components?: ActionRow[];
--- a/index.d.ts
+++ b/index.d.ts
@@ -62,6 +62,7 @@ declare namespace Eris {
     messageReference?: MessageReferenceReply;
     /** @deprecated */
     messageReferenceID?: string;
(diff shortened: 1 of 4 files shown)


DonovanDMC:
Bots only seem to be able to send stickers from the current server, so sending in direct messages is basically impossible

DonovanDMC:
Didn't think this was in the docs, but they are


Also, since the max is explicitly 3, we could do `[string?, string?, string?]` for between 0 and 3, but that might not go with the general types style around here

index.d.ts
```diff
@@ -2300,9 +2301,10 @@ declare namespace Eris {
     reactions: { [s: string]: { count: number; me: boolean } };
     referencedMessage?: Message | null;
     roleMentions: string[];
+    stickerItems?: StickerItems[];
     /** @deprecated */
     stickers?: Sticker[];
-    stickerItems?: StickerItems[];
+
```
conorwastakenwastaken: Random newline here 👀

conorwastakenwastaken (changes requested):
Also, is there a reason the property is called `stickerIDS`? Other properties (such as `channelIDs`, `guildIDs`, etc.) don't have a capital S at the end, so I'm just curious if there was a reason tbh

DonovanDMC:
> Also, is there a reason the property is called `stickerIDS`? Other properties (such as `channelIDs`, `guildIDs`, etc.) don't have a capital S at the end, so I'm just curious if there was a reason tbh

the reason is I didn't realize they were like that

lib/Client.js
```diff
@@ -588,6 +588,7 @@ class Client extends EventEmitter {
     * @arg {String} [content.messageReference.guildID] The guild ID of the referenced message
     * @arg {String} content.messageReference.messageID The message ID of the referenced message. This cannot reference a system message
     * @arg {String} [content.messageReferenceID] [DEPRECATED] The ID of the message should be replied to. Use `messageReference` instead
+    * @arg {Array<String>} [content.stickerIDs] An array of ids corresponding to stickers to send
```
conorwastakenwastaken: I didn't notice this earlier, but for consistency reasons, should `ids` be changed to `IDs` here? And the same for the JS docs in PrivateChannel & TextChannel.

index.d.ts
```diff
@@ -2300,9 +2301,9 @@ declare namespace Eris {
     reactions: { [s: string]: { count: number; me: boolean } };
     referencedMessage?: Message | null;
     roleMentions: string[];
+    stickerItems?: StickerItems[];
```
conorwastakenwastaken, line 2368: Idk how I didn't notice this was out of order when I made the PR for stickerItems, but I believe it's also alphabetized wrong in the Message structure too.

conorwastakenwastaken (changes requested):

conorwastakenwastaken:
> > Also, is there a reason the property is called `stickerIDS`? Other properties (such as `channelIDs`, `guildIDs`, etc.) don't have a capital S at the end, so I'm just curious if there was a reason tbh
> 
> the reason is I didn't realize they were like that

Ah, gotcha lol

lib/Client.js
```diff
@@ -603,8 +603,8 @@ class Client extends EventEmitter {
                 };
             } else if(content.content !== undefined && typeof content.content !== "string") {
                 content.content = "" + content.content;
-            } else if(content.content === undefined && !content.embed && !content.embeds && !file) {
-                return Promise.reject(new Error("No content, file, or embeds"));
+            } else if(content.content === undefined && !content.embed && !content.embeds && !file && (content.stickerIDS === undefined || content.stickerIDS.length === 0)) {
```
smhman: Why do you check twice that if all the others + stickers exists? Couldn't you just merge sticker check into the same if closure?

JustCat80: sir the red highlighted was removed and the green was added *lol the review I replied to was deleted*

conorwastakenwastaken:
Was 951bac5 meant to include a ton of spaces being added to the `if`/`catch` blocks?

DonovanDMC:
for future reference, in case anyone asks about it, it seems like stickers can't be sent via webhooks at all, since it gives an `Cannot send an empty message` error

index.d.ts
```diff
@@ -852,6 +839,21 @@ declare namespace Eris {
     command: Command;
     timeout: NodeJS.Timer;
   }
+
```
```suggestion
```

lib/Client.js
```suggestion
    * @arg {Array<String>} [content.stickerIDs] An array of IDs corresponding to stickers to send
```

lib/structures/PrivateChannel.js
```diff
@@ -72,6 +72,7 @@ class PrivateChannel extends Channel {
     * @arg {String} [content.messageReference.guildID] The guild ID of the referenced message
     * @arg {String} content.messageReference.messageID The message ID of the referenced message. This cannot reference a system message
     * @arg {String} [content.messageReferenceID] [DEPRECATED] The ID of the message should be replied to. Use `messageReference` instead
+    * @arg {Array<String>} [content.stickerIDs] An array of ids corresponding to the stickers to send
```
bsian03:
```suggestion
    * @arg {Array<String>} [content.stickerIDs] An array of IDs corresponding to the stickers to send
```

lib/structures/TextChannel.js
```diff
@@ -93,6 +93,7 @@ class TextChannel extends GuildChannel {
     * @arg {String} [content.messageReference.guildID] The guild ID of the referenced message
     * @arg {String} content.messageReference.messageID The message ID of the referenced message. This cannot reference a system message
     * @arg {String} [content.messageReferenceID] [DEPRECATED] The ID of the message should be replied to. Use `messageReference` instead
+    * @arg {Array<String>} [content.stickerIDs] An array of ids corresponding to the stickers to send
```
bsian03:
```suggestion
    * @arg {Array<String>} [content.stickerIDs] An array of IDs corresponding to the stickers to send
```

bsian03 (changes requested):
noob

bsian03 (approved):

PR #1254, opened 2021-08-04
Tweak eslint.ts

JustCat80:
i got annoyed so i found out how to disable it (I did try in the eslint file, but nothing I tried worked)


DonovanDMC:
It would be better to fully disable the rule in the eslint config, rather than just in this file

JustCat80:
> It would be better to fully disable the rule in the eslint config, rather than just in this file

yep done now, just found it, someone set it to 2 for some reason

bsian03:
What is the use case for the PR?

JustCat80:
> What is the use case for the PR?

Literally just to disable the 6 or so errors that don't actually mean anything when we run lint:ts
*it's very annoying*

bsian03:
@abalabahaha why did you enable this again

ESLint docs
> * `classes` (`boolean`) - The flag which shows whether or not this rule checks class declarations of upper scopes. If this is `true`, this rule warns every reference to a class before the class declaration. Otherwise, ignores those references if the declaration is in upper function scopes. Class declarations are not hoisted, so it might be danger. Default is `true`.

JustCat80:
classes was already disabled

abalabahaha:
This isn't default-on, wouldn't it be better to just delete from the config?

> @abalabahaha why did you enable this again

Probably me not knowing TypeScript, stupidity, or both

bsian03:
You're not you when you're hungry.

Rice. Get some carbs.

*Being serious tho, yeah it's better if it just went poof*

DonovanDMC:
I think we should also add in [@typescript-eslint/consistent-type-definitions](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-definitions.md) to keep `type` and `interface` usage consistent.

JustCat80:
Done, its nice that it auto fixes

abalabahaha (approved):

PR #1255, opened 2021-08-04
Require ClientOptions in Eris constructor

Ashesh3:
Class `Client` constructor makes `ClientOptions` a required parameter, since `intents` is no longer optional.

> Eris client options (only `intents` is required) 

However, the top level `Eris` function allows the user to mistakenly bypass this requirement of specifying `ClientOptions` with the `intent` property while calling the `Client` constructor.

This change should make it as so `ClientOptions` is required while calling `Eris` just as intended.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,7 +7,7 @@ import { URL } from "url";
 import { Socket as DgramSocket } from "dgram";
 import * as WebSocket from "ws";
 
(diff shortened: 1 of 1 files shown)


DonovanDMC:
mm Eric

Ashesh3:
> mm Eric

2am no coffee side effects, fixed

bsian03 (approved):

PR #1260, opened 2021-08-13
Prevent "User associated with Member not found"

retraigo:
When `client.users.limit` is set to `0`, new messages throw `Error: User associated with Member not found` due to the user not being cached. Virtually, it should also happen when the cache limit is set to a really low value and the client receives a lot of messages at once. 

As a fix, send "user" along with "member" when creating a Message. This should make the client use `data.user` in the absence of a cached user.

### DIFF
--- a/lib/structures/Message.js
+++ b/lib/structures/Message.js
@@ -124,7 +124,9 @@ class Message extends Base {
         if(this.channel.guild) {
             if(data.member) {
                 data.member.id = this.author.id;
-                if(data.author) data.member.user = data.author;
+                if(data.author) {
+                    data.member.user = data.author;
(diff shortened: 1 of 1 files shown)


JustCat80:
isn't the user already sent via message.author?

bit0r1n (changes requested):

retraigo:
> isn't the user already sent via message.author?

I meant the error thrown from the Member class when the user associated with the member is not found in cache. To reproduce this, just create an eris client with `client.users.limit` set to `0` and send a message in a channel the bot has access to.

abalabahaha (approved):

PR #1210, opened 2021-05-30
Support Interactions

JustCat80:
Moved to #1280

PR #1262, opened 2021-08-13
[Wip] CommandClient rework

JustCat80:
This pr will rework command client to use application commands as well as to improve the CommandClient in general

#1210 Needs to be pulled before this


bsian03:
We'd probably need to discuss design choices with slash commands before a PR is made. Perhaps a GitHub discussion would be best for this? I'm not able to set it up though

JustCat80:
@abalabahaha Enable discussions later maybe please?



> We'd probably need to discuss design choices with slash commands before a PR is made. Perhaps a GitHub discussion would be best for this? I'm not able to set it up though

Opened one https://github.com/Catboi8/eris/discussions/56

bsian03:
Will close this for #1266 discussion

PR #1265, opened 2021-08-19
Fixes for Partial ts objects

JustCat80:
PartialChannel.id isnt a number
PartialRole.id isnt a number


DonovanDMC:
*I already put these in #1252 because they're small changes that don't matter*
they really should just be shoved in another pull request instead of having their own
(I put them there because there's already a small change that was shoved on, since in theory it will get merged soon)

JustCat80:
fair enough xD

PR #1276, opened 2021-09-07
Use process warnings for deprecations

eritbh:
Addresses #1274. Adds process warning emits alongside existing deprecation-related client `warn` emits. To avoid emitting the same process warning multiple times, uses a wrapper around `process.emitWarning` that checks if the given warning code has already been used, and if so, skips emitting the warning again.


lib/util/emitDeprecation.js
```diff
@@ -0,0 +1,9 @@
+const emittedCodes = new Set();
+
+module.exports = function emitDeprecation(code, description) {
+    if(emittedCodes.has(code) ) {
+        return;
+    }
+    emittedCodes.add(code);
+    process.emitWarning(description, "DeprecationWarning", code);
+};
```
eritbh: Hopefully this matches what you meant by [this](https://github.com/abalabahaha/eris/discussions/1274#discussioncomment-1285642) @bsian03?

I've opted to use a `Set` because a quick search seemed to indicate `Set#has` is more performant than `Array#includes`, though the number of warnings we'll have is probably never going to be big enough for it to truly matter. Could use an array instead for simplicity.

lib/Client.js
```diff
@@ -286,6 +287,7 @@ class Client extends EventEmitter {
     */
     addMessageReaction(channelID, messageID, reaction, userID) {
         if(userID !== undefined) {
+            emitDeprecation("eris:1", "addMessageReaction() was called without an \"@me\" `userID` argument");
```
eritbh: I've made up our warning codes completely arbitrarily, just numbering them sequentially as I went. We could use more descriptive codes (e.g. associating them with issues/PRs that implement or describe the deprecation) or we could use a method like this; whatever works, as long as it's a standard naming scheme across the project.

JustCat80: I feel like to make it harder to overlap in different prs we should use the pr/issue or an actual name or something 
so for example either "eris:#1239"(maybe not because multiple different warnings so maybe a suffix of the deprecation number within the pr or smthn) or "eris:embed" or similar for Support sending multiple embeds (#1239)

lib/Client.js
```diff
@@ -286,6 +287,7 @@ class Client extends EventEmitter {
...
             this.emit("warn", "[DEPRECATED] addMessageReaction() was called without an \"@me\" `userID` argument");
```
JustCat80, line 328: Shouldnt these emit warns be removed? as the emitDeprecation should handle it no?
(github highlighted the wrong line or im dumb)

Catboi8: oop wrong line i meant the emit warn

eritbh: No, we're keeping them for now for backwards compatibility, but we can remove them in a future breaking change. See the summary I wrote in #1274.

JustCat80: Ah I see thats fair, kinda ironic that the deprecations are deprecated tho lol

lib/Client.js
bsian03: Could probably discuss this in the discussion

lib/util/emitDeprecation.js
bsian03: See my discussion comment

lib/util/emitDeprecation.js
```diff
@@ -0,0 +1,20 @@
+const warningMessages = {
+    CREATE_CHANNEL_OPTIONS: "createChannel() was called with a string `options` argument",
+    CREATE_CHANNEL_REASON: "createChannel() was called with a string `reason` argument",
```
eritbh, line 3: These two messages are basically the same, and both have the same root cause - the signature of `createChannel` has changed. Should we merge them into the same code?

eritbh: These two messages are basically the same, and both have the same root cause - the signature of `createChannel` has changed. Should we merge them into the same code?

eritbh (commented):
Changed structure around a bit in response to discussion activity. Looking for feedback on the codes themselves.

I also think it would be a good idea to rewrite some of these existing messages to include more context and provide migration paths away from the problem, rather than focusing on being short. How do we feel about this?

lib/util/emitDeprecation.js
bsian03: Yup

eritbh: Done in b3a75e4

lib/Client.js
eritbh: Discussed in https://github.com/abalabahaha/eris/discussions/1274#discussioncomment-1293529 and done in 672183b

lib/util/emitDeprecation.js
eritbh: Performance comparison done and discussed on Discord, https://discord.com/channels/831967755447828491/884930349119991849/885222846815678495, arrays are faster. Addressed in e3313be.

eritbh:
Rewrote the rest of the warning descriptions to be consistent with each other. I've elected not to copy these changes to warnings emitted on the client to avoid potentially breaking things, but if it's desired anyway, I can copy my changes to those as well. Please double check my grammar and format.

lib/util/emitDeprecation.js
```diff
@@ -0,0 +1,19 @@
...
+};
+
+const emittedCodes = [];
+
+module.exports = function emitDeprecation(code) {
+    if(emittedCodes.includes(code) ) {
+        return;
+    }
+    emittedCodes.push(code);
+    process.emitWarning(warningMessages[code], "DeprecationWarning", `eris:${code}`);
```
bsian03: Don't know how relevant this is since all of this is done in the backend but should invalid codes be ignored? Let's say someone accidentally pushes a typo

DonovanDMC: what about like `warningMessages[code] || Invalid Warning Code "${code}"`, so that it doesn't just stay hidden

eritbh: An invalid warning code would cause `warningMessages[code]` to be `undefined`, which results in a thrown error. Given this would come up immediately in testing any PR that adds new deprecation warnings, I don't think it's super important to explicitly handle this, but I can add it if you'd prefer the fallback behavior be more graceful anyway.

image

bsian03: Yes please implement a fallback

eritbh: Done in bc83b95.

eritbh:
Merged latest `dev` in again and this branch still appears to work fine for me, though I'd guess there are additional deprecated behaviors that have been introduced in the meantime which will need to be handled in a successive PR.

bsian03 (approved):

PR #1283, opened 2021-10-04
feat(connect): add token check

JustCat80:
Checks if token exists and is a string

Could add regex to check if the token is valid but I'm not sure that's worth it


reflqct (dismissed):
looks good

lib/Client.js
```diff
@@ -188,6 +188,14 @@ class Client extends EventEmitter {
             }
         }
 
+        if(token === undefined) {
```
eritbh: As it is, this check is redundant since `typeof undefined !== "string"` - it would be handled below.

If we're just adding runtime checks for the sake of runtime checks, can't hurt to take it further with `!token` to also check for empty strings. Might also make sense to do both checks at once and have a single error message ("Invalid token ${token}" or something like that would probably work well enough).

eritbh (changes requested):

lib/Client.js
JustCat80: fair ill do this

DonovanDMC: there are certain circumstances where someone doesn't need to provide a token (like with webhook only things), so I don't think we should toss out empty strings

Yes, they could just provide some text or whatever, but people generally won't provide an empty string by accident

eritbh: Is it valid to just not pass a token at all (resulting in `undefined`) in those cases?

JustCat80: oop okay

DonovanDMC: > Is it valid to just not pass a token at all (resulting in `undefined`) in those cases?

yes, which is likely why we don't have specific checks, but we can't really do types like that, so checks should in theory be put in place

honestly a specific rest/webhook only client would solve that issue, but that's a bit overkill

JustCat80: undefined already errors out aswell so

bsian03:
Want to stress that this isn't a change request but it makes more sense if this was put in the connect() method or in the request handler

curtisf:
Agreed about putting this check in connect(). The point about webhook clients not passing a token is valid (as I currently do so lol), and it makes a bit more sense to, if anything, put it in connect.

JustCat80:
Alright, I won't be able to really do this (unless I get bored and decide to do it on mobile) for just under 2 weeks but will do

abalabahaha (changes requested):
Just flagging as a change request. Check needs to move to `connect()`

JustCat80:
Done, also added regex check uwu

DonovanDMC (approved):

abalabahaha (approved):

PR #1285, opened 2021-10-16
Rework file attachments

### LINKED ISSUE #1356: `editMessage` and `editWebhookMessage` send attachments twice in the same request
When using `Client#editMessage`, `Client#editWebhookMessage`, or anything that depends on these functions (e.g. `CommandInteraction#editOriginalMessage`), the `file` key is not stripped from the input object. This results in attachments being sent over simultaneously in raw binary form via the `file` form data entry, and as a JSON array converted from a Node buffer in the `payload_data` entry:

image

This can result in requests that are much larger than they should be (e.g. a 4MB request being inflated to 16MB). It would probably be a good idea to remove `file` from the options object in `RequestHandler#request` if the file argument is also present.

bsian03:
This should futureproof endpoints requiring a multipart/form-data with a JSON body


index.d.ts
```diff
@@ -1399,6 +1406,7 @@ declare namespace Eris {
   }
   interface WebhookPayload {
     allowedMentions?: AllowedMentions;
+    attachments: PartialAttachment[];
```
DonovanDMC:
```suggestion
    attachments?: PartialAttachment[];
```
This should be optional

lib/rest/RequestHandler.js
```diff
@@ -108,36 +108,26 @@ class RequestHandler {
...
+                            data.attach("files[0]", file.file, file.name);
                         } else {
                             throw new Error("Invalid file object");
                         }
+                        if(body) {
+                            for(const key in body) {
+                                data.attach(key, body[key]);
+                            }
+                        }
+                        data = data.finish();
```
DonovanDMC: `payload_json` is still a thing, this does not work

DonovanDMC (dismissed):
see above

lib/Client.js
```diff
@@ -2095,16 +2128,7 @@ class Client extends EventEmitter {
             }
```
DonovanDMC, line 2140: the above code for `embed` needs to be changed to assign to `data`

DonovanDMC (changes requested):

lib/Client.js
```diff
@@ -1987,13 +2003,14 @@ class Client extends EventEmitter {
...
-            if(!options.embeds) {
-                options.embeds = [];
-            }
-            options.embeds.push(options.embed);
+        const file = options.file;
+        delete options.file;
+
+        if(options.embed && !options.embeds) {
+            options.embeds = [options.embed];
         }
```
DonovanDMC: this shouldn't be conditional on embeds not being present, they should be combined as the code was before

index.d.ts
```diff
@@ -157,7 +157,7 @@ declare namespace Eris {
   type StageInstancePrivacyLevel = Constants["StageInstancePrivacyLevel"][keyof Constants["StageInstancePrivacyLevel"]];
 
   // Webhook
-  type MessageWebhookContent = Pick<WebhookPayload, "content" | "embeds" | "file" | "allowedMentions" | "components">;
+  type MessageWebhookContent = Pick<WebhookPayload, "attachments" | "content" | "embeds" | "file" | "allowedMentions" | "components">;
```
DonovanDMC: `embed` needs to be included in this
```suggestion
  type MessageWebhookContent = Pick<WebhookPayload, "attachments" | "content" | "embed" | "embeds" | "file" | "allowedMentions" | "components">;
```
This might be better suited as an `Omit`
This also could do with being renamed to its actual use, editing (something like `WebhookPayloadEdit` or `WebhookEditPayload` etc)

DonovanDMC (changes requested):

lib/Client.js
```diff
@@ -1667,6 +1674,10 @@ class Client extends EventEmitter {
     * @arg {Boolean} [content.allowedMentions.everyone] Whether or not to allow @everyone/@here.
     * @arg {Boolean | Array<String>} [content.allowedMentions.roles] Whether or not to allow all role mentions, or an array of specific role mentions to allow.
     * @arg {Boolean | Array<String>} [content.allowedMentions.users] Whether or not to allow all user mentions, or an array of specific user mentions to allow.
+    * @arg {Array<Object>} [content.attachments] Attachment objects to retain and/or edit the name and/or description for currently attached files and add the filename and description for new files
+    * @arg {String} [content.attachments[].description] The file description
+    * @arg {String} [content.attachments[].filename] The file's name. This is not needed if you are attaching a new file
+    * @arg {Number} content.attachments[].id The file's ID. If you are attaching a new file, this would be the index ID of the attaching file
```
```suggestion
    * @arg {Number | String} content.attachments[].id The file's ID. If you are attaching a new file, this would be the index ID of the attaching file
```

lib/Client.js
```diff
@@ -1958,13 +1969,18 @@ class Client extends EventEmitter {
     * @arg {Boolean} [options.allowedMentions.repliedUser] Whether or not to mention the author of the message being replied to.
     * @arg {Boolean | Array<String>} [options.allowedMentions.roles] Whether or not to allow all role mentions, or an array of specific role mentions to allow.
     * @arg {Boolean | Array<String>} [options.allowedMentions.users] Whether or not to allow all user mentions, or an array of specific user mentions to allow.
+    * @arg {Boolean} [options.allowedMentions.repliedUser] Whether or not to mention the author of the message being replied to.
+    * @arg {Array<Object>} [options.attachments] Attachment objects to retain and/or edit the name and/or description for currently attached files and add the filename and description for new files
+    * @arg {String} [options.attachments[].description] The file description
+    * @arg {String} [options.attachments[].filename] The file's name. This is not needed if you are attaching a new file
+    * @arg {Number} options.attachments[].id The file's ID. If you are attaching a new file, this would be the index ID of the attaching file
```
```suggestion
    * @arg {Number | String} options.attachments[].id The file's ID. If you are attaching a new file, this would be the index ID of the attaching file
```

TheEssem:
Seems to break sending an attachment alongside text content when using `CommandInteraction#createMessage`, causing Discord to return a 50035 error with the content `Only dictionaries may be used in a ModelType`.

DonovanDMC:
Seeing as this has a lot of ongoing issues it would be nice for them to either be fixed or get at least some kind of time frame

HeadTriXz:
Any update on this PR?

lib/Client.js
```diff
@@ -851,6 +859,9 @@ class Client extends EventEmitter {
                 this.emit("warn", "[DEPRECATED] content.messageReferenceID is deprecated. Use content.messageReference instead");
                 content.message_reference = {message_id: content.messageReferenceID};
             }
+            if(file) {
+                content = {payload_json: content};
+            }
```
HeadTriXz, line 867:
```suggestion
```

lib/Client.js
```diff
@@ -2089,6 +2113,17 @@ class Client extends EventEmitter {
...
+            content: options.content,
+            embeds: options.embeds,
+            username: options.username,
+            avatar_url: options.avatarURL,
+            tts: options.tts,
+            flags: options.flags,
+            allowed_mentions: this._formatAllowedMentions(options.allowedMentions),
+            components: options.components
+        };
+
```
HeadTriXz:
```suggestion
```

lib/Client.js
```diff
@@ -1994,13 +2013,14 @@ class Client extends EventEmitter {
...
-        if(options.embed) {
-            if(!options.embeds) {
-                options.embeds = [];
-            }
-            options.embeds.push(options.embed);
+        const file = options.file;
+        delete options.file;
+
+        if(options.embed && !options.embeds) {
+            options.embeds = [options.embed];
```
HeadTriXz:
```suggestion
        if(options.embed) {
            if(!options.embeds) {
                options.embeds = [];
            }
            options.embeds.push(options.embed);
```

lib/Client.js
```diff
@@ -1994,13 +2013,14 @@ class Client extends EventEmitter {
...
         }
-        return this.requestHandler.request("PATCH", Endpoints.WEBHOOK_MESSAGE(webhookID, token, messageID), false, options, options.file).then((response) => new Message(response, this));
+
+        return this.requestHandler.request("PATCH", Endpoints.WEBHOOK_MESSAGE(webhookID, token, messageID), false, file ? {payload_json: options} : options, file).then((response) => new Message(response, this));
```
HeadTriXz, line 2031:
```suggestion
        return this.requestHandler.request("PATCH", Endpoints.WEBHOOK_MESSAGE(webhookID, token, messageID), false, options, options.file).then((response) => new Message(response, this));
```

lib/Client.js
```diff
@@ -2102,16 +2137,7 @@ class Client extends EventEmitter {
...
-            content: options.content,
-            embeds: options.embeds,
-            username: options.username,
-            avatar_url: options.avatarURL,
-            tts: options.tts,
-            flags: options.flags,
-            allowed_mentions: this._formatAllowedMentions(options.allowedMentions),
-            components: options.components
-        }, options.file).then((response) => options.wait ? new Message(response, this) : undefined);
+        return this.requestHandler.request("POST", Endpoints.WEBHOOK_TOKEN(webhookID, token) + (qs ? "?" + qs : ""), !!options.auth, options.file ? {payload_json: data} : data, options.file).then((response) => options.wait ? new Message(response, this) : undefined);
```
HeadTriXz:
```suggestion
        return this.requestHandler.request("POST", Endpoints.WEBHOOK_TOKEN(webhookID, token) + (qs ? "?" + qs : ""), !!options.auth, {
            content: options.content,
            embeds: options.embeds,
            username: options.username,
            avatar_url: options.avatarURL,
            tts: options.tts,
            flags: options.flags,
            allowed_mentions: this._formatAllowedMentions(options.allowedMentions),
            components: options.components
        }, options.file).then((response) => options.wait ? new Message(response, this) : undefined);
```

HeadTriXz (changes requested):
Revert broken changes (thanks GitHub)

bsian03:
> Seems to break sending an attachment alongside text content when using `CommandInteraction#createMessage`, causing Discord to return a 50035 error with the content `Only dictionaries may be used in a ModelType`.

This should be fixed from my testing, I'm getting @beanjo55 to double check as well

PR #1287, opened 2021-10-21
fix(rest): combine routes for GET /guilds/:id/channels

bsian03:
Confirmed by Zeyla that /guilds/:id/channels share a single rate limit bucket regardless of the major param (differing guild IDs)

### DIFF
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -404,7 +404,7 @@ class RequestHandler {
                 method += "_OLD";
             }
             route = method + route;
-        } else if(method === "GET" && /\/guilds\/\d+\/channels$/.test(route)) {
+        } else if(method === "GET" && /\/guilds\/[0-9]+\/channels$/.test(route)) {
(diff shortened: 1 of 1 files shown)


lib/rest/RequestHandler.js
```diff
@@ -404,6 +404,8 @@ class RequestHandler {
                 method += "_OLD";
             }
             route = method + route;
+        } else if(method === "GET" && /\/guilds\/\d+\/channels$/.test(route)) {
```
curtisf, line 407: constructing this regex for every request could be a problem

curtisf: Constructing this regex over and over could be problematic

curtisf (dismissed):

lib/rest/RequestHandler.js
abalabahaha: Apparently faster than match() on a regex constant on Node 16

abalabahaha (approved):

PR #1289, opened 2021-10-23
Fix GUILD_CREATE event

reinacchi:
`guildAvailable` should emit before the client turns ready


bsian03:
This is intentional, the guild doesn't become unavailable then available again when a bot logs onto the gateway and as such would be a misleading report

reinacchi:
The issue is that guildAvailable event doesn't get fired at all although it was called.

bsian03:
In what case would it do that? The logic seems fine

reinacchi:
tbh I'm not that really sure yet

PR #1290, opened 2021-10-28
Prevent heartbeat during identify

bsian03:
Discord confirmed that you should not be heartbeating in between both identify/resume and ready/resumed events respectively


curtisf (approved):
LGTM

PR #1292, opened 2021-11-01
Interactions MinMax

JustCat80:
Interactions min and max value options for Int and Number

Also fixed some typings, removes required for Sub command things, and adds focused

Adds AnyInteraction

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2577,6 +2577,8 @@ declare namespace Eris {
   }
 
   //Interaction
+  type AnyInteraction = PingInteraction | CommandInteraction | ComponentInteraction | AutocompleteInteraction;
+
   export class Interaction extends Base {
     acknowledged: boolean;
     applicationID: string;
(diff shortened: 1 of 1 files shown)


DonovanDMC:
Suggestion for an `AnyInteraction` type that's a union of all of the interaction types (excluding unknown)
useful for types

JustCat80:
so just
`type AnyInteraction: PingInteraction | CommandInteraction | ComponentInteraction | AutocompleteInteraction`
?

DonovanDMC:
> so just `type AnyInteraction: PingInteraction | CommandInteraction | ComponentInteraction | AutocompleteInteraction` ?

type * = but yes

DonovanDMC:
I'm also pretty sure the `data.type` of AutocompleteInteraction should only be 1

DonovanDMC:
since Message/User Apps don't take arguments, they can't use autocomplete, so it should be impossible for them to be there

DonovanDMC:
pretty sure the top level `data.options` will always be present for autocomplete interactions since we need to know what field we're responding to

abalabahaha (approved):
@DonovanDMC says it's fine

PR #1293, opened 2021-11-06
premiumProgressBarEnabled & nullable maxPresences

DonovanDMC:
There's also some `hub_type` I'm seeing but there's no documentation on it.

`premium_progress_bar_enabled` also isn't on the docs but it's easy enough to discern what it is.

And for nullable `maxPresences`:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2305,7 +2305,7 @@ declare namespace Eris {
     joinedAt: number;
     large: boolean;
     maxMembers: number;
-    maxPresences: number;
+    maxPresences: number | null;
     maxVideoChannelUsers?: number;
     memberCount: number;
     members: Collection<Member>;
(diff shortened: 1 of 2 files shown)


bsian03:
What's the optional case for max_presences?

bsian03:
Also discord/discord-api-docs#3757 might be useful

DonovanDMC:
> What's the optional case for max_presences?

I'm not seeing the actual conditions because I've seen both undefined and null in multiple of my servers
null in one with ~500 members
undefined in one with ~50 & ~500

bsian03:
Can you mark that as optional as well? Looks good apart from that

DonovanDMC:
> Can you mark that as optional as well? Looks good apart from that

meant to do that earlier and just never got around to it, done it now

bsian03 (approved):

PR #1295, opened 2021-11-11
Fix CommandInteraction and rawWS deep reference editing

Dramex:
While i hate this but it's the only way to fix data.resolved being overwrited in rawWS. 
this is the style of code that has been followed in all other events.

### DIFF
--- a/lib/structures/CommandInteraction.js
+++ b/lib/structures/CommandInteraction.js
@@ -52,7 +52,7 @@ class CommandInteraction extends Interaction {
             this.data.options = {
                 name: info.data.options.name,
                 type: info.data.options.type
-            }
+            };
         }
 
         if(info.data.resolved !== undefined) {
--- a/lib/structures/CommandInteraction.js
+++ b/lib/structures/CommandInteraction.js
@@ -41,9 +41,22 @@ class CommandInteraction extends Interaction {
             id: info.channel_id
         };
 
-        this.data = info.data;
+        this.data = {
+            id: info.data.id,
+            name: info.data.name,
+            type: info.data.type,
+            target_id: info.data.target_id,
+        };
+
+        if(info.data.options) {
+            this.data.options = {
+                name: info.data.options.name,
+                type: info.data.options.type
+            }
+        }
 
         if(info.data.resolved !== undefined) {
+            this.data.resolved = {};
             //Users
             if(info.data.resolved.users !== undefined) {
                 const usermap = new Collection(User);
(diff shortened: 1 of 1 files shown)


JustCat80:
You are missing some of info.data.options, https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-interaction-data-option-structure

DonovanDMC:
considering how often interactions are changing, we shouldn't be manually doing options like that

DonovanDMC:
if you want to avoid references, cloning the object would be 1000x easier
something like 
```
Object.assign(Object.create(null), object)
```
is easy to do, though I don't know the performance impacts (and deep links can still exist)

Dramex:
> considering how often interactions are changing, we shouldn't be manually doing options like that

what do you suggest to fix the overwrite

JustCat80:
> > considering how often interactions are changing, we shouldn't be manually doing options like that
> 
> what do you suggest to fix the overwrite

in case you missed it
https://github.com/abalabahaha/eris/pull/1295#issuecomment-966563904

Dramex:
> if you want to avoid references, cloning the object would be 1000x easier something like
> 
> ```
> Object.assign(Object.create(null), object)
> ```
> 
> is easy to do, though I don't know the performance impacts (and deep links can still exist)

i just followed other events. 
i'll leave this decision to @Catboi8, tell me what you think and I'll do another PR for what you purpose to clean the commits by this PR.

DonovanDMC:
Something along the lines of object assigning and null prototypes could easily solve it, but once again, someone else needs to weigh in on the performance impacts (this is messy, but works, and doesn't hardcode anything but resolved)
```
this.data = Object.assign(
	Object.create(null), info.data, {
		resolved: Object.assign(Object.create(null), info.data.resolved)
	}
);
```

DonovanDMC:
hardcoding properties is fine for other events because they don't change often
interactions are actively changing

they also aren't nested properties, they're top level properties with different names than what we get so they can't do regular assignment like this

JustCat80:
Personally that's what I would do atm atleast until interactions is more stable, tho yea I agree the performance impacts should probably be tested

Dramex:
A better solution is to keep this as a known issue until interaction become more stable 😝

DonovanDMC:
that will likely be 5 months down the line

JustCat80:
If you really want to do this way then probably just mark the pr as a draft and wait

bsian03 (changes requested):
Can just JSON clone info.data

JustCat80 (approved):

bsian03 (approved):
There is a more efficient way of cloning as mentioned in the server but we'll use this for now

PR #1279, opened 2021-10-03
Discord API v9

bsian03:
Recreated from #1216

## Breaking
- Guild member joinedAt is nullable
- Guild#region has been removed, replaced by GuildChannel#rtcRegion

## Additions
- Add thread channel type constants
- Add HAS_THREAD message flag
- Add thread related message types
- Add thread channel permissions
- Add thread channel URLs
- NewsThreadChannel, PrivateThreadChannel, PublicThreadChannel classes extending ThreadChannel
- Client methods
- ThreadMember class
- `threads` property in Guild
- Thread gateway events
- Threads support in Audit Logs
- Support adding/editing attachments on editMessage
- Stage Instance class and relevant methods
- Stage Instance caching
- Stage Instance gateway events
- Stage Instance support in Audit Logs
- Message#applicationID
- TextChannel#defaultAutoArchiveDuration
- Invite#stageInstance
- Base.getDiscordEpoch

## Changes
- Gateway/REST versions 8 => 9
- Minor documentation updates
- Client#getChannel checks for stage channels as well
- Channel.from now supports Thread channels
- Option to send webhook message to a thread channel (querystring logic has been changed)
- Channel#permissionsOf case for thread channel (incorrect commit message)
- Message#guildID/member/webhookID case for thread channel
- Guild#shard calculation now has fallbacks
- Base.getCreatedAt now uses Base.getDiscordEpoch
- Other misc

Note for self: UTD docs discord/discord-api-docs#3672 tree 0d954e67b4b32703bf895d8d9c1d8853184652d1

PR #1275, opened 2021-09-06
Guild Scheduled Events

Loliticos:
Discord is implementing `Scheduled Guild Events` (discord/discord-api-docs#3586). This pull request is a start in order to implement this new feature.

## New Events

- `GUILD_SCHEDULED_EVENT_CREATE` (guildScheduledEventCreate)
- `GUILD_SCHEDULED_EVENT_UPDATE` (guildScheduledEventUpdate)
- `GUILD_SCHEDULED_EVENT_DELETE` (guildScheduledEventDelete)
- `GUILD_SCHEDULED_EVENT_USER_ADD` (guildScheduledEventUserAdd)
- `GUILD_SCHEDULED_EVENT_USER_REMOVE`(guildScheduledEventUserRemove)

## Additions
- `GuildEvent`
- `Client#guildEventMap`
- `Client#createGuildEvent()`
- `Client#deleteGuildEvent()`
- `Client#editGuildEvent()`
- `Client#getRESTGuildEvent()`
- `Guild#events`

## TODO

- [x] Document new Client, Guild and GuildEvent methods
- [x] `GUILD_SCHEDULED_EVENT_USER_ADD` (Gateway Event)
- [x] `GUILD_SCHEDULED_EVENT_USER_REMOVE` (Gateway Event)

PR #1271, opened 2021-08-29
Add, update, reorganize Constants

DonovanDMC:
This converts several unions into constants, and converts several hardcoded numbers and such into constant references (haven't searched through the js), as well as introduces several parity fixes for staying as close to Discord as reasonably possible.

### Added Constants
* [`ConnectionVisibilityTypes`](https://discord.com/developers/docs/resources/user#connection-object-visibility-types)
* [`DefaultMessageNotificationLevels`](https://discord.com/developers/docs/resources/guild#guild-object-default-message-notification-level)
* [`ExplicitContentFilterLevels`](https://discord.com/developers/docs/resources/guild#guild-object-explicit-content-filter-level)
* [`GuildIntegrationExpireBehavior`](https://discord.com/developers/docs/resources/guild#integration-object-integration-expire-behaviors)
* [`GuildIntegrationTypes`](https://discord.com/developers/docs/resources/guild#integration-object-integration-structure)
* [`GuildNSFWLevels`](https://discord.com/developers/docs/resources/guild#guild-object-guild-nsfw-level)
* [`InviteTargetTypes`](https://discord.com/developers/docs/resources/invite#invite-object-invite-target-types)
* [`MFALevels`](https://discord.com/developers/docs/resources/guild#guild-object-mfa-level)
* [`MessageActivityTypes`](https://discord.com/developers/docs/resources/channel#message-object-message-activity-types)
* [`PermissionOverwriteTypes`](https://discord.com/developers/docs/resources/channel#overwrite-object)
* [`PremiumTiers`](https://discord.com/developers/docs/resources/guild#guild-object-premium-tier)
* [`PremiumTypes`](https://discord.com/developers/docs/resources/user#user-object-premium-types)
* [`StageInstancePrivacyLevel`](https://discord.com/developers/docs/resources/stage-instance#stage-instance-object-privacy-level)
* [`SystemChannelFlags`](https://discord.com/developers/docs/resources/guild#guild-object-system-channel-flags)
* [`VerificationLevels`](https://discord.com/developers/docs/resources/guild#guild-object-verification-level)
* [`VideoQualityModes`](https://discord.com/developers/docs/resources/channel#channel-object-video-quality-modes)
* `ThreadMemberFlags` ([not documented](https://butts-are.cool/notepad__11-14-2021_02-14-54.png))
* [`WebhookTypes`](https://discord.com/developers/docs/resources/webhook#webhook-object-webhook-types)

[...]

PR #1296, opened 2021-11-13
New Permissions

DonovanDMC:
This pull request adds in all of the permissions we're currently missing:
* `manageEvents` (1 << 33)
* `manageThreads` (1 << 34)
* `createPublicThreads` (1 << 35)
* `createPrivateThreads` (1 << 36)
* `sendMessagesInThreads` (1 << 38)
* `startEmbeddedActivities` (1 << 39)
It also aligns all of the permissions in the types like how they are in the js file.

The majority of these are already included in other pull requests (33, #1275) (34-38, #1279), but we've been missing them for quite a while so I feel it's time to add them in.

`manageEvents`  isn't in the docs yet, but making a role with only `manageEvents` produces `8589934592` (1 << 33), so it can be safely assumed.


bsian03:
33 is in #1275

DonovanDMC:
Yeah I saw that after I said it, pretty sure 39 still isn't anywhere

DonovanDMC:
Folding this into #1271

PR #1301, opened 2021-11-16
fix(structs): fix incorrect thread method references

ZixeSea:
There were 2 broken methods there instead of **this._client** it was still using **this.client** resulting in a "Cannot read property 'methods' of undefined" error


bsian03 (dismissed):

bsian03:
Man... find and replace in js do be broke

bsian03 (approved):
I should really look at these PRs better...

PR #1302, opened 2021-11-16
fix(voice): properly ignore non-voice packets

davidffa:
- Fixed opus decode error when receiving voice from browser clients.
Error:
image

- Fixed constant "Failed to decrypt received packet" warnings by ignoring unknown payload types (2nd byte different from 0x78).
- Fixed the error below when joining voice channels by using the internal method to send initial silence frame instead of the public one.
image


abalabahaha (approved):
For the third error, added a `frameSize` option in 7c39e1a

PR #1308, opened 2021-11-25
remove most self bot features

numselli:
I have not removed any of the groupChannels for frear of breaking stuff. 
all changes are untested


ray-1337:
w  h  y

numselli:
To remove bloat so my bot (and others) use less ram

eritbh:
See also #489, the previous attempt.

> so my bot (and others) use less ram

Removing these features shouldn't noticeably impact resource usage. Bot accounts should never receive API events that populate the relationships cache or the other client properties you've removed, so I'm pretty sure the only memory savings this would net you would be the size of a couple empty `Map`s.

bsian03:
Since you unforked, and because Discord confirmed that we need to remove all user account APIs, I'm going to close and open a new PR to remove everything

PR #1310, opened 2021-12-01
Guild Scheduled Events Intent

DonovanDMC:
We somehow forgot the guild scheduled events intent
[](https://discord.com/developers/docs/topics/gateway#list-of-intents)


abalabahaha:
This should be part of #1275

PR #1311, opened 2021-12-07
Fix typing for `Message.createThreadWithMessage`

frobinsonj:
Function should not have `messageID` parameter.


Linker-123:
You only changed the typings, the function still accepts a messageID parameter. And according to the [discord docs](https://discord.com/developers/docs/resources/channel#start-thread-with-message) this endpoint requires the message-id.

bsian03 (approved):

PR #1313, opened 2021-12-11
fix(Member): User property reference

ghost:
The user's values do not get updated in member objects. This pr fixes this issue.

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -57,7 +57,11 @@ class Member extends Base {
                 throw new Error("User associated with Member not found: " + data.id);
             }
         } else if(data.user) {
-            this.user = new User(data.user, client);
+            if(!client) {
+                this.user = new User(data.user, client);
+            } else {
+                this.user = client.users.add(data.user, client);
+            }
         } else {
             this.user = null;
         }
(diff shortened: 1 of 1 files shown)


lib/structures/Member.js
```diff
@@ -57,7 +57,11 @@ class Member extends Base {
                 throw new Error("User associated with Member not found: " + data.id);
             }
         } else if(data.user) {
-            this.user = new User(data.user, client);
+            if(!client) {
+                this.user = new User(data.user, client);
```
bsian03: Useless client

lib/structures/Member.js
```diff
@@ -57,7 +57,11 @@ class Member extends Base {
...
+            } else {
+                this.user = client.users.add(data.user, client);
```
bsian03: Can probably update instead of add

bsian03 (changes requested):

lib/structures/Member.js
ghost: Oh okay, i just thought that the client property may not always be passed.

lib/structures/Member.js
```diff
@@ -57,7 +57,7 @@ class Member extends Base {
                 throw new Error("User associated with Member not found: " + data.id);
             }
         } else if(data.user) {
-            this.user = new User(data.user, client);
+            this.user = client.users.update(data.user, client);
```
bsian03: Client is still not guaranteed to exist at this location

bsian03 (approved):

PR #1315, opened 2021-12-14
Use the guild_id field from Discord payload.

GweeKwee:
Currently when we uncache a guild text channel, the Message instance returns always null for the member field.
We can use the "message.guild_id" coming from the Discord payload to search for the guild and member.

### DIFF
--- a/lib/structures/Message.js
+++ b/lib/structures/Message.js
@@ -120,21 +120,22 @@ class Message extends Base {
             this.interaction = null;
         }
 
-        if(this.channel.guild) {
+        if(data.guild_id) {
+            const guild = this._client.guilds.get(data.guild_id);
             if(data.member) {
                 data.member.id = this.author.id;
                 if(data.author) {
                     data.member.user = data.author;
                 }
-                this.member = this.channel.guild.members.update(data.member, this.channel.guild);
-            } else if(this.channel.guild.members.has(this.author.id)) {
-                this.member = this.channel.guild.members.get(this.author.id);
+                this.member = guild.members.update(data.member, guild);
+            } else if(guild.members.has(this.author.id)) {
+                this.member = guild.members.get(this.author.id);
             } else {
                 this.member = null;
             }
 
             if(!this.guildID) {
-                this.guildID = this.channel.guild.id;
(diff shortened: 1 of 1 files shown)


lib/structures/Message.js
```diff
@@ -120,21 +120,22 @@ class Message extends Base {
             this.interaction = null;
         }
 
-        if(this.channel.guild) {
+        if(data.guild_id) {
+            const guild = this._client.guilds.get(data.guild_id);
```
bsian03: This will cause uncached guilds to throw errors

lib/structures/Message.js
```diff
@@ -120,21 +120,22 @@ class Message extends Base {
...
+                this.member = guild.members.update(data.member, guild);
+            } else if(guild.members.has(this.author.id)) {
+                this.member = guild.members.get(this.author.id);
             } else {
                 this.member = null;
             }
 
             if(!this.guildID) {
-                this.guildID = this.channel.guild.id;
+                this.guildID = data.guild_id;
```
bsian03, line 140: This is already covered in L66

lib/structures/Message.js
GweeKwee: Oh wopsie, added a check for uncache guild

lib/structures/Message.js
```diff
@@ -120,21 +120,24 @@ class Message extends Base {
...
         }
 
-        if(this.channel.guild) {
-            if(data.member) {
-                data.member.id = this.author.id;
-                if(data.author) {
-                    data.member.user = data.author;
+        if(data.guild_id) {
+            if(this._client.guilds.has(data.guild_id)) {
+                const guild = this._client.guilds.get(data.guild_id);
```
bsian03, line 125: This is redundant

lib/structures/Message.js
bsian03: To clarify, this was done in the rare case Discord screws up their payload.. which has happened before. The reliance on channel being cached with a missing guild_id meant that guildID would always be present if either the channel was cached and belongs in a guild, or guild_id is missing

PR #1318, opened 2021-12-21
fix(Member): Parse Member#premiumSince to timestamp

### LINKED ISSUE #1316: Member.premiumsince not converted to epoch timestamp
Both the JSdocs and typescript type definition declare that the `Member#premiumsince` property is a numbers, but [discord gives an ISO timestamp](https://discord.com/developers/docs/resources/guild#guild-member-object-guild-member-structure) and the transforms that the Member class does doesnt match the docs.

Member should show the timestamp as a number

bsian03:
Fixes #1316

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -81,7 +81,7 @@ class Member extends Base {
             this.activities = data.activities;
         }
         if(data.premium_since !== undefined) {
-            this.premiumSince = Date.parse(data.premium_since);
+            this.premiumSince = data.premium_since === null ? null : Date.parse(data.premium_since);
         }
         if(data.hasOwnProperty("mute") && this.guild) {
             const state = this.guild.voiceStates.get(this.id);
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -34,7 +34,7 @@ const VoiceState = require("./VoiceState");
 * @prop {Boolean?} pending Whether the member has passed the guild's Membership Screening requirements
 * @prop {Permission} permission [DEPRECATED] The guild-wide permissions of the member. Use Member#permissions instead
 * @prop {Permission} permissions The guild-wide permissions of the member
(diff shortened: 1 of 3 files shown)


DonovanDMC:
`premium_since` is defined as both optional & nullable in the docs, optional is handled, nullable is not -- this results in NaN

bsian03:
What if you don't want a number smh

JustCat80:
in that case perfect because it is not, luck with your project 👍

danny-may:
From what im reading in the docs, cant the value be `null`, `undefined` or a string? Shouldnt the type guard be a nullish check or a `typeof x === 'string'` check? Strict null equality would break I believe

bsian03:
But.. we already have an undefined check

lib/structures/Member.js
```diff
@@ -81,7 +81,7 @@ class Member extends Base {
             this.activities = data.activities;
         }
         if(data.premium_since !== undefined) {
-            this.premiumSince = data.premium_since;
+            this.premiumSince = data.premium_since === null ? null : Date.parse(data.premium_since);
```
danny-may, line 84:
```suggestion
            this.premiumSince = data.premium_since == null ? null : Date.parse(data.premium_since);
```

danny-may:
Ah, on the lines above, I missed that. Ignore me :)

DonovanDMC (dismissed):

DonovanDMC (changes requested):
typings (nullable & optional on both Member & OldMember)
https://github.com/abalabahaha/eris/blob/b2a90ea390bf3f74cec6772cc29956d31d8a3391/index.d.ts#L654
https://github.com/abalabahaha/eris/blob/b2a90ea390bf3f74cec6772cc29956d31d8a3391/index.d.ts#L3046

DonovanDMC (approved):

PR #1306, opened 2021-11-18
m.presence can be null in THREAD_MEMBERS_UPDATE

ZixeSea:
In the THREAD_MEMBERS_UPDATE event, the user is updated what also includes a presence update (m.presence). The m.presence object can be null resulting in a "Cannot read property 'id' of null" error.

PR #1324, opened 2022-01-15
update member on THREAD_MEMBERS_UPDATE

GweeKwee:
Updating the guild member on THREAD_MEMBERS_UPDATE without passing a guild param will throw a "Missing client in constructor" in user structure for #1306 and #1313 

This happens when an offline member is mentioned in a thread.


rilysh (dismissed):
absolutely

bsian03 (dismissed):

bsian03 (approved):

PR #1325, opened 2022-01-17
fix(slash): Snake case defaultPermissions to payload

bsian03:


coolcalcacol:
I thought this was in another PR @bsian03?

DonovanDMC:
I left it as snake case as most things around interactions is already snake case

bsian03:
Did the PR with the intention that it would be merged first - Donovan's PR would also technically cause a breaking change from the docs perspective

DonovanDMC:
I'll just take it out of my pr, since this will almost certainly make it in first, and that is true

See887777 (approved):

DonovanDMC (approved):

curtisf (approved):

PR #1327, opened 2022-01-19
Add the Message#reply method

Kxrozz-jpg:
It's time consuming to put that much code to send a message, so putting this in would make doing all of that commendable.

In addition to not having to use `Message#channel` so that the bot knows which channel to send the message to, it has inline replies using the id of the `Message` class, so we save lines of code. But in this code you don't have a good check of the properties that the `options` object should carry, but that's easy.
```js
/**
 *
 * @param {Object} options
 * @returns {Promise<Message>}
 */
reply(options = {}) {
  if (!this.channel) return Promise.reject(new Error(":)"));
  options.messageReference = {
    messageID: this.id,
    failIfNotExists: false,
  };
  return this._client.createMessage.call(this._client, this.channel.id, options);
}
   ```


eritbh:
It doesn't look like you've PRed the correct branch; these commits are from someone else's fork, and they don't add the thing you're talking about.

Kxrozz-jpg:
> It doesn't look like you've PRed the correct branch; these commits are from someone else's fork, and they don't add the thing you're talking about.

Sorry, I confused the repository. I didn't realize until now

ray-1337:
wow non english PRs, oui baguette

PR #1328, opened 2022-01-20
feat(editChannelPositions): add support for endpoint

carloslimaborges:
Implements this part of Discord API: https://discord.com/developers/docs/resources/guild#modify-guild-channel-positions

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2288,7 +2288,7 @@ declare namespace Eris {
     editCommand(commandID: string, command: ApplicationCommandStructure): Promise<ApplicationCommand>;
     editCommandPermissions(guildID: string, commandID: string, permissions: ApplicationCommandPermissions[]): Promise<GuildApplicationCommandPermissions>;
     editGuild(guildID: string, options: GuildOptions, reason?: string): Promise<Guild>;
-    editGuildChannelPositions(guildID: string, channelPositions: ChannelPosition[]): Promise<void>;
+    editGuildChannelPositions(guildID: string, channelPositions: ChannelPosition[], reason?: string): Promise<void>;
     editGuildCommand(guildID: string, commandID: string, command: ApplicationCommandStructure): Promise<ApplicationCommand>;
     editGuildDiscovery(guildID: string, options?: DiscoveryOptions): Promise<DiscoveryMetadata>;
     editGuildEmoji(
@@ -2718,7 +2718,7 @@ declare namespace Eris {
     dynamicSplashURL(format?: ImageFormat, size?: number): string | null;
     edit(options: GuildOptions, reason?: string): Promise<Guild>;
     editCommand(commandID: string, command: ApplicationCommandStructure): Promise<ApplicationCommand>;
-    editChannelPositions(channelPositions: ChannelPosition[]): Promise<void>;
+    editChannelPositions(channelPositions: ChannelPosition[], reason?: string): Promise<void>;
     editCommandPermissions(permissions: ApplicationCommandPermissions[]): Promise<GuildApplicationCommandPermissions[]>;
     editDiscovery(options?: DiscoveryOptions): Promise<DiscoveryMetadata>;
     editEmoji(emojiID: string, options: { name: string; roles?: string[] }, reason?: string): Promise<Emoji>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -281,6 +281,12 @@ declare namespace Eris {
     channel_id: string;
     webhook_id: string;
   }
+  interface ChannelPosition {
+    id: string;
+    position: number;
+    lockPermissions?: boolean;
+    parentID?: string;
+  }
   interface CreateChannelOptions {
     bitrate?: number;
     nsfw?: boolean;
@@ -840,12 +846,6 @@ declare namespace Eris {
   }
 
   // Guild
-  interface GuildChannelPosition {
-    id: string;
-    position: number;
-    lockPermissions?: boolean;
-    parentID?: string;
-  }
   interface CreateGuildOptions {
     afkChannelID?: string;
     afkTimeout?: number;
@@ -2288,7 +2288,7 @@ declare namespace Eris {
     editCommand(commandID: string, command: ApplicationCommandStructure): Promise<ApplicationCommand>;
     editCommandPermissions(guildID: string, commandID: string, permissions: ApplicationCommandPermissions[]): Promise<GuildApplicationCommandPermissions>;
     editGuild(guildID: string, options: GuildOptions, reason?: string): Promise<Guild>;
-    editGuildChannelPositions(guildID: string, guildChannelPositions: GuildChannelPosition[]): Promise<void>;
+    editGuildChannelPositions(guildID: string, channelPositions: ChannelPosition[]): Promise<void>;
     editGuildCommand(guildID: string, commandID: string, command: ApplicationCommandStructure): Promise<ApplicationCommand>;
     editGuildDiscovery(guildID: string, options?: DiscoveryOptions): Promise<DiscoveryMetadata>;
     editGuildEmoji(
@@ -2718,7 +2718,7 @@ declare namespace Eris {
     dynamicSplashURL(format?: ImageFormat, size?: number): string | null;
     edit(options: GuildOptions, reason?: string): Promise<Guild>;
     editCommand(commandID: string, command: ApplicationCommandStructure): Promise<ApplicationCommand>;
-    editChannelPositions(guildChannelPositions: GuildChannelPosition[]): Promise<void>;
+    editChannelPositions(channelPositions: ChannelPosition[]): Promise<void>;
     editCommandPermissions(permissions: ApplicationCommandPermissions[]): Promise<GuildApplicationCommandPermissions[]>;
     editDiscovery(options?: DiscoveryOptions): Promise<DiscoveryMetadata>;
     editEmoji(emojiID: string, options: { name: string; roles?: string[] }, reason?: string): Promise<Emoji>;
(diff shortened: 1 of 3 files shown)


lib/Client.js
```diff
@@ -1328,6 +1328,30 @@ class Client extends EventEmitter {
...
     }
 
+    /**
+    * Edit guild channels' positions. Note that channel position numbers are lowest on top and highest at the bottom.
+    * @arg {String} guildID The ID of the guild
+    * @arg {Array<Object>} channelPositions An array of [channel positions](https://discord.com/developers/docs/resources/guild#modify-guild-channel-positions)
+    * @arg {String} channelPositions[].id The ID of the channel
+    * @arg {Number} channelPositions[].position The new position of the channel
+    * @arg {Boolean} [channelPositions[].lockPermissions] Whether to sync the permissions with the new parent if moving to a new category
+    * @arg {String} [channelPositions[].parentID] The new parent ID (category channel) for the channel that is moved (only one channel can change category per request)
```
eritbh: > (only one channel can change category per request)

I don't see anything about this in Discord's docs; is there a source for this info I'm missing? Am I blind?

eritbh (commented):
Would probably make sense to add a corresponding helper on `Guild`, and types for the new method(s) should be added to `index.d.ts` as well. The added code code looks fine to me though.

lib/Client.js
carloslimaborges: That was based on experience because I tried to change more than one channel's category and got the error message stating this.

carloslimaborges:
> Would probably make sense to add a corresponding helper on `Guild`, and types for the new method(s) should be added to `index.d.ts` as well. The added code code looks fine to me though.

Done 😃

index.d.ts
```diff
@@ -840,6 +840,12 @@ declare namespace Eris {
   }
 
   // Guild
+  interface GuildChannelPosition {
+    id: string;
+    position: number;
+    lockPermissions?: boolean;
+    parentID?: string;
+  }
```
bsian03: This can probably be called ChannelPosition btw, and should go under the Channel category - please also put this in the relevant alphabetical location. I know Discord calls it GuildChannelPosition, but it's obvious that it belongs in a guild

lib/Client.js
```diff
@@ -1420,6 +1420,30 @@ class Client extends EventEmitter {
...
+        const guild = this.guilds.get(guildID);
+        if(!guild) {
+            return Promise.reject(new Error(`Guild ${guildID} not found`));
+        }
+        guildChannelPositions.forEach((channelPosition) => {
+            const channel = guild.channels.get(channelPosition.id);
+            if(!channel) {
+                return Promise.reject(new Error(`Channel ${channelPosition.id} not found`));
+            }
+        });
```
bsian03: This type of handling should not be here, that's up to the developer to filter out

bsian03 (changes requested):
Endpoint also accepts an audit log entry and thus should include `reason` when calling the method

lib/Client.js
```diff
@@ -1420,6 +1420,30 @@ class Client extends EventEmitter {
...
+        return this.requestHandler.request("PATCH", Endpoints.GUILD_CHANNELS(guildID), true, guildChannelPositions);
```
HeadTriXz: The `guildChannelPositions` should be formatted correctly for Discord. E.g. `lockPermissions` > `lock_permissions`. Also, the name of the parameter is quite long right now, `positions` would work better in my opinion.

HeadTriXz (changes requested):

lib/Client.js
DonovanDMC: the length of the parameter name doesn't matter, this isn't part of an object or something you have to actually specify

HeadTriXz: > the length of the parameter name doesn't matter, this isn't part of an object or something you have to actually specify

Still, looks cleaner ;)

carloslimaborges:
> Endpoint also accepts an audit log entry and thus should include `reason` when calling the method

I will need help on this one, cause the body I send is an array. And I can't add a reason to that array to send to the RequestHandler. Changing the RequestHandler for this made me a bit uncomfortable, considering this is my first big contribution.

index.d.ts
carloslimaborges: Changed it as requested

lib/Client.js
carloslimaborges: Changed this too as requested

carloslimaborges (commented):
Please check if it is ok after the changes

eritbh:
> the body I send is an array. And I can't add a reason to that array to send to the `RequestHandler`.

Maybe a little unorthodox, but manually assigning `body.reason` after the initial assignment should be fine. It doesn't look like the request handler does anything special other than reading that property and then deleting it, so I think this is a case where adding an extra property to an array is safe.

It might be better to restructure the request handler's `request` method to take an explicit argument for the `X-Audit-Log` header, or maybe an object argument for setting arbitrary headers, but that feels like a bigger change that would require updates to a bunch of other code at the same time. I'm not a maintainer, but my take is that sort of change should probably be a separate PR.

carloslimaborges:
> > the body I send is an array. And I can't add a reason to that array to send to the `RequestHandler`.
> 
> Maybe a little unorthodox, but manually assigning `body.reason` after the initial assignment should be fine. It doesn't look like the request handler does anything special other than reading that property and then deleting it, so I think this is a case where adding an extra property to an array is safe.
> 
> It might be better to restructure the request handler's `request` method to take an explicit argument for the `X-Audit-Log` header, or maybe an object argument for setting arbitrary headers, but that feels like a bigger change that would require updates to a bunch of other code at the same time. I'm not a maintainer, but my take is that sort of change should probably be a separate PR.

It did work. So I finished the change.

carloslimaborges:
> Endpoint also accepts an audit log entry and thus should include `reason` when calling the method

I resolved this, how do I mark it as resolved at github?

lib/Client.js
```diff
@@ -1423,6 +1423,30 @@ class Client extends EventEmitter {
         }).then((guild) => new Guild(guild, this));
     }
 
+    /**
+    * Edit guild channels' positions. Note that channel position numbers are lowest on top and highest at the bottom.
```
```suggestion
    * Edit guild channel positions. Note that channel position numbers are lowest on top and highest at the bottom.
```
grammar police alert

bsian03: also channel permissions are a bit niche, and requires in depth documentation for it to be explained properly. your wording isn't wrong, but it also isn't right. maybe we can sort this out sometime in the future

lib/Client.js
```diff
@@ -1423,6 +1423,30 @@ class Client extends EventEmitter {
...
+    * @arg {String} guildID The ID of the guild
+    * @arg {Array<Object>} channelPositions An array of [ChannelPosition] (https://discord.com/developers/docs/resources/guild#modify-guild-channel-positions)
```
```suggestion
    * @arg {Array<Object>} channelPositions An array of [ChannelPosition](https://discord.com/developers/docs/resources/guild#modify-guild-channel-positions)
```

lib/Client.js
```diff
@@ -1423,6 +1423,30 @@ class Client extends EventEmitter {
...
+    * @arg {String} ChannelPosition[].id The ID of the channel
+    * @arg {Number} ChannelPosition[].position The new position of the channel
+    * @arg {Boolean} [ChannelPosition[].lockPermissions] Whether to sync the permissions with the new parent if moving to a new category
+    * @arg {String} [ChannelPosition[].parentID] The new parent ID (category channel) for the channel that is moved (only one channel can change category per request)
```
bsian03: what's with the PascalCase and the mismatch param name

lib/structures/Guild.js
```diff
@@ -616,6 +616,20 @@ class Guild extends Base {
         return this._client.editGuild.call(this._client, this.id, options, reason);
     }
 
+    /**
+    * Edit guild channels' positions. Note that channel position numbers are lowest on top and highest at the bottom.
+    * @arg {Array<Object>} channelPositions An array of [ChannelPosition] (https://discord.com/developers/docs/resources/guild#modify-guild-channel-positions)
+    * @arg {String} ChannelPosition[].id The ID of the channel
+    * @arg {Number} ChannelPosition[].position The new position of the channel
+    * @arg {Boolean} [ChannelPosition[].lockPermissions] Whether to sync the permissions with the new parent if moving to a new category
+    * @arg {String} [ChannelPosition[].parentID] The new parent ID (category channel) for the channel that is moved (only one channel can change category per request)
```
bsian03: see above

bsian03 (dismissed):

lib/Client.js
carloslimaborges: I just copied this formatting from somewhere else in the code. How should it be?

eritbh: Should always match the name of the parameter, like so:

```suggestion
    * @arg {String} [channelPositions[].parentID] The new parent ID (category channel) for the channel that is moved (only one channel can change category per request)
```

See the `activities` parameter of [`editStatus`](https://abal.moe/Eris/docs/0.16.1/Client#method-editStatus) for another example of documenting an array of objects.

carloslimaborges: I just pushed the corrections. Hope it is ok now. Thanks for all the help and feedback

lib/Client.js
```diff
@@ -1427,6 +1427,30 @@ class Client extends EventEmitter {
         }).then((guild) => new Guild(guild, this));
     }
 
+    /**
+    * Edit guild channel positions. Note that channel position numbers are lowest on top and highest at the bottom.
```
bsian03: Channel position numbers are a bit more niche than this. They're top to bottom lowest to highest for each type of channel (cat, text, voice). For anyone else confused by this, a diagram I made a little while ago should help

image

carloslimaborges: I didn’t understand what you want me to change and what should the text be.

carloslimaborges:
Is there any change request I haven't dealt with yet? I get a little bit confused by the github interface, sorry

abalabahaha (approved):

PR #1331, opened 2022-01-27
Support file descriptions / alt-text

andromika:
This incorporates a json body for every attachment following the `files[n]` format.

### DIFF
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -160,10 +160,9 @@ class RequestHandler {
 
                                 if(body) {
                                     if(typeof body === "string"){
-                                        // eslint-disable-next-line object-shorthand
                                         body = {
                                             content: body,
-                                            attachments
+                                            attachments: attachments
                                         };
                                     } else {
                                         body.attachments = attachments;
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -131,11 +131,10 @@ class RequestHandler {
 
                             if(body) {
                                 if(typeof body === "string"){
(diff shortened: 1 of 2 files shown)


lib/rest/RequestHandler.js
```diff
@@ -110,21 +111,72 @@ class RequestHandler {
...
                             });
+
                             if(body) {
-                                data.attach("payload_json", body);
+                                if(typeof body === "string"){
+                                    // eslint-disable-next-line object-shorthand
+                                    body = {
+                                        content: body,
+                                        attachments
+                                    };
```
DonovanDMC: Instead of disabling the eslint rule, you should follow the styling
```suggestion
                                        body = {
                                            content: body,
                                            attachments: attachments
                                        };
```

lib/rest/RequestHandler.js
```diff
@@ -110,21 +111,72 @@ class RequestHandler {
...
+                                    filename: file.name
+                                }];
+
+                                if(body) {
+                                    if(typeof body === "string"){
+                                        // eslint-disable-next-line object-shorthand
+                                        body = {
+                                            content: body,
+                                            attachments
+                                        };
```
DonovanDMC:
```suggestion
                                        body = {
                                            content: body,
                                            attachments: attachments
                                        };
```

DonovanDMC (changes requested):

bsian03:
I haven't looked at it properly but isn't this #1285?

HeadTriXz:
> I haven't looked at it properly but isn't this #1285?

Yup!

PR #1338, opened 2022-02-10
Remove all user account features

bsian03:


coolcalcacol:
Ain't GUILD_SYNC a legacy user implementation as well in shard?

DonovanDMC:
you left some interfaces dangling around (`UserProfile`/`UserSettings`) and a bunch of `GroupChannel` remnants

And some very old bots can technically be in Group Channels iirc, I assume we aren't considering that as a valid enough reason to keep the around though (made a pr, bsian03#21)

index.d.ts
```diff
@@ -701,23 +701,15 @@ declare namespace Eris {
...
-    callDelete: [call: Call];
-    callRing: [call: Call];
-    callUpdate: [call: Call, oldCall: OldCall];
     channelCreate: [channel: AnyChannel];
     channelDelete: [channel: AnyChannel];
     channelPinUpdate: [channel: TextableChannel, timestamp: number, oldTimestamp: number];
-    channelRecipientAdd: [channel: GroupChannel, user: User];
-    channelRecipientRemove: [channel: GroupChannel, user: User];
     channelUpdate: [channel: AnyGuildChannel, oldChannel: OldGuildChannel | OldGuildTextChannel | OldGuildVoiceChannel]
     | [channel: GroupChannel, oldChannel: OldGroupChannel];
```
coolcalcacol, line 877: Missed the group channel stuff here

index.d.ts
```diff
@@ -2078,40 +2054,31 @@ declare namespace Eris {
...
-    userSettings: UserSettings;
     voiceConnections: VoiceConnectionManager;
     constructor(token: string, options?: ClientOptions);
-    acceptInvite(inviteID: string): Promise<Invite<"withoutCount">>;
-    addGroupRecipient(groupID: string, userID: string): Promise<void>;
     addGuildDiscoverySubcategory(guildID: string, categoryID: string, reason?: string): Promise<DiscoverySubcategoryResponse>;
     addGuildMemberRole(guildID: string, memberID: string, roleID: string, reason?: string): Promise<void>;
     addMessageReaction(channelID: string, messageID: string, reaction: string): Promise<void>;
     /** @deprecated */
     addMessageReaction(channelID: string, messageID: string, reaction: string, userID: string): Promise<void>;
```
coolcalcacol: Is this not a user feature too? If so, time to remove the UserID arg?

lib/Client.js
```diff
@@ -2620,11 +2404,10 @@ class Client extends EventEmitter {
 
     /**
     * Get data on an OAuth2 application
-    * @arg {String} [appID="@me"] The client ID of the application to get data for (user accounts only). "@me" refers to the logged in user's own application
     * @returns {Promise<Object>} The bot's application data. Refer to [the official Discord API documentation entry](https://discord.com/developers/docs/topics/oauth2#get-current-application-information) for object structure
     */
     getOAuthApplication(appID) {
```
coolcalcacol: remove appID arg here?

lib/gateway/Shard.js
```diff
@@ -1871,20 +1642,9 @@ class Shard extends EventEmitter {
...
-                } else {
-                    this.client.bot = false;
-                    this.client.userGuildSettings = {};
-                    if(packet.d.user_guild_settings) {
-                        packet.d.user_guild_settings.forEach((guildSettings) => {
-                            this.client.userGuildSettings[guildSettings.guild_id] = guildSettings;
-                        });
-                    }
-                    this.client.userSettings = packet.d.user_settings;
+                this.client.bot = true;
```
coolcalcacol: is there a case where this will be false? If not, perhaps best to remove it?

lib/gateway/Shard.js
```diff
@@ -2321,11 +1994,8 @@ class Shard extends EventEmitter {
                 this.emit("stageInstanceDelete", guild.stageInstances.remove(packet.d) || new StageInstance(packet.d, this.client));
                 break;
             }
-            case "MESSAGE_ACK": // Ignore these
-            case "GUILD_INTEGRATIONS_UPDATE":
-            case "USER_SETTINGS_UPDATE":
-            case "CHANNEL_PINS_ACK": {
-                break;
+            case "GUILD_INTEGRATIONS_UPDATE": {
+                break; // REVIEW
```
coolcalcacol: This sends a guildID on twitch/youtube integration update, there are some other integration events on the discord docs but they seem to not be sent

index.d.ts
DonovanDMC: bsian03#21

index.d.ts
DonovanDMC: is what a user feature, adding reactions? as far as `userID` is concerned, no idea why it's there, only value that would work for both bots & users is "@me" - there *might* be something in OAuth but I doubt that

GoldenAngel2:
`Client#editGuildIntegration`, `Client#syncGuildIntegration`, `Guild#editIntegration`, `Guild#syncIntegration`, `GuildIntegration#`, `GuildIntegration#sync`, `GuildIntegration#edit` can be removed as well, since those can only be used by user accounts.

bsian03:
I'm in the process of going through all of them now that I'm on my laptop

lib/gateway/Shard.js
bsian03: We ignored these for idk how long. Am asking abal for clarification (wish me luck for a reply)

bsian03: So it seems like when it was first implemented it basically did nothing, I'm going to unignore this event

lib/voice/VoiceConnection.js
```diff
@@ -200,7 +200,7 @@ class VoiceConnection extends EventEmitter {
                 this.connectionTimeout = null;
             }
             this.sendWS(VoiceOPCodes.IDENTIFY, {
-                server_id: this.id === "call" ? data.channel_id : this.id,
+                server_id: this.id === this.id,
```
TTtie: noticed a tiny oversight here:
```suggestion
                server_id: this.id,
```

DonovanDMC: I must've been on something with making the pr I did to remove half ofthis to miss something so obvious

lib/gateway/Shard.js
coolcalcacol: 👍

lib/gateway/Shard.js
```diff
@@ -560,8 +538,6 @@ class Shard extends EventEmitter {
     sendStatusUpdate() {
         this.sendWS(GatewayOPCodes.PRESENCE_UPDATE, {
             activities: this.presence.activities,
```
DonovanDMC, line 554:
```suggestion
            activities: this.presence.activities,
            afk: !!this.presence.afk,
            since: this.presence.status === "idle" ? Date.now() : null,
```

Presence updates don't work without these lines

HeadTriXz: `since` is optional.

DonovanDMC: it doesn't work without it, so no it isn't

DonovanDMC: since is *nullable*
image

DonovanDMC:
`Client#getChannel` still checks for group channels, which no longer exist

various remnants of Group Channels have been left dotted around

HeadTriXz:
> `Client#getChannel` still checks for group channels, which no longer exist
> 
> various remnants of Group Channels have been left dotted around

Is it worth it to still support group channels though?

DonovanDMC:
> > `Client#getChannel` still checks for group channels, which no longer exist
> > various remnants of Group Channels have been left dotted around
> 
> Is it worth it to still support group channels though?

bsian half removed them and left behind remnants of them
they aren't worth having around

DonovanDMC (changes requested):
* setting presence does not work (missing required properties, see previous comment)
* `Client#getChannel` still checks for group channels, which no longer exist
* various remnants of GroupChannels have been left around (like [Client#groupChannels](https://github.com/bsian03/eris/blob/selfbot/lib/Client.js#L57) jsdoc, meanwhile neither the property nor class exist)

DonovanDMC:
More things I've just found, the event definitions for `callCreate`, `callDelete`, `callRing` & `callUpdate` are still floating around

lib/Client.js
```diff
@@ -2691,11 +2455,10 @@ class Client extends EventEmitter {
 
     /**
     * Get data on an OAuth2 application
```
DonovanDMC: This should be rephrased to reflect the singular nature of the function

gamer0mega:
you may aswell want to remove the hit property from Message, as it is meant for a user-only search feature

index.d.ts
```diff
@@ -3068,11 +2874,9 @@ declare namespace Eris {
     createdAt: CT extends "withMetadata" ? number : null;
     guild: CT extends "withMetadata"
       ? Guild // Invite with Metadata always has guild prop
-      : CH extends Extract<InviteChannel, GroupChannel> // Invite without Metadata
-        ? never // If the channel is GroupChannel, there is no guild
-        : CH extends Exclude<InviteChannel, InvitePartialChannel> // Invite without Metadata and not GroupChanel
-          ? Guild // If the invite channel is not partial
-          : Guild | undefined; // If the invite channel is partial
```
DonovanDMC: Bots can still fetch group invites
image

lib/Client.js
```diff
@@ -218,23 +211,19 @@ class Client extends EventEmitter {
         this.lastConnect = 0;
         this.channelGuildMap = {};
         this.threadGuildMap = {};
-        this.groupChannels = new Collection(GroupChannel);
+        this.shards = new ShardManager(this);
```
DonovanDMC: duplicate shard manager declaration

lib/voice/VoiceConnection.js
```diff
@@ -70,7 +70,7 @@ converterCommand.pickCommand = function pickCommand() {
 * @prop {Number} current.pausedTimestamp The timestamp of the most recent pause
 * @prop {Number} current.playTime How long the current stream has been playing for, in milliseconds
 * @prop {Number} current.startTime The timestamp of the start of the current stream
-* @prop {String} id The ID of the voice connection (guild ID)
```
DonovanDMC: having `(guild ID)` in parenthesis seems clearer than "the guild id of the voice connection"

DonovanDMC (changes requested):
besides the above, LGTM

lib/Client.js
bsian03: what I thought I removed that during the merge conflict

DonovanDMC:
I've gone back through some documentation, and now see that GroupChannels are not purely userbot features. ~~Any removals or changes should be undone.~~ Correction: The removals from they gateway are fine, all the other removals should be undone.

#1406 properly implements the official api methods for them.

HeadTriXz:
> I've gone back through some documentation, and now see that GroupChannels are not purely userbot features. Any removals or changes should be undone.
> 
> #1406 properly implements the official api methods for them.

Bots can not join group channels anymore though, some bots may still be in one because it was possible in the past. But they do not receive any messages.

DonovanDMC:
I didn't say they could join them, they can't be received over the gateway sure but they can be retrieved from rest just fine if the bot created it

bsian03:
~~Don't think it's worth maintaining/keeping in Eris. The % of bots having continuous access to group channels is miniscule, if anything~~
Removed gateway references

index.d.ts
bsian03: Can probably just type this as InvitePartial in that case

bsian03: Wait, is this not the correct behaviour anyways? We could check for channel type but that causes more headaches for not much profit

HeadTriXz: UnknownChannel when? 👉👈

bsian03:
https://github.com/bsian03/eris/compare/f4db90e3e4d0271e8bfcd583f5d182182d7d6e6c...516883192a40014dca6ef434b121be134bf44a76 for reference

index.d.ts
```diff
@@ -958,13 +951,9 @@ declare namespace Eris {
     messageReactionRemoveAll: [message: PossiblyUncachedMessage];
     messageReactionRemoveEmoji: [message: PossiblyUncachedMessage, emoji: PartialEmoji];
     messageUpdate: [message: Message<PossiblyUncachedTextableChannel>, oldMessage: OldMessage | null];
-    presenceUpdate: [other: Member | Relationship, oldPresence: Presence | null];
```
TTtie, line 930: re-add back

index.d.ts
```diff
@@ -3168,9 +3081,6 @@ declare namespace Eris {
     removeMessageReaction(channelID: string, messageID: string, reaction: string, userID?: string): Promise<void>;
     removeMessageReactionEmoji(channelID: string, messageID: string, reaction: string): Promise<void>;
     removeMessageReactions(channelID: string, messageID: string): Promise<void>;
-    removeRelationship(userID: string): Promise<void>;
-    searchChannelMessages(channelID: string, query: SearchOptions): Promise<SearchResults>;
-    searchGuildMembers(guildID: string, query: string, limit?: number): Promise<Member[]>;
     searchGuildMessages(guildID: string, query: SearchOptions): Promise<SearchResults>;
```
TTtie: re-introduce searchGuildMembers back (searchGuildMessages is a user-only/undocumented endpoint)
```suggestion
    searchGuildMembers(guildID: string, query: string, limit?: number): Promise<Member[]>;
```

lib/gateway/Shard.js
```diff
@@ -819,6 +759,18 @@ class Shard extends EventEmitter {
                 }
                 if((!member && packet.d.user.username) || oldPresence) {
                     member = guild.members.update(packet.d, guild);
+                    /**
+                     * Fired when a guild member or relationship's status or game changes
+                     * @event Client#presenceUpdate
+                     * @prop {Member | Relationship} other The updated member or relationship
```
TTtie:
```suggestion
                     * Fired when a guild member's status or game changes
                     * @event Client#presenceUpdate
                     * @prop {Member} other The updated member
```

lib/gateway/Shard.js
```diff
@@ -1945,116 +1848,11 @@ class Shard extends EventEmitter {
...
-                /**
-                 * Fired when a user leaves a group channel
-                 * @event Client#channelRecipientRemove
-                 * @prop {GroupChannel} channel The channel
-                 * @prop {User} user The user
-                 */
-                this.emit("channelRecipientRemove", channel, channel.recipients.remove(packet.d.user));
-                break;
-            }
             case "FRIEND_SUGGESTION_CREATE": {
```
TTtie: FRIEND_SUGGESTION_{CREATE,DELETE} are user-only (ditto in typedefs)

TTtie (changes requested):
took a very, very quick glance over this, here's what I've found

index.d.ts
bsian03: Thanks, must have removed this by accident

DonovanDMC (changes requested):
Getting the call property is still possible in some capacity:
https://github.com/discord/discord-api-docs/pull/6864

PR #1339, opened 2022-02-11
Use maxShards instead of lastShardID

GweeKwee:
If we spawn a client with a defined range of shards, a guild may point to the wrong shard id.

### DIFF
--- a/lib/structures/Guild.js
+++ b/lib/structures/Guild.js
@@ -81,7 +81,7 @@ class Guild extends Base {
     constructor(data, client) {
         super(data.id);
         this._client = client;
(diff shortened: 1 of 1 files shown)


DonovanDMC (approved):
I've never noticed any issues, but an issue happening here would make sense.

bsian03 (approved):
untested, but lgtm

curtisf (approved):

Dramex:
What if the client doesn’t have all the shards (aka cluster) is this still gonna work? 🤔

DonovanDMC:
~~That's the problem this fixes
maxShards is passed on to Discord on all gateway sessions, so guild ids should be based off of the total number of shards overall, not what's just present in the current process/cluster~~

That shouldn't be a problem since each process will only get the guilds for the shards they have

PR #1340, opened 2022-02-14
Add position to `CreateChannelOptions`

frobinsonj:
`position` should be included as an optional parameter when creating or modifying a guild channel. 

Create: https://discord.com/developers/docs/resources/guild#:~:text=%2C%20are%20unaffected-,position,-integer

Modify: https://discord.com/developers/docs/resources/channel#:~:text=Text%2C%20News-,position,-%3Finteger

NOTE: Also fixed the order of `options` parameters in `GuildChannel#edit()` (should have been alphabetical)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -286,6 +286,7 @@ declare namespace Eris {
     nsfw?: boolean;
     parentID?: string;
     permissionOverwrites?: Overwrite[];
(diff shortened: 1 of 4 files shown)


bsian03 (approved):
untested, but lgtm

PR #1344, opened 2022-03-01
Handle Shared Ratelimits & Parse retryAfter Correctly

curtisf:
Currently, Eris uses parseInt to parse the ratelimit x-ratelimit-reset-after header. parseInt slices off the decimal of the ratelimit header, and as such, Eris makes requests too early sometimes. Try to send any request as fast as possible, and log the debug event. When you're at the limit exhausting a ratelimit bucket, Eris will make requests too early, and the printed retryAfter is the time sliced off the decimal + network latency.

Similarly, Eris does not handle shared ratelimits. If you log the debug event and send 100 webhook messages to a channel as fast as possible, you will encounter a shared ratelimit after many messages are sent. Eris continues to use x-ratelimit-reset-after, while the retry-after header has the true shared ratelimit. In production, this can cause Cloudflare bans when using webhooks heavily (Eris waiting 2000ms from x-ratelimit-reset-after header, while retry-after is 40+ seconds).

### DIFF
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -350,12 +350,6 @@ class RequestHandler {
                             } else {
                                 err = new DiscordHTTPError(req, resp, response, stack);
                             }
-                            if (err.code === 429) {
-                                Object.defineProperty(err, 'headers', {
-                                    value: resp.headers,
-                                    enumerable: false
-                                })
-                            }
                             reject(err);
                             return;
                         }
--- a/lib/rest/RequestHandler.js
+++ b/lib/rest/RequestHandler.js
@@ -273,7 +273,7 @@ class RequestHandler {
 
                         this.ratelimits[route].remaining = resp.headers["x-ratelimit-remaining"] === undefined ? 1 : +resp.headers["x-ratelimit-remaining"] || 0;
 
-                        const retryAfter = parseInt(resp.headers["x-ratelimit-reset-after"] || resp.headers["retry-after"]) * 1000;
+                        const retryAfter = Number(resp.headers["x-ratelimit-reset-after"] || resp.headers["retry-after"]) * 1000;
                         if(retryAfter >= 0) {
                             if(resp.headers["x-ratelimit-global"]) {
                                 this.globalBlock = true;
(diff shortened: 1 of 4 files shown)


lib/rest/RequestHandler.js
```diff
@@ -299,12 +299,13 @@ class RequestHandler {
                         if(resp.statusCode >= 300) {
                             if(resp.statusCode === 429) {
                                 const content = typeof body === "object" ? `${body.content} ` : "";
-                                this._client.emit("debug", `${resp.headers["x-ratelimit-global"] ? "Global" : "Unexpected"} 429 (╯°□°）╯︵ ┻━┻: ${response}\n${content} ${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
-                                if(retryAfter) {
+                                const delay = resp.headers['retry-after'] ? Number(resp.headers['retry-after']) * 1000 : retryAfter
```
HeadTriXz: Handle shared ratelimits by using `response.retry_after`, not from the header (keep in mind that response is a string and needs to be deserialized).

HeadTriXz (changes requested):

lib/rest/RequestHandler.js
bsian03: that - should be _ but yes, or you can use x-ratelimit-reset-after

bsian03 (changes requested):
see above conversation

lib/rest/RequestHandler.js
HeadTriXz: Headers aren't correct/accurate for shared ratelimits, you'll have to use `response.retry_after`.

bsian03:
Also would it be possible to add a `shared` prop to the error body, I know it comes from the header but it would be useful for diagnostic purposes

lib/rest/RequestHandler.js
```diff
@@ -299,12 +299,21 @@ class RequestHandler {
                         if(resp.statusCode >= 300) {
                             if(resp.statusCode === 429) {
                                 const content = typeof body === "object" ? `${body.content} ` : "";
-                                this._client.emit("debug", `${resp.headers["x-ratelimit-global"] ? "Global" : "Unexpected"} 429 (╯°□°）╯︵ ┻━┻: ${response}\n${content} ${now} ${route} ${resp.statusCode}: ${latency}ms (${this.latencyRef.latency}ms avg) | ${this.ratelimits[route].remaining}/${this.ratelimits[route].limit} left | Reset ${this.ratelimits[route].reset} (${this.ratelimits[route].reset - now}ms left)`);
-                                if(retryAfter) {
+                                let delay = retryAfter
```
HeadTriXz: Add a semicolon.

lib/rest/RequestHandler.js
```diff
@@ -299,12 +299,21 @@ class RequestHandler {
...
+                                if (resp.headers['x-ratelimit-scope'] === 'shared') {
```
HeadTriXz: Remove space after `if`.

HeadTriXz (changes requested):

HeadTriXz:
In my opinion, adding that to the response is just dumb. You can already access it through the `rawREST` event.

curtisf:
Talk to bsian idk

HeadTriXz:
> Talk to bsian idk

@bsian03 ?

bsian03:
We decided to add headers to any http/rest error response with `enumerable: false` instead of modifying Discord's response to include scope

bsian03 (changes requested):
See above

HeadTriXz:
> We decided to add headers to any http/rest error response with `enumerable: false` instead of modifying Discord's response to include scope

The response already exists on the error, which includes the header. What's the point of adding it?

If you do decide to add this, make sure to also update the types.

bsian03:
> The response already exists on the error, which includes the header. What's the point of adding it?

We do the same thing for the response and status code. It's easier to access

bsian03 (approved):

PR #1350, opened 2022-03-13
Fixed editing command permissions not working

NotChris123:


coolcalcacol:
Based Fix.

face-hh:
Based Fix.

PR #1355, opened 2022-03-28
fix(constants): Include manageThreads, manageEvents in Permissions.all

eritbh:
Adds the permissions constants `manageThreads` to `allText` and `manageEvents` to `allGuild`. Recalculates the constant types of `all`, `allText`, and `allGuild` in typings. #1271 mistakenly removed `manageThreads` from `allText` and introduced `manageEvents` without adding it to `allGuild`.


DonovanDMC (approved):

rilysh (approved):

bsian03 (approved):
Not merging yet

PR #1358, opened 2022-04-12
Attachment type

coolcalcacol:


ghost:
Is this solely for the attachment type? There is also an attachments map in command interaction.

coolcalcacol:
@Linker-123 Is that better?

ghost:
Add types for attachment class

index.d.ts
```diff
@@ -2005,6 +2008,19 @@ declare namespace Eris {
...
+    ephemeral?: boolean;
+    filename: string;
+    height?: number;
+    id: string;
+    proxy_url: string;
+    size: number;
+    url: string;
+    width?: number;
+  }
+
```
bsian03: Follow structure of other classes

lib/structures/Attachment.js
```diff
@@ -0,0 +1,49 @@
...
+ * @prop {id} id The attachment ID
+ * @prop {string} filename The filename of the attachment
+ * @prop {string?} description The description of the attachment
+ * @prop {string?} contentType The content type of the attachment
+ * @prop {number} size The size of the attachment
+ * @prop {string} url The URL of the attachment
+ * @prop {string} proxyUrl The proxy URL of the attachment
+ * @prop {string?} height The height of the attachment
+ * @prop {string?} width The width of the attachment
+ * @prop {boolean?} ephemeral Whether the attachment is ephemeral
```
bsian03: Invalid type names

bsian03 (changes requested):

lib/structures/Attachment.js
```diff
@@ -4,16 +4,17 @@ const Base = require("./Base");
...
+ * @prop {id} data.id The attachment ID
+ * @prop {string} data.filename The filename of the attachment
+ * @prop {string?} data.description The description of the attachment
+ * @prop {string?} data.contentType The content type of the attachment
+ * @prop {number} data.size The size of the attachment
+ * @prop {string} data.url The URL of the attachment
+ * @prop {string} data.proxyUrl The proxy URL of the attachment
+ * @prop {string?} data.height The height of the attachment
+ * @prop {string?} data.width The width of the attachment
+ * @prop {boolean?} data.ephemeral Whether the attachment is ephemeral
```
DonovanDMC:
```suggestion
 * @prop {Object} data The attachment data
 * @prop {String} data.id The attachment ID
 * @prop {String} data.filename The filename of the attachment
 * @prop {String?} data.description The description of the attachment
 * @prop {String?} data.contentType The content type of the attachment
 * @prop {Number} data.size The size of the attachment
 * @prop {String} data.url The URL of the attachment
 * @prop {String} data.proxyUrl The proxy URL of the attachment
 * @prop {Number?} data.height The height of the attachment
 * @prop {Number?} data.width The width of the attachment
 * @prop {Boolean?} data.ephemeral Whether the attachment is ephemeral
 ```

Comment (HeadTriXz) [reply lib/structures/Attachment.js]: `id` isn't a type. And the `data` object is not a `string`

**Review by HeadTriXz — CHANGES_REQUESTED** [review]

File: lib/structures/Attachment.js (hunk 3 above)
Comment (DonovanDMC) [reply lib/structures/Attachment.js]: > `id` isn't a type. And the `data` object is not a `string`

Those were his 🤷‍♂️

File: lib/structures/Attachment.js [hunk 4]
```diff
@@ -0,0 +1,50 @@
...
+ * @prop {String} data.id The attachment ID
+ * @prop {String} data.filename The filename of the attachment
+ * @prop {String?} data.description The description of the attachment
+ * @prop {String?} data.contentType The content type of the attachment
+ * @prop {Number} data.size The size of the attachment
+ * @prop {String} data.url The URL of the attachment
+ * @prop {String} data.proxyUrl The proxy URL of the attachment
+ * @prop {Number?} data.height The height of the attachment
+ * @prop {Number?} data.width The width of the attachment
+ * @prop {Boolean?} data.ephemeral Whether the attachment is ephemeral
```
bsian03: What's with all the `data.`s?

bsian03 (changes requested):

lib/structures/Attachment.js
HeadTriXz: Oh, it's for a class, I thought it were options for a method or something 💀

coolcalcacol: should be fixed there, I'm just being stupid

lib/structures/Attachment.js
```diff
@@ -0,0 +1,49 @@
...
+ * @prop {String} id The attachment ID
+ * @prop {String} filename The filename of the attachment
+ * @prop {String?} description The description of the attachment
+ * @prop {String?} contentType The content type of the attachment
+ * @prop {Number} size The size of the attachment
+ * @prop {String} url The URL of the attachment
+ * @prop {String} proxyUrl The proxy URL of the attachment
+ * @prop {Number?} height The height of the attachment
+ * @prop {Number?} width The width of the attachment
+ * @prop {Boolean?} ephemeral Whether the attachment is ephemeral
```
bsian03: Alphabetical order

index.d.ts
```diff
@@ -2005,6 +2008,19 @@ declare namespace Eris {
...
+  export class Attachment extends Base {
+    content_type?: string;
+    ephemeral?: boolean;
+    filename: string;
+    height?: number;
+    id: string;
+    proxy_url: string;
+    size: number;
+    url: string;
+    width?: number;
```
bsian03: camelCase matching class

index.d.ts
```diff
@@ -2005,6 +2008,19 @@ declare namespace Eris {
...
+    constructor(data: BaseData);
```
bsian03, line 2021: Useless constructor

bsian03 (changes requested):

lib/structures/CommandInteraction.js
```diff
@@ -95,6 +98,14 @@ class CommandInteraction extends Interaction {
                 });
                 this.data.resolved.messages = messagemap;
             }
+            //Attachments
+            if(info.data.resolved.attachments !== undefined) {
+                const attachmentsmap = new Collection(Attachment);
+                Object.entries(info.data.resolved.attachments).forEach(([id, attachment]) => {
+                    attachmentsmap.set(id, new Attachment(attachment, this._client));
```
TTtie: Passing `this._client` to the constructor is redundant (only attachment data is needed in the `Attachment` class constructor):

```suggestion
                    attachmentsmap.set(id, new Attachment(attachment));
```

coolcalcacol:
what was the conclusion with the attachments class vs just using objects?

DonovanDMC (dismissed):
sort properties alphabetically

DonovanDMC (changes requested):
Why do we have both an interface and a class named `Attachment`? These will be combined when using `Attachment`, as can be seen here

image

coolcalcacol:
The interface pre-dates this pr. Before I headed away I was considering merging it but that would mean processing the message to create attachment classes for each attachment

DonovanDMC:
That sounds like the best idea
we've got it now, so why not use it

it's either that or rename one of them

lib/structures/Message.js
```diff
@@ -279,8 +288,15 @@ class Message extends Base {
         if(data.tts !== undefined) {
             this.tts = data.tts;
         }
-        if(data.attachments !== undefined) {
-            this.attachments = data.attachments;
+        if(data.attachments) {
+            for(const [id] of this.attachments.entries()) {
```
HeadTriXz: `this.attachments.keys()` exists. Also, you could also just clear the collection, and add the new ones.

HeadTriXz (changes requested):

lib/structures/Message.js
DonovanDMC: clearing would recreate the objects, I'm doing this specifically to avoid that

DonovanDMC (approved):

index.d.ts
```diff
@@ -2029,6 +2021,20 @@ declare namespace Eris {
...
   }
 
+  export class Attachment extends Base {
+    contentType?: string;
+    description?: string;
+    ephemeral?: boolean;
+    filename: string;
+    height?: number;
+    id: string;
+    proxyUrl: string;
```
```suggestion
    proxyURL: string;
```

lib/structures/Attachment.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Represents an attachment
+ * @prop {String?} contentType The content type of the attachment
+ * @prop {String?} description The description of the attachment
+ * @prop {Boolean?} ephemeral Whether the attachment is ephemeral
+ * @prop {String} filename The filename of the attachment
+ * @prop {Number?} height The height of the attachment
+ * @prop {String} id The attachment ID
+ * @prop {String} proxyUrl The proxy URL of the attachment
```
bsian03:
```suggestion
 * @prop {String} proxyURL The proxy URL of the attachment
```

lib/structures/Attachment.js
```diff
@@ -0,0 +1,59 @@
...
+ * @prop {Number?} width The width of the attachment
+ */
+class Attachment extends Base {
+    constructor(data) {
+        super(data.id);
+
+        this.filename = data.filename;
+        this.size = data.size;
+        this.url = data.url;
+        this.proxyUrl = data.proxy_url;
```
bsian03:
```suggestion
        this.proxyURL = data.proxy_url;
```

bsian03 (changes requested):
`URL` is used everywhere else, not `Url`

lib/structures/CommandInteraction.js
```diff
@@ -6,35 +6,36 @@ const User = require("./User");
...
+* @prop {Array<Object>?} data.options[].options The run Application Command options (Mutually exclusive with value)
+* @prop {Object?} data.resolved converted users + roles + channels
+* @prop {Collection<User>?} data.resolved.users converted users
+* @prop {Collection<Member>?} data.resolved.members converted members
+* @prop {Collection<Role>?} data.resolved.roles converted roles
+* @prop {Collection<Channel>?} data.resolved.channels converted channels
+* @prop {String?} guildID The ID of the guild in which the interaction was created
+* @prop {Member?} member The member who triggered the interaction (This is only sent when the interaction is invoked within a guild)
+* @prop {User?} user The user who triggered the interaction (This is only sent when the interaction is invoked within a dm)
+*/
```
coolcalcacol: Was this sorted in a recent PR and I've unsorted it?

bsian03: Yes

bsian03:
discord/discord-api-docs#6272 to be added
discord/discord-api-docs#6945

lib/structures/Message.js
```diff
@@ -166,6 +169,12 @@ class Message extends Base {
       this.member = null;
     }
 
+    if (data.attachments) {
+      for (const attachment of data.attachments) {
+        this.attachments.add(attachment, this);
```
bsian03, line 174: This needs to be able to update/delete, or overwrite completely

lib/structures/Attachment.js
bsian03, line 1: I feel like some of these properties can be modified

bsian03 (changes requested):

PR #1363, opened 2022-04-16
feat(getGuildBans): support pagination

HeadTriXz:
Adds support for guild bans pagination: https://discord.com/developers/docs/resources/guild#get-guild-bans

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -2345,7 +2345,7 @@ class Client extends EventEmitter {
             limit: options.limit && Math.min(options.limit, 1000)
         });
 
-        for(let ban of bans) {
+        for(const ban of bans) {
             ban.user = this.users.update(ban.user, this);
         }
 
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -2339,33 +2339,31 @@ class Client extends EventEmitter {
     * @returns {Promise<Array<Object>>} Resolves with an array of { reason: String, user: User }
     */
     async getGuildBans(guildID, options = {}) {
-        const bans = await this.requestHandler.request("GET", Endpoints.GUILD_BANS(guildID), true, {
+        let bans = await this.requestHandler.request("GET", Endpoints.GUILD_BANS(guildID), true, {
             after: options.after,
             before: options.before,
-            limit: Math.min(options.limit, 1000) 
+            limit: options.limit && Math.min(options.limit, 1000)
         });
-        
-        bans.forEach((ban) => {
-            ban.user = this.users.update(ban.user, this);
-        });
-        
-        if(options.limit && options.limit > 1000 && bans.length >= 1000) {
-            if(options.before) {
-                options.after = undefined;
-                options.before = bans[0].user.id;
-            } else {
-                options.after = bans[bans.length - 1].user.id;
-            }
 
+        for(let ban of bans) {
+            ban.user = this.users.update(ban.user, this);
+        }
+
+        if(options.limit && options.limit > 1000 && bans.length >= 1000) {
             const page = await this.getGuildBans(guildID, {
-                after: options.after,
-                before: options.before,
+                after: options.before ? undefined : bans[bans.length - 1].user.id,
+                before: options.before ? bans[0].user.id : undefined,
                 limit: options.limit - bans.length
             });
- 
-            bans.push(...page);
+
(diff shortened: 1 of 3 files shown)


lib/Client.js
```diff
@@ -2332,15 +2332,41 @@ class Client extends EventEmitter {
...
+    * @returns {Promise<Array<Object>>} Resolves with an array of { reason: String, user: User }
+    */
+    async getGuildBans(guildID, options = {}) {
+        const bans = await this.requestHandler.request("GET", Endpoints.GUILD_BANS(guildID), true, {
+            after: options.after,
+            before: options.before,
+            limit: Math.min(options.limit, 1000) 
         });
+        
+        bans.forEach((ban) => {
```
ghost: forEach isn't synchronous

HeadTriXz: Discussed in Discord.

ghost (dismissed):

lib/Client.js
bsian03: Change this to `map`

lib/Client.js
```diff
@@ -2332,15 +2332,41 @@ class Client extends EventEmitter {
...
-    getGuildBans(guildID) {
-        return this.requestHandler.request("GET", Endpoints.GUILD_BANS(guildID), true).then((bans) => {
-            bans.forEach((ban) => {
-                ban.user = new User(ban.user, this);
-            });
-            return bans;
+    * @arg {Object} [options] Options for the request
+    * @arg {String} [options.after] Only get users after given user ID
+    * @arg {String} [options.before] Only get users before given user ID
+    * @arg {Number} [options.limit] The maximum number of users to return
```
bsian03: Add the default limit here

lib/Client.js
```diff
@@ -2332,15 +2332,41 @@ class Client extends EventEmitter {
...
+            limit: Math.min(options.limit, 1000) 
         });
+        
+        bans.forEach((ban) => {
+            ban.user = this.users.update(ban.user, this);
+        });
+        
+        if(options.limit && options.limit > 1000 && bans.length >= 1000) {
+            if(options.before) {
+                options.after = undefined;
```
bsian03: May cause unintended side effects if a user supplies both options.before and options.after

bsian03 (dismissed):

lib/Client.js
HeadTriXz: Discord API docs state that they will use `before` if both are supplied.

lib/Client.js
HeadTriXz: What's the default limit?

lib/Client.js
DonovanDMC: Why should this be a map? We aren't returning a value

HeadTriXz: `.map` will return an array with new objects. `.forEach` was used before and only updates the value of the property. An other option would be to use a for loop.

lib/Client.js
DonovanDMC: https://discord.com/developers/docs/resources/guild#get-guild-bans

1000

lib/Client.js
bsian03: Should also say move this logic to the return itself with the `.map`

HeadTriXz: Would mess up the recursion.

bsian03: Resolved within discord. turns out i thought similar circumstance logic was different. whoops

DonovanDMC (dismissed):

lib/Client.js
```diff
@@ -2332,15 +2332,39 @@ class Client extends EventEmitter {
...
+        if(options.limit && options.limit > 1000 && bans.length >= 1000) {
+            const page = await this.getGuildBans(guildID, {
+                after: options.before ? undefined : bans[bans.length - 1].user.id,
+                before: options.before ? bans[0].user.id : undefined,
+                limit: options.limit - bans.length
+            });
+
+            if(options.before) {
+                page.push(...bans);
+                bans = page;
```
HeadTriXz: `Array#unshift` exists.

DonovanDMC: we literally.. already talked about that

HeadTriXz: I have great wifi in the train 🚂

abalabahaha (approved):

PR #1298, opened 2021-11-15
Text in Guild Voice Channels

beanjo55:
Adds support for upcoming text in guild voice channels, draft until discord finalizes the spec for them and is no longer subject to change. Only applies to GUILD_VOICE, stage channels not included

TODO:
- [X] Create GuildVoiceChannel class, extends VoiceChannel, and add appropriate text props since no multiple inheritance :(
- [X] Update channel struct to use new GuildVoiceChannel class
- [X] First pass as TS typings
- [x] Check jsdoc to use the new channel class and the channel is referenced where appropriate
- [x] Fix TS type issue with messages collection
- [x] Ensure lint is correct

Please let me know if i missed anything, or if theres a better or more appropriate way to implement!

Fixes #1381

PR #1365, opened 2022-04-23
feat(VoiceChannel): Textable Voice Channels

DonovanDMC:
This pr is currently just some guesswork & assumptions since no documentation exists for this yet.

Currently, `VoiceChannel` has been fitted to extend `TextChannel` & `GuildStage` now extends `GuildChannel` instead of `VoiceChannel`.


bsian03:
Dupe #1298

PR #1366, opened 2022-04-27
fix(userUpdate): Uncached users throwing undefined

Dramex:
when you update the bot username or avatar or discriminator it fires an undefined on the userUpdate event.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -1965,7 +1965,7 @@ class Shard extends EventEmitter {
                     discriminator: user.discriminator,
                     avatar: user.avatar
                 };
-                this.emit("userUpdate", user.update(packet.d), oldUser);
+                this.emit("userUpdate", this.client.users.update(packet.d), oldUser);
                 break;
             }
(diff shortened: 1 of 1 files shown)


bsian03 (dismissed):
Not merging yet

bsian03 (approved):

Dramex:
@bsian03  The old logic should be fine btw because USER_UPDATE fires only with the current user update ( the bot ) in our case which should always be cached on ready

the cache was fine, the issue was user.update() does not return anything, and its actually not being used outside so I used client.users.update() instead
https://github.com/abalabahaha/eris/blob/017c18aa3d6b2ccdc1ff5db7b61f52b9f2d974d4/lib/structures/User.js#L37

the event userUpdate on **eris** is used when members get updated too. which different than discord USER_UPDATE

^ just to clerify

bsian03:
That was my brain freaking out that the thing would've been breaking (didn't clock it would be only the bot earlier, but did last night), hence the change. Also you missed out the client param in the update.

Oh well, doesn't hurt to have a fallback. Who knows, discord might change something 🤷

PR #1367, opened 2022-04-28
fix(getGuildAuditLog): cache users & threads earlier

DonovanDMC:
Currently, users are added to the cache too late when audit log entries are fetched, so they will not be present in the entry itself if the user was not previously cached. This pull request remedies that by pulling it out of the object itself, and making it before the entries are constructed.

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -2293,11 +2293,13 @@ class Client extends EventEmitter {
         }
         return this.requestHandler.request("GET", Endpoints.GUILD_AUDIT_LOGS(guildID), true, options).then((data) => {
             const guild = this.guilds.get(guildID);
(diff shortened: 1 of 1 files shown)


bsian03:
~~Cannot replicate this issue~~ retracting comment, misunderstood where the issue was

DonovanDMC:
I can easily reproduce this with a large server

```
const Eris = require(".");
const client = new Eris.Client("Bot [TOKEN]");
client.on("ready", async() => {
    const log = await client.getGuildAuditLog("");
    console.log(log.entries.filter((l) => l.user === undefined).length);
});
client.connect();
```
image

abalabahaha (approved):

PR #1368, opened 2022-04-28
Corrected various typings

oathompsonjones:
- Corrected return types involving Message to specify the type of channel the Message is in where possible.
- Corrected guildId and member types on Interaction interfaces to reflect whether the channel is in a guild or not, as is done with the Message type.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2384,12 +2384,12 @@ declare namespace Eris {
     createGuildSticker(guildID: string, options: CreateStickerOptions, reason?: string): Promise<Sticker>;
     createGuildTemplate(guildID: string, name: string, description?: string | null): Promise<GuildTemplate>;
     createInteractionResponse(interactionID: string, interactionToken: string, options: InteractionOptions, file?: FileContent | FileContent[]): Promise<void>;
-    createMessage<T extends TextableChannel = TextableChannel>(channelID: string, content: MessageContent, file?: FileContent | FileContent[]): Promise<Message<T>>;
+    createMessage(channelID: string, content: MessageContent, file?: FileContent | FileContent[]): Promise<Message>;
     createRole(guildID: string, options?: RoleOptions | Role, reason?: string): Promise<Role>;
     createStageInstance(channelID: string, options: StageInstanceOptions): Promise<StageInstance>;
     createThreadWithMessage(channelID: string, messageID: string, options: CreateThreadOptions): Promise<NewsThreadChannel | PublicThreadChannel>;
     createThreadWithoutMessage(channelID: string, options: CreateThreadWithoutMessageOptions): Promise<PrivateThreadChannel>;
-    crosspostMessage<T extends TextableChannel = TextableChannel>(channelID: string, messageID: string): Promise<Message<T>>;
+    crosspostMessage(channelID: string, messageID: string): Promise<Message>;
     deleteChannel(channelID: string, reason?: string): Promise<void>;
     deleteChannelPermission(channelID: string, overwriteID: string, reason?: string): Promise<void>;
     deleteCommand(commandID: string): Promise<void>;
@@ -2449,7 +2449,7 @@ declare namespace Eris {
     editGuildVoiceState(guildID: string, options: VoiceStateOptions, userID?: string): Promise<void>;
     editGuildWelcomeScreen(guildID: string, options: WelcomeScreenOptions): Promise<WelcomeScreen>;
     editGuildWidget(guildID: string, options: Widget): Promise<Widget>;
-    editMessage<T extends TextableChannel = TextableChannel>(channelID: string, messageID: string, content: MessageContentEdit): Promise<Message<T>>;
+    editMessage(channelID: string, messageID: string, content: MessageContentEdit): Promise<Message>;
     /** @deprecated */
     editNickname(guildID: string, nick: string, reason?: string): Promise<void>;
     editRole(guildID: string, roleID: string, options: RoleOptions, reason?: string): Promise<Role>; // TODO not all options are available?
@@ -2471,12 +2471,12 @@ declare namespace Eris {
       token?: string,
       reason?: string
     ): Promise<Webhook>;
-    editWebhookMessage<T extends GuildTextableChannel = GuildTextableChannel>(
+    editWebhookMessage(
       webhookID: string,
       token: string,
       messageID: string,
       options: WebhookPayloadEdit
-    ): Promise<Message<T>>;
+    ): Promise<Message>;
     emit<K extends keyof ClientEvents>(event: K, ...args: ClientEvents[K]): boolean;
     emit(event: string, ...args: any[]): boolean;
     enableSelfMFATOTP(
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -2910,23 +2910,23 @@ declare namespace Eris {
...
         channels?: Collection<PartialChannel>;
-        messages?: Collection<Message>;
+        messages?: Collection<Message<T>>;
       };
       options?: InteractionDataOptions[];
     };
-    guildID?: string;
-    member?: Member;
+    guildID: T extends GuildTextableWithThread ? string : undefined;
+    member: T extends GuildTextableWithThread ? Member : null;
```
DonovanDMC: Member is not nullable - it's the same a guildID, and these types set the guildID as the explicit value "undefined", which it is not - never would be more accurate (making the value as not present at all)

index.d.ts
```diff
@@ -2943,22 +2943,22 @@ declare namespace Eris {
   export class ComponentInteraction<T extends PossiblyUncachedTextable = TextableChannel> extends Interaction {
     channel: T;
     data: ComponentInteractionButtonData | ComponentInteractionSelectMenuData;
-    guildID?: string;
-    member?: Member;
-    message: Message;
+    guildID: T extends GuildTextableWithThread ? string : undefined;
+    member: T extends GuildTextableWithThread ? Member : null;
```
DonovanDMC: same here

index.d.ts
```diff
@@ -2979,22 +2979,22 @@ declare namespace Eris {
   export class UnknownInteraction<T extends PossiblyUncachedTextable = TextableChannel> extends Interaction {
     channel?: T;
     data?: unknown;
-    guildID?: string;
-    member?: Member;
-    message?: Message;
+    guildID: T extends GuildTextableWithThread ? string : undefined;
+    member: T extends GuildTextableWithThread ? Member : null;
```
DonovanDMC: same here

DonovanDMC (changes requested):

DonovanDMC:
also, the AutocompleteInteraction interface should be included in this

DonovanDMC:
Another thing to mention, if you're explicitly setting member as being present or not, the same should be done to the user property

it's the exact opposite, won't be present when in a guild

DonovanDMC (dismissed):

index.d.ts
```diff
@@ -2240,12 +2240,12 @@ declare namespace Eris {
...
     createInteractionResponse(interactionID: string, interactionToken: string, options: InteractionOptions, file?: FileContent | FileContent[]): Promise<void>;
-    createMessage(channelID: string, content: MessageContent, file?: FileContent | FileContent[]): Promise<Message>;
+    createMessage<T extends TextableChannel = TextableChannel>(channelID: string, content: MessageContent, file?: FileContent | FileContent[]): Promise<Message<T>>;
     createRole(guildID: string, options?: RoleOptions | Role, reason?: string): Promise<Role>;
     createStageInstance(channelID: string, options: StageInstanceOptions): Promise<StageInstance>;
     createThreadWithMessage(channelID: string, messageID: string, options: CreateThreadOptions): Promise<NewsThreadChannel | PublicThreadChannel>;
     createThreadWithoutMessage(channelID: string, options: CreateThreadWithoutMessageOptions): Promise<PrivateThreadChannel>;
-    crosspostMessage(channelID: string, messageID: string): Promise<Message>;
+    crosspostMessage<T extends TextableChannel = TextableChannel>(channelID: string, messageID: string): Promise<Message<T>>;
```
bsian03: Developers should type cast here instead of relying on generics

index.d.ts
```diff
@@ -2910,23 +2910,23 @@ declare namespace Eris {
...
         channels?: Collection<PartialChannel>;
-        messages?: Collection<Message>;
+        messages?: Collection<Message<T>>;
       };
       options?: InteractionDataOptions[];
     };
-    guildID?: string;
-    member?: Member;
+    guildID: T extends GuildTextableWithThread ? string : never;
+    member: T extends GuildTextableWithThread ? Member : never;
```
bsian03, line 3128: How would this be displayed if TS can't figure out if it's a dm channel or guild channel

bsian03 (changes requested):
Remove the other generic functions too please

index.d.ts
DonovanDMC: ~~it doesn't display at all due to the never~~
actually it looks like it's just showing it as present either way due to the default value having both guild channels & private channels included
image
image
~~there really isn't a better solution besides not having a generic because an explicit undefined is wrong (unless we want to make them explicitly undefined in the js?)~~
the best solution would likely be not having a generic, it's a lot more mess than it's worth imo

DonovanDMC: on further inspection, this does not work as expected in any scenario

`guildID` & `member` are both being shown as present despite a non-guild interface being provided
image

index.d.ts
```diff
@@ -358,7 +358,9 @@ declare namespace Eris {
     getMessageReaction(messageID: string, reaction: string, limit?: number, before?: string, after?: string): Promise<User[]>;
     getMessages(options?: GetMessagesOptions): Promise<Message<this>[]>;
     /** @deprecated */
-    getMessages(limit?: number, before?: string, after?: string, around?: string): Promise<Message[]>;
+    getMessages(limit?: number, before?: string, after?: string, around?: string): Promise<Message<this>[]>;
+    getPins(): Promise<Message<this>[]>;
+    pinMessage(messageID: string): Promise<void>;
```
bsian03, line 363: These are already covered by Pinnable

index.d.ts
```diff
@@ -2475,16 +2476,16 @@ declare namespace Eris {
       token: string,
       messageID: string,
       options: WebhookPayloadEdit
-    ): Promise<Message<GuildTextableChannel>>;
```
bsian03, line 2478: Not sure why this was removed? Webhooks cannot exist in non guild channels

index.d.ts
```diff
@@ -2475,16 +2476,16 @@ declare namespace Eris {
...
     emit(event: string, ...args: any[]): boolean;
     enableSelfMFATOTP(
       secret: string,
       code: string
     ): Promise<{ backup_codes: { code: string; consumed: boolean }[]; token: string }>;
     executeSlackWebhook(webhookID: string, token: string, options: Record<string, unknown> & { auth?: boolean; threadID?: string }): Promise<void>;
-    executeSlackWebhook(webhookID: string, token: string, options: Record<string, unknown> & { auth?: boolean; threadID?: string; wait: true }): Promise<Message<GuildTextableChannel>>;
-    executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait: true }): Promise<Message<GuildTextableChannel>>;
+    executeSlackWebhook(webhookID: string, token: string, options: Record<string, unknown> & { auth?: boolean; threadID?: string; wait: true }): Promise<Message>;
+    executeWebhook(webhookID: string, token: string, options: WebhookPayload & { wait: true }): Promise<Message>;
```
bsian03, line 2488: See other webhook comment

index.d.ts
```diff
@@ -2598,7 +2599,7 @@ declare namespace Eris {
     getUserProfile(userID: string): Promise<UserProfile>;
     getVoiceRegions(guildID?: string): Promise<VoiceRegion[]>;
     getWebhook(webhookID: string, token?: string): Promise<Webhook>;
-    getWebhookMessage(webhookID: string, token: string, messageID: string): Promise<Message<GuildTextableChannel>>;
+    getWebhookMessage(webhookID: string, token: string, messageID: string): Promise<Message>;
```
bsian03, line 2602: See other webhook comment

index.d.ts
bsian03: Will try to get some more feedback on this

bsian03:
#1469

PR #1370, opened 2022-05-06
Improved max_concurrency aware ShardManager

HcgRandon:
This PR makes ShardManager `max_concurrency` aware. Shards in different `rate_limit_key` (https://discord.com/developers/docs/topics/gateway#sharding-max-concurrency) buckets start in parallel resulting in significantly faster startup times across the board for large scale discord bots. Here is a table (these were tested from intent enabled tokens):

| Old time | New Time | concur | % faster      | shards/total | other              |
|----------|----------|--------|---------------|--------------|--------------------|
| 224.251s | 68.312s  | 16     | ~69.5% (nice) | 16/1024      | getAllUsers: true  |
| 431.849s | 70.237s  | 16     | ~83.7%        | 32/1024      | getAllUsers: false |
| 79.591s  | 76.640s   | 1      | margin of err | 16/16        | getAllUsers: false |

### DIFF
--- a/lib/gateway/ShardManager.js
+++ b/lib/gateway/ShardManager.js
@@ -5,8 +5,6 @@ const Collection = require("../util/Collection");
 const Endpoints = require("../rest/Endpoints");
 const Shard = require("./Shard");
 
-const SHARD_IDENTIFY_RATELIMIT = 5000; // 5s
-
 class ShardManager extends Collection {
     constructor(client) {
         super(Shard);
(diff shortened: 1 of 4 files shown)


lib/gateway/ShardManager.js
```diff
@@ -2,29 +2,39 @@
 
 const Base = require("../structures/Base");
 const Collection = require("../util/Collection");
+const Endpoints = require("../rest/Endpoints");
 const Shard = require("./Shard");
 
+const SHARD_IDENTIFY_RATELIMIT = 5000; // 5s
```
DonovanDMC: this should be put into lib/Constants.js, or left hardcoded in the places it's used

DonovanDMC (changes requested):

DonovanDMC (dismissed):

DonovanDMC (approved):

PR #1371, opened 2022-05-07
feat(gateway/REST): V10 API changes

coolcalcacol:
Re-created due to github being github

## Breaking
- getActiveThreads has been removed, as the endpoint is no longer available 

## Additions
- messageContent intent added
- all and allPrivileged types and bit fields updated

## Changes
- Gateway/REST version 9 -> 10
- application.summary type changed from string to explicit ""

## TODO:
- document attachments changes, if anyone cares
- check if achievements are actually cared about

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1034,7 +1034,7 @@ declare namespace Eris {
     icon: string | null;
     id: string;
     name: string;
-    summary: string;
+    summary: ""; // Returns an empty string
   }
   interface IntegrationOptions {
     enableEmoticons?: string;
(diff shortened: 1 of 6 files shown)


DonovanDMC (dismissed):
If all the current changes are present, LGTM

frobinsonj:
https://discord.com/developers/docs/change-log#api-v10

> To specify a reason for an administrative action in audit logs, apps must now pass the `X-Audit-Log-Reason` header rather than the reason parameter for **all endpoints**.

The following should exception should probably be removed:
https://github.com/abalabahaha/eris/blob/4ea56e45ffe68e95d89d7f3edcc181097f4f9f8d/lib/rest/RequestHandler.js#L104

bsian03 (approved):
Not doing attachments since Game SDK is out of scope within Eris. Attachment changes were already documented in the previous PR

PR #1373, opened 2022-05-09
fix(deleteMessages): Properly pass through `reason` option

DonovanDMC:
Currently, the `reason` option is dropped within subsequent requests, or if only one message is provided.


bsian03 (approved):
not merging yet

PR #1379, opened 2022-05-30
Overall Consistency Fixes & Minor Feature Additions

DonovanDMC:
This is nowhere near done, I'm planning on combing through most classes & interfaces and fixing up things as I go - I'm just creating the pr now to open discussions


DonovanDMC:
I'm looking further into thread channels, and we're reporting them as having a few properties they don't have. I don't know if we're just okay with that, or if we want to do something about it. The only properties shared between all channels is `id`, `name`, `parent_id` & `type`. We're reporting `position` & `permission_overwrites` along with those, which is incorrect.

HeadTriXz:
If it's incorrect, it should be fixed.

DonovanDMC:
Store channels don't really exist anymore, so I'm mostly just dancing around store channels right now
I'd love for #1405 to go in before this (or at least with this) so store channels can be fully done away with

PR #1382, opened 2022-06-05
fix(ShardManager): refactor concurrency check, add option

abalabahaha:
- Reduce gateway calls
- Check bottiness

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -149,6 +149,7 @@ class Client extends EventEmitter {
             rest: {},
             restMode: false,
             seedVoiceConnections: false,
+            shardConcurrency: "auto",
             ws: {},
             reconnectDelay: (lastDelay, attempts) => Math.pow(attempts + 1, 0.7) * 20000
         }, options);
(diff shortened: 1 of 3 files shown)


lib/Client.js
```diff
@@ -418,7 +424,7 @@ class Client extends EventEmitter {
     */
     async connect() {
         try {
-            const data = await (this.options.maxShards === "auto" ? this.getBotGateway() : this.getGateway());
+            const data = await (this.options.maxShards === "auto" || (this.options.shardConcurrency === "auto" && this.bot) ? this.getBotGateway() : this.getGateway());
```
DonovanDMC, line 428: ~~this won't actually try concurrent sharding if the token isn't manually prefixed~~

Resolved; this is the intended behavior

abalabahaha: Sounds like a reasonable compromise given that:
- `concurrency: 1` is default behavior
- concurrency > 1 is a fairly exclusive permission
- `getBotGateway()` is restricted to bots only
- Forcing manual token prefixing is under discussion

HcgRandon:
toJSON of ShardManager needs to change lastConnect to buckets i missed this in my inital PR

DonovanDMC (approved):

PR #1387, opened 2022-06-16
fix(getMessages): Emit Error object in "error" event

frobinsonj:
Didn't really know how to title this, apologies if I got it wrong.

Error event `err` param should be type `Error` - https://github.com/abalabahaha/eris/blob/dev/index.d.ts#L711

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -2673,7 +2673,7 @@ class Client extends EventEmitter {
             try {
                 return new Message(message, this);
             } catch(err) {
-                this.emit("error", `Error creating message from channel messages\n${err.stack}\n${JSON.stringify(messages)}`);
+                this.emit("error", new Error(`Error creating message from channel messages\n${err.stack}\n${JSON.stringify(messages)}`));
                 return null;
             }
         });
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -2673,7 +2673,8 @@ class Client extends EventEmitter {
             try {
                 return new Message(message, this);
             } catch(err) {
-                this.emit("error", new Error(`Error creating message from channel messages\n${JSON.stringify(message)}`, {cause: err}));
+                err.message = `Error creating message from channel messages\n${JSON.stringify(message)}`;
+                this.emit("error", err);
                 return null;
             }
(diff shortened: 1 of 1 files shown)


DonovanDMC:
It would probably be better to construct the error object, then overwrite the stack
Or even just override the message of the caught error then emit it rather than both making a new error object & emitting the original stack

lib/Client.js
```diff
@@ -2673,7 +2673,7 @@ class Client extends EventEmitter {
             try {
                 return new Message(message, this);
             } catch(err) {
-                this.emit("error", `Error creating message from channel messages\n${err.stack}\n${JSON.stringify(messages)}`);
+                this.emit("error", new Error(`Error creating message from channel messages\n${err.stack}\n${JSON.stringify(messages)}`));
```
```suggestion
                err.message = `Error creating message from channel messages\n${JSON.stringify(messages)}`;
                this.emit("error", err);
```

DonovanDMC (approved):

DonovanDMC (changes requested):
[options.cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error) was added in node 16.9.0, we're currently on v10 (v14 at the latest), so we cannot use it

frobinsonj:
> [options.cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error) was added in node 16.9.0, we're currently on v10 (v14 at the latest), so we cannot use it

Went with your original suggestion. Thanks.

DonovanDMC (approved):

abalabahaha (changes requested):
Setting `.message` doesn't affect `.stack`. TBH I don't think the full JSON belongs in the error message. `warn` (also not great) and `debug` events are used to log JSON for other parsing errors

DonovanDMC:
We can slice the first line off the stack and remake it with the new message

or we can just ignore the original stack and construct a new error

bsian03:
Any other alternatives to this? Current solution (although error might be hard to read) is the best one imo

PR #1389, opened 2022-06-17
feat(shards): Support Non-Sequential Shard IDs

### LINKED ISSUE #1385: Posibility to set our own shardlList for client.
It could be interessting to set our own shard list so we have more controll on the sharding. We could also use different modules and implement clustering

HcgRandon:
This PR adds support for non-sequential shardsIDs to be used in eris. Currently I wrote it to override firstShardID and lastShardID if any numbers exist in the array. But I can see valid cases for the following as well:
- Throw an error if both firstShardID/lastShardID and shardIDs are defined
- firstShardID/lastShardID take priority and must be set to null for shardIDs to be used

Please let me know if I should change the logic to reflect one of these other states or to something entirely different if someone has a better idea.

I have performed basic testing of this and seems fine, but I want to test it more thoroughly before we merge.

Closes #1385

### DIFF
--- a/examples/sharding.js
+++ b/examples/sharding.js
@@ -4,6 +4,7 @@ const Eris = require("eris");
 const bot = new Eris("Bot TOKEN", {
     firstShardID: 0,
     lastShardID: 15,
+    shardIDs: [5, 9, 1, 4], // Overrides firstShardID and lastShardID above
     maxShards: 16,
     getAllUsers: false,
     intents: ["guilds", "guildMembers", "guildPresences"]
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -120,6 +120,7 @@ class Client extends EventEmitter {
     * @arg {Boolean} [options.restMode=false] Whether to enable getting objects over REST. Even with this option enabled, it is recommended that you check the cache first before using REST
     * @arg {Boolean} [options.seedVoiceConnections=false] Whether to populate bot.voiceConnections with existing connections the bot account has during startup. Note that this will disconnect connections from other bot sessions
     * @arg {Number | String} [options.shardConcurrency="auto"] The number of shards that can start simultaneously. If "auto" Eris will use Discord's recommended shard concurrency.
+    * @arg {Array<Number>} [options.shardIDs=null] An array of shard IDs to run for this client. If this is specified, the firstShardID and lastShardID options are ignored.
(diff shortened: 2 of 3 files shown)


HcgRandon:
#1385 related

lib/Client.js
```diff
@@ -120,6 +120,7 @@ class Client extends EventEmitter {
     * @arg {Boolean} [options.restMode=false] Whether to enable getting objects over REST. Even with this option enabled, it is recommended that you check the cache first before using REST
     * @arg {Boolean} [options.seedVoiceConnections=false] Whether to populate bot.voiceConnections with existing connections the bot account has during startup. Note that this will disconnect connections from other bot sessions
     * @arg {Number | String} [options.shardConcurrency="auto"] The number of shards that can start simultaneously. If "auto" Eris will use Discord's recommended shard concurrency.
+    * @arg {Array<Number>} [options.shardIDs=null] An array of shard IDs to run for this client. If this is specified, the firstShardID and lastShardID options are ignored.
```
```suggestion
    * @arg {Array<Number>} [options.shardIDs] An array of shard IDs to run for this client. If this is specified, the firstShardID and lastShardID options are ignored.
```

lib/Client.js
```diff
@@ -151,6 +152,7 @@ class Client extends EventEmitter {
             restMode: false,
             seedVoiceConnections: false,
             shardConcurrency: "auto",
+            shardIDs: null,
```
```suggestion
            shardIDs: [],
```

DonovanDMC (changes requested):

DonovanDMC (approved):
as far as I can tell, this works fine

examples/sharding.js
```diff
@@ -4,6 +4,7 @@ const Eris = require("eris");
 const bot = new Eris("Bot TOKEN", {
     firstShardID: 0,
     lastShardID: 15,
+    shardIDs: [5, 9, 1, 4], // Overrides firstShardID and lastShardID above
```
DonovanDMC, line 7: it might be worth noting shards will connect in the order specified
image

HcgRandon: It's a bit more nuanced then that. They are initialized in the order specified however concurrency will start them as fast as possible per bucket so the order is psuedo random in that case.

HcgRandon:
Marking this as a draft until #1384 is merged, as it makes more sense to just do these together rather then immediately deprecating options.shardIDs

PR #1390, opened 2022-06-18
feat(AutoModeration): Support Auto Moderation

DonovanDMC:


frobinsonj (commented):
Create/Modify/Delete Auto Moderation Rule endpoints support the `X-Audit-Log-Reason` header.

It should be possible to specify a reason. Not sure whether that should be put in the options or have a separate parameter.

DonovanDMC:
> Create/Modify/Delete Auto Moderation Rule endpoints support the `X-Audit-Log-Reason` header.
> 
> It should be possible to specify a reason. Not sure whether that should be put in the options or have a separate parameter.

Done, as far as I'm aware we're actively moving to providing reasons in the `options` object if possible

lib/gateway/Shard.js
```diff
@@ -606,6 +606,55 @@ class Shard extends EventEmitter {
 
     wsEvent(packet) {
         switch(packet.t) { /* eslint-disable no-redeclare */ // (╯°□°）╯︵ ┻━┻
```
HeadTriXz, line 608: Misses the `AUTO_MODERATION_ACTION_EXECUTION` event

index.d.ts
```diff
@@ -694,6 +743,9 @@ declare namespace Eris {
     selfVideo: boolean;
   }
   interface EventListeners {
```
HeadTriXz, line 758: Missing `autoModerationActionExecution`

HeadTriXz (changes requested):

index.d.ts
```diff
@@ -694,6 +756,10 @@ declare namespace Eris {
...
+    autoModerationActionExecution: [guild: Guild, action: AutoModerationAction];
```
HeadTriXz:
```suggestion
    autoModerationActionExecution: [guild: Guild, action: AutoModerationActionExecution];
```

HeadTriXz (changes requested):

index.d.ts
```diff
@@ -255,6 +262,61 @@ declare namespace Eris {
     permissions?: ApplicationCommandPermissions[];
   }
 
+  // Auto Moderation
+  interface AutoModerationAction {
+    metadata?: AutoModerationActionMetadata;
+    type: AutoModerationActionType;
+  }
+  interface AutoModerationActionExecution {
```
HeadTriXz, line 270: Shouldn't this be made a class?

HeadTriXz (changes requested):

index.d.ts
bsian03: Yeah probably. Or at least not make it snake_case, but a class would probably be better

DonovanDMC: Why should it be a class? I doesn't have any specific methods and there's no other reason for this to be a class
a camel cased object would work just fine

bsian03: Audit Log Entry is a class

HeadTriXz: Class or not, properties should be camelCase and you could turn channel_id into a channel.

DonovanDMC: > Audit Log Entry is a class

Existing useless classes are not a reason to make more useless classes

coolcalcacol: Was this not the exact conversion we had with the attachment class, where we ruled that it's probably better to just have it as a class?

DonovanDMC: Attachments are used in more than one place and are used in collections

PR #1391, opened 2022-06-28
fix(createGuild): use greater than or equal (>=)

lnpotter:
If it supports only 10 guilds, then we can use >= 10 instead of > 9


doomestee:
It wouldn't make any difference at all.

PR #1396, opened 2022-07-18
feat(ClientOptions): Support disabling all events

MertBhey:
Support disabling all events or just enabling some of them. Also typings for disableEvents


DonovanDMC:
`disableEvents` is specifically untyped as new events can appear out of nowhere, and disabling all events is not a thing we need to support, as it makes the library useless

PR #1397, opened 2022-07-22
feat(bans): Seconds Support

DonovanDMC:
discord/discord-api-docs#5219

This adds support for `delete_message_seconds`, and removes the range check, as we've moved away from checking things like that for end users.

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -354,8 +354,10 @@ class Client extends EventEmitter {
         if(reason !== undefined) {
             options.reason = reason;
         }
+        if(options.deleteMessageDays && !isNaN(options.deleteMessageDays) && !Object.hasOwn(options, "deleteMessageSeconds")) {
+            options.deleteMessageSeconds = options.deleteMessageDays * 24 * 60 * 60;
+        }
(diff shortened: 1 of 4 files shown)


DonovanDMC:
unsure if when transforming days to seconds, if we should consider an explicitly provided `undefined` a value we shouldn't override (current code treats it as such)

lib/Client.js
```diff
@@ -339,17 +339,27 @@ class Client extends EventEmitter {
...
+    banGuildMember(guildID, userID, options, reason) {
+        if(!options || typeof options !== "object") {
+            options = {
+                deleteMessageDays: options
+            };
+        }
+        if(reason !== undefined) {
+            options.reason = reason;
+        }
+        if(options.deleteMessageDays && !isNaN(options.deleteMessageDays) && !Object.hasOwn(options, "deleteMessageSeconds")) {
```
frobinsonj:
```suggestion
        if(options.deleteMessageDays && !isNaN(options.deleteMessageDays) && !options.hasOwnProperty("deleteMessageSeconds")) {
```

`Object.hasOwn()` was added in 16.9.0. 100% in favour of switching to 16.9.0, just wanted to point this out as we're still on 10.4.0.

PR #1398, opened 2022-07-23
refactor(createChannel): reduce overloads from 30 to 6

DonovanDMC:
This is functionally the same, but with much less duplicated code.


coolcalcacol:
Dono babes whys there a test file been committed 😳

PR #1401, opened 2022-08-01
fix(types): somes types error

flav-code:


index.d.ts
```diff
@@ -528,8 +528,8 @@ declare namespace Eris {
     description?: string;
     fields?: EmbedField[];
     footer?: EmbedFooterOptions;
-    image?: EmbedImageOptions;
-    thumbnail?: EmbedImageOptions;
+    image?: EmbedImage;
+    thumbnail?: EmbedImage;
```
DonovanDMC: This is not a typo. Undo this.

DonovanDMC (changes requested):

index.d.ts
DonovanDMC: Don't resolve this without actually doing what was asked.

bsian03 (changes requested):
Fix the comment please

bsian03 (approved):

PR #1402, opened 2022-08-03
fix(types): Wrong type in Message.edit()

flav-code:


DonovanDMC:
Pull requests for just types should be kept together, fold this into #1401 and make the title more generic

PR #1410, opened 2022-08-10
feat(gateway): Use New Resume URL

DonovanDMC:
A naïve attempt at using the new resume url.

### DIFF
--- a/lib/gateway/Shard.js
+++ b/lib/gateway/Shard.js
@@ -1880,6 +1880,10 @@ class Shard extends EventEmitter {
                     break;
                 } else {
                     this._resumeURL = `${packet.d.resume_gateway_url}?v=${Constants.GATEWAY_VERSION}&encoding=${Erlpack ? "etf" : "json"}`;
+                    
+                    if(this.client.options.compress) {
+                        this._resumeURL += "&compress=zlib-stream";
+                    }
                 }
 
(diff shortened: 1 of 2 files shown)


lib/gateway/Shard.js
```diff
@@ -1871,6 +1878,8 @@ class Shard extends EventEmitter {
                     */
                     super.emit("resume");
                     break;
+                } else {
+                    this._resumeURL = `${packet.d.resume_gateway_url}?v=${Constants.GATEWAY_VERSION}&encoding=${Erlpack ? "etf" : "json"}`;
```
TTtie: [zlib-stream compression handling](https://github.com/abalabahaha/eris/blob/4ea56e4/lib/Client.js#L445-L447) seems to be missed out:

```suggestion
                    this._resumeURL = `${packet.d.resume_gateway_url}?v=${Constants.GATEWAY_VERSION}&encoding=${Erlpack ? "etf" : "json"}`;
                    
                    if(this.client.options.compress) {
                        this._resumeURL += "&compress=zlib-stream";
                    }
```

TTtie (changes requested):

lib/gateway/Shard.js
DonovanDMC: As I said, naïve

PR #1335, opened 2022-02-09
feat(interactions): Localization

DonovanDMC:
Locale stuff for interactions, extracted from #1309

Draft until #1309 is merged

PR #1423, opened 2022-11-18
refactor(interactions): Interactions Refactor

bsian03:
Recreating #1309

PR #1309, opened 2021-11-28
refactor(interactions): Interactions Refactor

DonovanDMC:
Currently interactions can be hard to use due to the way the types have been made. This pull request completely rewrites most things surrounding interactions.

<details>
<summary>To Use In Projects</summary>
<br>
<code>
npm i DonovanDMC/eris#interactions-rewrite
</code>

I'd also recommend looking into [everything](https://github.com/DonovanDMC/eris/tree/everything) and [everything-v10](https://github.com/DonovanDMC/eris/tree/everything-v10)
</details>

PR #1424, opened 2022-11-18
feat(interactions): Localization

bsian03:
Recreating #1335


bsian03:
oh nvm this was folded into #1309 which is now #1423

PR #1426, opened 2022-11-18
feat(options): Gateway Options

bsian03:
Recreating #1384


flav-code:
Could you also add an option to choose the ip to connect with the websocket

bsian03:
Is there a valid use case with that? Discord's gateway isn't designed in a way where it's easy to have a single shard distribute and forward data onto separate connecting clients
P.s. this is on the back burner until all discord features are merged first

flav-code:
If you are gateway ratelimit you can change ip

bsian03:
Not a valid use case, if you are rate limited, you should wait until it has expired

PR #1438, opened 2022-12-15
Removed editGuildVanity

Larsundso:
The vanity endpoint was disabled for bots
(PATCH not GET)
```
{
    "message": "Bots cannot use this endpoint",
    "code": 20001
}
```

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -1677,18 +1677,6 @@ class Client extends EventEmitter {
     }
 
     /**
-    * Modify a guild's vanity code
-    * @arg {String} guildID The ID of the guild
-    * @arg {String?} code The new vanity code
-    * @returns {Promise<Object>}
-    */
-    editGuildVanity(guildID, code) {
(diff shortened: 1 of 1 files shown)


lib/Client.js
```diff
@@ -1676,18 +1676,6 @@ class Client extends EventEmitter {
...
     }
 
-    /**
-    * Modify a guild's vanity code
-    * @arg {String} guildID The ID of the guild
-    * @arg {String?} code The new vanity code
-    * @returns {Promise<Object>}
-    */
-    editGuildVanity(guildID, code) {
-        return this.requestHandler.request("PATCH", Endpoints.GUILD_VANITY_URL(guildID), true, {
```
coolcalcacol, line 1686: remove the endpoint too?

DonovanDMC: `getGuildVanity` should still work
https://github.com/abalabahaha/eris/blob/e43418dae558a8fc7108fbfd820ff6cd8c6a2db3/lib/Client.js#L2614-L2621

bsian03: *does

PR #1440, opened 2022-12-30
Add tag properties to User

ghost:
It's util.

### DIFF
--- a/lib/structures/User.js
+++ b/lib/structures/User.js
@@ -83,6 +83,10 @@ class User extends Base {
     get mention() {
         return `<@${this.id}>`;
     }
+    
+    get tag() {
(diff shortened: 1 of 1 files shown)


DonovanDMC:
#1231 #1019 #925 #504 #325 #252

That makes 7

DonovanDMC:
If you really want this in your own codebase, just put this somewhere before you start using Eris
```js
Object.defineProperty(Eris.Member.prototype, "tag", {
    get() { return `${this.username}#${this.discriminator}`; },
  });

Object.defineProperty(Eris.User.prototype, "tag", {
  get() { return `${this.username}#${this.discriminator}`; },
});
```

PR #1446, opened 2023-02-07
feat(Guild): support addGuildMember

ebroder:
Not sure how much explanation is required here, but I'm looking to switch an application of mine to use Eris and this is an endpoint that I'm using today.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -868,6 +868,12 @@ declare namespace Eris {
   }
 
   // Guild
+  interface AddGuildMemberOptions {
+    nick?: string;
+    roles?: string[];
+    deaf?: boolean;
+    mute?: boolean;
+  }
   interface CreateGuildOptions {
     afkChannelID?: string;
     afkTimeout?: number;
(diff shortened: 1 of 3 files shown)


coolcalcacol:
Lol.

jaw0r3k:
You forgot to add typings for `addUser`

ebroder:
> You forgot to add typings for `addUser`

Good flag, thanks. Should be fixed now.

index.d.ts
```diff
@@ -2802,6 +2809,7 @@ declare namespace Eris {
     widgetEnabled?: boolean | null;
     constructor(data: BaseData, client: Client);
     addDiscoverySubcategory(categoryID: string, reason?: string): Promise<DiscoverySubcategoryResponse>;
+    addMember(userID: string, accessToken: string, options?: AddGuildMemberOptions): Promise<void>;
```
jaw0r3k, line 2819:
```suggestion
    addUser(userID: string, accessToken: string, options?: AddGuildMemberOptions): Promise<void>;
```

abalabahaha:
`addMember` would be more consistent with `addGuildMember`, did I miss something?

ebroder:
Hmm. Yeah sure, that seems reasonable. I think in my head I was naming the method after the noun that it acts on (since you pass in a _user_ ID), but yeah I agree that parity makes more sense. Pushed a fix (and a catchup to `dev`)

abalabahaha (approved):

abalabahaha:
Yeah member/user is a bit more iffy here. Though I think more of the lib has endpoint-related names, so let's go with member for now.

PR #1448, opened 2023-03-04
feat(voice): support resuming voice connection

james58899:
https://discord.com/developers/docs/topics/voice-connections#resuming-voice-connection

The Discord voice gateway will disconnect with code 1006 after a period of time, but the UDP connection is still normal.
We can use **[Opcode] 7 Resume** to restore the websocket connection without affecting other states.
If use **[Opcode] 0 Identify** to reconnect the voice gateway will disconnect with code 4006.

[Opcode]: https://discord.com/developers/docs/topics/opcodes-and-status-codes#voice-voice-opcodes


abalabahaha:
Need to test, but I think 1000 is still a normal close where reconnect isn't wanted, so `=== 1006` may fit better than `< 4000`

james58899:
In discord.js it is `code === 4_015 || code < 4_000`[^1], because discord usually uses 4014 to disconnect, but I only see 1006 in my test, so I think changing `< 4000` to `=== 1006` is ok.

[^1]: https://github.com/discordjs/discord.js/blob/dc142c47e4021d6d37712453cc2fad6e4fe4e9f7/packages/voice/src/networking/Networking.ts#L357

bsian03 (changes requested):
Need to document `resuming` as a property, but lgtm otherwise. Unless @abalabahaha has any thoughts about that 1006 handling

bsian03 (approved):

PR #1458, opened 2023-06-10
Type Error CommandInteraction.js

YalphaFR:
Change type Boolean to Number for ephemeral message flags


bsian03:
Next time please combine all the same issues into one PR, added in #1457

PR #1459, opened 2023-06-10
Support new username system

conorwastakenwastaken:
I've probably forgotten to do something

### DIFF
--- a/lib/structures/Guild.js
+++ b/lib/structures/Guild.js
@@ -172,7 +172,7 @@ class Guild extends Base {
                 if(!this.members.get(presence.user.id)) {
                     let userData = client.users.get(presence.user.id);
                     if(userData) {
-                        userData = `{username: ${userData.username}, id: ${userData.id}, discriminator: ${userData.discriminator}}`;
+                        userData = `{username: ${userData.username}, id: ${userData.id + (userData.discriminator === "0" ? "" : `, discriminator: ${userData.discriminator}`)}}`;
                     }
                     client.emit("debug", `Presence without member. ${presence.user.id}. In global user cache: ${userData}. ` + JSON.stringify(presence), this.shard.id);
                     continue;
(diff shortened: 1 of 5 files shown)


lib/structures/Guild.js
```diff
@@ -172,7 +172,7 @@ class Guild extends Base {
                 if(!this.members.get(presence.user.id)) {
                     let userData = client.users.get(presence.user.id);
                     if(userData) {
-                        userData = `{username: ${userData.username}, id: ${userData.id}, discriminator: ${userData.discriminator}}`;
+                        userData = `{username: ${userData.username}, id: ${userData.id + (userData.discriminator === "0" ? "" : `, discriminator: ${userData.discriminator}`)}}`;
```
bsian03: Not sure what you meant to do here? You're trying to edit user ID with discrim

lib/structures/User.js
```diff
@@ -178,7 +182,7 @@ class User extends Base {
             "avatar",
             "banner",
             "bot",
-            "discriminator",
```
bsian03, line 181: It's worth keeping this

bsian03 (changes requested):

lib/structures/Guild.js
conorwastakenwastaken: No that's just a string for a debug log. This change just omits the users discrim from the log since it's not necessary anymore. Although this change isn't really necessary either, but if I didn't do this then it'd send `#0` inside the log

lib/structures/User.js
conorwastakenwastaken: Yeah I forgot that bots weren't included in the new usernames system lol

lib/structures/Guild.js
bsian03: Oh nevermind I see it now. It just looks a little messy imo. Would it be better if the whole discrim handling was done as its own separate thing either as an if or by doing `"id: ..." + discrim ternary` here?

bsian03: Or just let them see "0" anyways. It should be clear that the user was migrated to the new username system

conorwastakenwastaken: Reverted sir

bsian03 (approved):

PR #1462, opened 2023-08-18
Add color convert and Colors constants.

sulvii:
## Changes

- **Color Constants**: Added some basic colors to the Constants
- **ColorConverter**: Added `hsl`, `hsv`, `hex`, `rgb`, `cmyk` methods to convert them to Discord supported `Byte10`. This method utilizes the "color-convert" module. The module has been installed as an "optional" dependency since it's not a critical feature.

## Review Request

I kindly request your review of the changes. Thank you for your time.


bsian03:
This is not within the scope of Eris. If people want to use this, they should install the module in their own projects

PR #1465, opened 2024-01-05
 fix(Client): wrong request url in editGuildWidget

MertBhey:
Fixes wrong endpoint used in edtGuildWidget, it was requesting to widget.json instead of widget for some reason

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2450,7 +2450,7 @@ declare namespace Eris {
     editGuildVanity(guildID: string, code: string | null): Promise<GuildVanity>;
     editGuildVoiceState(guildID: string, options: VoiceStateOptions, userID?: string): Promise<void>;
     editGuildWelcomeScreen(guildID: string, options: WelcomeScreenOptions): Promise<WelcomeScreen>;
-    editGuildWidget(guildID: string, options: Widget): Promise<Widget>;
+    editGuildWidget(guildID: string, options: Widget & { channelID?: string }): Promise<Widget>;
     editMessage(channelID: string, messageID: string, content: MessageContentEdit): Promise<Message>;
     /** @deprecated */
     editNickname(guildID: string, nick: string, reason?: string): Promise<void>;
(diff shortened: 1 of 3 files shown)


bsian03 (changes requested):
Worth documenting `options` params, also supports `reason` passing to audit log

lib/Client.js
```diff
@@ -1779,10 +1779,11 @@ class Client extends EventEmitter {
     * Modify a guild's widget
     * @arg {String} guildID The ID of the guild
     * @arg {Object} options The widget object to modify (https://discord.com/developers/docs/resources/guild#modify-guild-widget)
```
bsian03, line 1781: Can you document channelID and enabled? Reassign `channelID` to `channel_id` inside the method ensuring passthrough backwards compatibility

lib/Client.js
```diff
@@ -1779,10 +1779,16 @@ class Client extends EventEmitter {
...
-    * @arg {String} [options.reason] The reason to be displayed in audit logs
+    * @arg {Boolean} [options.enabled] Whether the guild widget is enabled
+    * @arg {String?} [options.channelID] The channel ID for the guild widget
+    * @arg {String?} [options.reason] The reason to be displayed in audit logs
     * @returns {Promise<Object>} A guild widget object
     */
     editGuildWidget(guildID, options) {
+        if(options.channelID) {
```
bsian03: Check specifically if not undefined
```suggestion
        if(options.channelID !== undefined) {
```

lib/Client.js
```diff
@@ -1779,10 +1779,16 @@ class Client extends EventEmitter {
...
+            options.channel_id = options.channelID;
+            delete options.channelID;
```
```suggestion
```

bsian03 (changes requested):

index.d.ts
```diff
@@ -1070,6 +1070,7 @@ declare namespace Eris {
   interface Widget {
     channel_id?: string;
     enabled: boolean;
```
bsian03, line 1072: add channelID definition here, not in the parameter

bsian03 (approved):

PR #1469, opened 2024-01-14
Various docs/typings enhancements and corrections

bsian03:
Combines #1368 #1425 #1429

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2488,8 +2488,7 @@ declare namespace Eris {
     createGuildTemplate(guildID: string, name: string, description?: string | null): Promise<GuildTemplate>;
     createInteractionResponse(interactionID: string, interactionToken: string, options: InteractionOptions, file?: FileContent | FileContent[]): Promise<void>;
     createMessage(channelID: string, content: MessageContent, file?: FileContent | FileContent[]): Promise<Message>;
-    createRole(guildID: string, options?: RoleOptions, reason?: string): Promise<Role>;
-    createRole(guildID: string, options?: Role, reason?: string): Promise<Role>;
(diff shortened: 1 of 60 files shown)


index.d.ts
```diff
@@ -3310,11 +3347,11 @@ declare namespace Eris {
     editedTimestamp?: number;
     embeds: Embed[];
     flags: number;
-    guildID: T extends GuildTextableWithThread ? string : undefined;
+    guildID: T extends GuildTextableWithThreads ? string : void;
```
oathompsonjones: Why has this been changed to `void`? `void` should only be used as a function return type, as giving a property a type of `void` means that the only value you can assign to that property is `undefined`, so you should use `undefined` instead to avoid confusion. If this is meant to mean that the property does not exist, then you should use the `never` type. See [here](https://www.typescriptlang.org/play?#code/DYUwLgBAbg9glgEwCogM5gFzXgg3AKFkRXQgF4IAGAo5NSCgIksZpxIYgDMBDYVEG2L1yEANoBdIXVIUA3gF9pHUQAoAlFlrkAfBACuAOwQgucQyDyF2Iiof3Bgy2weOnzlgvlCQLUEABOHFh+gQShQS7U+BEqTCzhIP6Rstx8AonJceJSMUmB2YqZBS4aWji6riZmFlaxLvaOxSmcRtUeeEA) for an example.

bsian03: Simply because I had been awake for over 18 hours at that point :) was intending to type it as never since I thought there was specific undefined check handling but turns out there isn't.

```suggestion
    guildID: T extends GuildTextableWithThreads ? string : undefined;
```

PR #1470, opened 2024-01-14
feat(Reactions): Support super reactions

conorwastakenwastaken:
## Super Reactions!

Adds support for super reactions! Pretty sure Discord is still in the process of adding some of it from the docs, so some of the properties Ive used are from open PRs on the Discord API repo

- discord/discord-api-docs#6375
- discord/discord-api-docs#6056

Also adds support for specifying a reaction type when using `Message#getReaction()` and all `getMessageReaction()` functions

- discord/discord-api-docs#6377

## Changes/Additions

- `burst_colors`, `count_details` & `me_burst` added to `Message#reactions` reaction objects
- `messageReactionAdd` & `messageReactionRemove` events send additional `burst` (boolean) parameter to bound function
- `type` option added to the `getReaction()` function within Message, as well as the `getMessageReaction()` function within Client, DMChannel & GuildTextableChannel, 

### Testing:

<details>
<summary>MESSAGE_REACTION_ADD & MESSAGE_REACTION_REMOVE rawWS screenshots</summary>

#### Message Reaction Add (Normal Emoji)
image

#### Message Reaction Remove (Normal Emoji)
image

#### Message Reaction Add (Burst Emoji)
image

#### Message Reaction Remove (Burst Emoji)
image

***Note: `burst_colors` is never sent with MESSAGE_REACTION_REMOVE, even when it's a burst reaction, so I'll leave that property out completely rather than just sending an empty array to the event.***

</details>

<details>
<summary>Message object with burst reaction</summary>

image

This screenshot shows a message with 1 normal reaction, and 1 burst reaction. I'll add more screenshots here when I've finished the todo list and can continue testing 👍🏻 

</details>


bsian03:
Has someone tested this? I can't reliably test this myself

bsian03 (changes requested):
Please lint, merge conflict resolution is invalid

conorwastakenwastaken:
> Has someone tested this? I can't reliably test this myself

This PR is still a draft cuz Discord haven't updated their docs yet to match all of the props I've used here. Probably best to test it when both of those PRs are merged just in case they change anything else 👍🏻

bsian03:
I personally don't think it's worth waiting until Discord confirms docs. Also we previously had stuff which was undocumented (can't remember which ones exactly), and if there are any changes, we can just update accordingly

lib/gateway/Shard.js
```diff
@@ -1136,8 +1148,9 @@ class Shard extends EventEmitter {
                  * @prop {String?} emoji.id The emoji ID (null for non-custom emojis)
                  * @prop {String} emoji.name The emoji name
                  * @prop {Member | Object} reactor The member, if the reaction is in a guild. If the reaction is not in a guild, this will be an object with an `id` key. No other property is guaranteed
+                 * @prop {Boolean} burst Whether the reaction is a super reaction
                  */
-                this.emit("messageReactionAdd", message, packet.d.emoji, member || {id: packet.d.user_id});
+                this.emit("messageReactionAdd", message, packet.d.emoji, member || {id: packet.d.user_id}, !!packet.d.burst);
```
bsian03: This is a boolean anyways, the `!!` isn't needed

bsian03 (commented):
There are some props which aren't documented in the PRs, I've tried to ask the api team what's going on with them but haven't actually received a reply (yes i know that contradicts what i said in my earlier comment)

lib/gateway/Shard.js
conorwastakenwastaken: `packet.d.burst` is only sent if it's a super reaction, according to docs. I did it this way so it'd be either true or false. Did I misread the docs? To me it sounded like that prop would be undefined if it's not a super reaction lol (I wouldnt put it past Discord to do something like that)

conorwastakenwastaken: Ive tested this btw (without the `!!`), and Discord does still send it when it's false. I misunderstood their docs lol, but made the change now 👍🏻

lib/gateway/Shard.js
```diff
@@ -1108,13 +1108,25 @@ class Shard extends EventEmitter {
                     const reaction = packet.d.emoji.id ? `${packet.d.emoji.name}:${packet.d.emoji.id}` : packet.d.emoji.name;
                     if(message.reactions[reaction]) {
                         ++message.reactions[reaction].count;
+                        ++message.reactions[reaction].count_details[packet.d.burst ? "burst" : "normal"];
                         if(packet.d.user_id === this.client.user.id) {
-                            message.reactions[reaction].me = true;
+                            if(packet.d.burst) {
+                                message.reactions[reaction].me_burst = true;
+                            } else {
+                                message.reactions[reaction].me = true;
```
conorwastakenwastaken: @bsian03 Should I **always** make `.me` true, even when it's not a super reaction? And then obviously `me_burst` would also be true when it's a super reaction. Just for backwards compatability reasons really. What d'ya think?

bsian03:
I'll take a further look at this once I get my PR done

lib/gateway/Shard.js
bsian03: Yes, do this. Would be better to check against whether it's burst or not (or by checking type), if applicable

bsian03 (changes requested):
Missing `type` in both reaction add and reaction remove

index.d.ts
```diff
@@ -3771,7 +3777,7 @@ declare namespace Eris {
     messageReference: MessageReference | null;
     pinned: boolean;
     prefix?: string;
-    reactions: { [s: string]: { count: number; me: boolean } };
+    reactions: { [s: string]: { burst_colors: string[], count: number, count_details: { burst: number, normal: number }, me: boolean, me_burst: boolean, type: number; } };
```
bsian03: Needs to reference ReactionTypes union

bsian03 (changes requested):

index.d.ts
conorwastakenwastaken: Oops mb, fixed 👍🏻

index.d.ts
```diff
@@ -3776,7 +3782,7 @@ declare namespace Eris {
     messageReference: MessageReference | null;
     pinned: boolean;
     prefix?: string;
-    reactions: { [s: string]: { count: number; me: boolean } };
+    reactions: { [s: string]: { burst_colors: string[], count: number, count_details: { burst: number, normal: number }, me: boolean, me_burst: boolean, type: ReactionTypes; } };
```
bsian03: Would prefer this to be interfaced

bsian03 (approved):

PR #1471, opened 2024-01-14
feat(User): Support avatar decorations

conorwastakenwastaken:
Adds support for user avatar decorations. Avatar decoration properties will also exist within members, to be consistent with other user properties that do the same.

You may notice that some of the properties aren't the same as what's shown on the Discord docs rn. Discord have changed the field names and CDN endpoint for avatar decorations apparently so I used discord/discord-api-docs#6464 to make this PR so I wouldnt have to potentially go back and make changes lol

I'm sure I probably made a mistake in TS docs so any help would be great if you happen to spot any

### DIFF
--- a/lib/structures/User.js
+++ b/lib/structures/User.js
@@ -195,6 +195,7 @@ class User extends Base {
         return super.toJSON([
             "accentColor",
             "avatar",
+            "avatarDecorationData",
             "banner",
             "bot",
             "discriminator",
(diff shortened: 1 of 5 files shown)


bsian03 (changes requested):
Please lint, merge conflict resolution is invalid

lib/structures/User.js
bsian03, line 1: Include in JSON please

bsian03 (changes requested):

lib/structures/User.js
conorwastakenwastaken: Done 👍🏻

bsian03 (approved):

PR #1475, opened 2024-01-17
feat(Presence): Allow bots to set custom status

conorwastakenwastaken:
[Docs reference](https://github.com/discord/discord-api-docs/pull/6345)

~~WIP. Needs reviewing by a collab to tell me whether they prefer Eris handling it this way, or whether they want me to add an `activities[].state` prop that users specifically need to use to set custom statuses.~~

## Changes
 - Added `activities[].state` prop to use when setting a custom status
 - Removed `BotActivityType` type from TSdocs, since it is now the same as `ActivityType`
 - ~~Validate the `type` prop on all activity objects, rather than just the first one~~

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1520,6 +1520,7 @@ declare namespace Eris {
   }
   interface ActivityPartial<T extends ActivityType> {
     name: string;
+    state?: string;
     type?: T;
     url?: string;
   }
--- a/index.d.ts
+++ b/index.d.ts
@@ -155,7 +155,7 @@ declare namespace Eris {
 
   // Presence/Relationship
   type ActivityFlags = Constants["ActivityFlags"][keyof Constants["ActivityFlags"]];
-  type ActivityType = Constants["ActivityTypes"];
+  type ActivityType = Constants["ActivityTypes"][keyof Constants["ActivityTypes"]];
   type FriendSuggestionReasons = { name: string; platform_type: string; type: number }[];
   type SelfStatus = Status | "invisible";
   type Status = "online" | "idle" | "dnd";
--- a/index.d.ts
+++ b/index.d.ts
@@ -155,8 +155,7 @@ declare namespace Eris {
 
   // Presence/Relationship
   type ActivityFlags = Constants["ActivityFlags"][keyof Constants["ActivityFlags"]];
-  type ActivityType = BotActivityType | Constants["ActivityTypes"]["CUSTOM"];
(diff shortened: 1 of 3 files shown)


lib/Client.js
```diff
@@ -2173,6 +2173,16 @@ class Client extends EventEmitter {
             activities = [activities];
         }
         if(activities !== undefined) {
+            if(activities.length > 0) {
+                activities = activities.map((a) => {
+                    if(a.hasOwnProperty("type")) {
+                        a.type = a.url ? 1 : 0;
+                    } else if(a.type === 4) {
+                        a.state = a.name;
+                    }
```
bsian03: Code here will always result in the type being set to 0 or 1 if a type is provided, even if it is `4`. Would rather see a state prop to be in line with Discord

bsian03 (changes requested):

lib/Client.js
```diff
@@ -2181,15 +2182,10 @@ class Client extends EventEmitter {
...
-            activities = [];
-        } else if(activities && !Array.isArray(activities)) {
+        if(activities && !Array.isArray(activities)) {
             activities = [activities];
         }
-        if(activities !== undefined) {
-            this.presence.activities = activities;
-        }
-
+        this.presence.activities = activities || [];
```
bsian03: If `activities` is undefined, this should just be left alone. There shouldn't be a change in activities. 

TLDR: previous handling of `null` should continue having that behaviour. Your change would also be breaking

The PR should just add support for custom status

bsian03 (changes requested):

lib/Client.js
conorwastakenwastaken: Tfw the TLDR is longer than the original message
I reverted it back👍🏻

index.d.ts
```diff
@@ -155,8 +155,7 @@ declare namespace Eris {
 
   // Presence/Relationship
   type ActivityFlags = Constants["ActivityFlags"][keyof Constants["ActivityFlags"]];
-  type ActivityType = BotActivityType | Constants["ActivityTypes"]["CUSTOM"];
-  type BotActivityType = Constants["ActivityTypes"][Exclude<keyof Constants["ActivityTypes"], "CUSTOM">];
+  type ActivityType = Constants["ActivityTypes"];
```
```suggestion
  type ActivityType = Constants["ActivityTypes"][keyof Constants["ActivityTypes"]];
```
was referring to the object itself before, not the values

bsian03 (changes requested):

bsian03 (approved):

PR #1488, opened 2024-04-07
fix: make file argument optional for createMessage method

flav-code:


bsian03:
Thanks, this is why I wanted people to test my PR

PR #1493, opened 2024-04-18
feat: Support polls

conorwastakenwastaken:
## Support Polls!

Refs:
- [Docs](https://discord.com/developers/docs/resources/poll)
- discord/discord-api-docs#6746)

## Additions

- `Message#poll` property
- `sendPolls` permission (`1n` << `49n`)
- `guildMessagePolls` (`1` << `24`) & `directMessagePolls` (`1` << `25`) intents
- New `getPollAnswerVoters` function (in GuildTextableChannel & Client)
- New `endPoll` function (in GuildTextableChannel & Client) -- Note: You cannot end polls from other users
- New `messagePollVoteAdd` & `messagePollVoteRemove` events
- Documented `poll` option when creating messages (`createMessage` & `executeWebhook`)

## To Do

- [x] TSdoc
- [x] Complete documenting `poll` option for interaction responses
- [x] Listen for and document `MESSAGE_POLL_VOTE_ADD` & `MESSAGE_POLL_VOTE_REMOVE` gateway events
- [x] Support `poll` in message create/delete/update events where applicable
- [x] Add `getAnswerVoters` function (as well as the endpoint)
- [x] Add `endPoll` function (as well as the endpoint)


conorwastakenwastaken:
@bsian03 While it's early, could you advise on whether you want an entire Poll structure, whether you just want us to support camelCase poll properties for creating them, or whether you want me to leave it the way I've done it atm

bsian03:
creating polls in dms is dumb
Leaving them snake_case will be fine. At some point in the not-too-soon future I want to sort out when we should create structs, camelCase or snake_case properties, but that's for another day

conorwastakenwastaken:
> creating polls in dms is dumb Leaving them snake_case will be fine. At some point in the not-too-soon future I want to sort out when we should create structs, camelCase or snake_case properties, but that's for another day

I agree it is dumb, but it's still possible for some reason. And okay sounds good 👍🏻

index.d.ts
```diff
@@ -2966,6 +3003,7 @@ declare namespace Eris {
     getNitroStickerPacks(): Promise<{ sticker_packs: StickerPack[] }>;
     getOAuthApplication(appID?: string): Promise<OAuthApplicationInfo>;
     getPins(channelID: string): Promise<Message[]>;
+    getPollAnswerVoters(channelID: string, messageID: string, answerID: string, options?: GetRESTGuildMembersOptions): Promise<User[]>;
```
conorwastakenwastaken: > `options?: GetRESTGuildMembersOptions`

@bsian03 This is temporary until you tell me what you want me to do w/ it. These options expect the same options. I can either create a duplicate options interface with a different name (seems p dumb), or you can give me something to rename `GetRESTGuildMembersOptions` to so it doesn't look odd using that interface name in this context, or I can just leave it as is

lib/gateway/Shard.js
```diff
@@ -1258,6 +1260,64 @@ class Shard extends EventEmitter {
...
+                    }
+                }
+                /**
+                 * Fired when a user votes on a poll. If the poll allows multiple selection, one event will be sent per answer
+                 * @event Client#messagePollVoteAdd
+                 * @prop {Message | Object} message The message object. If the message is not cached, this will be an object with `id` and `channel` keys. If the channel is not cached, channel key will be an object with an id and a `guild` key with only an id. No other property is guaranteed
+                 * @prop {User | Object} user The user who voted on the poll. If the user is not cached, this will be an object with only an `id` key
+                 * @prop {Number} answerID The ID of the answer
+                 */
+                this.emit("messagePollVoteAdd", message, user, packet.d.answer_id);
```
bsian03: Can we add/remove answer counts if the message is cached please

index.d.ts
bsian03: I'm going to say create a duplicate interface (non extending) as all other instances of interfaces containing a combination of after/before/limit etc do not extend each other, instead are listed separately in each relevant interface

bsian03 (changes requested):

index.d.ts
conorwastakenwastaken: Okidoki. Done

lib/gateway/Shard.js
conorwastakenwastaken: Yeah nw. That should be okay 👍🏻

index.d.ts
```diff
@@ -1574,6 +1582,34 @@ declare namespace Eris {
...
+  }
+  interface PollAnswer {
+    answer_id: number;
+    poll_media: PollMedia;
+  }
+  interface PollCreateOptions extends Omit<Poll, "expiry" | "results"> {
+    duration: number;
+  }
+  interface PollMedia {
+    emoji?: Partial<PartialEmoji>;
```
bsian03: Why the partial type here?

index.d.ts
```diff
@@ -1574,6 +1582,34 @@ declare namespace Eris {
     filename?: string;
     id: string | number;
   }
+  interface Poll {
+    allow_multiselect: boolean;
+    answers: PollAnswer[];
+    expiry: string | null;
+    layout_type: number;
+    question: PollMedia;
```
bsian03: This should only read from `text`, or rather `emoji` should be removed. Also `text` is required

index.d.ts
```diff
@@ -1574,6 +1582,34 @@ declare namespace Eris {
     filename?: string;
     id: string | number;
   }
+  interface Poll {
+    allow_multiselect: boolean;
+    answers: PollAnswer[];
+    expiry: string | null;
+    layout_type: number;
```
bsian03: This should be enumerated

bsian03 (changes requested):
`expiry` needs to be properly handled

index.d.ts
conorwastakenwastaken: Fixed. `emoji` is used in poll answers so I've kept that but removed the Partial type 👍🏻

index.d.ts
```diff
@@ -1574,6 +1582,34 @@ declare namespace Eris {
     filename?: string;
     id: string | number;
   }
+  interface Poll {
+    allow_multiselect: boolean;
+    answers: PollAnswer[];
+    expiry: string | null;
+    layout_type: 1;
```
bsian03: This still needs to be enumerated, not as a constant value. A new constants entry needs to be created

lib/structures/Message.js
```diff
@@ -40,6 +40,7 @@ const User = require("./User");
  * @prop {String?} messageReference.guildID The ID of the guild this message was crossposted from
  * @prop {String?} messageReference.messageID The ID of the original message this message was crossposted from
  * @prop {Boolean} pinned Whether the message is pinned or not
+ * @prop {Object?} poll A poll object. See [the official Discord API documentation entry](https://discord.com/developers/docs/resources/poll#poll-object) for object structure
```
bsian03: Worth noting that `Date` is an int, not ISO8601 as mentioned in Discord docs

bsian03 (changes requested):
lgtm otherwise

lib/Client.js
```diff
@@ -2455,6 +2456,16 @@ class Client extends EventEmitter {
         });
     }
 
+    /**
+     * Immediately end a poll. Note: You cannot end polls from other users.
```
```suggestion
     * Immediately end a poll. Note: You cannot end polls created by another user.
```

lib/gateway/Shard.js
```diff
@@ -1258,6 +1260,80 @@ class Shard extends EventEmitter {
...
+                        id: packet.d.message_id,
+                        channel: channel || {id: packet.d.channel_id}
+                    };
+                    if(packet.d.guild_id) {
+                        message.guildID = packet.d.guild_id;
+                        if(!message.channel.guild) {
+                            message.channel.guild = {id: packet.d.guild_id};
+                        }
+                    }
+                } else if(message.poll.results.answer_counts.find((answer) => answer.id === packet.d.answer_id)) {
```
bsian03: This will error when a vote is added/removed (respectively) while the poll is still active (no `answer_counts` during an active poll)

bsian03 (changes requested):

lib/gateway/Shard.js
```diff
@@ -1258,6 +1260,84 @@ class Shard extends EventEmitter {
...
+                        message.guildID = packet.d.guild_id;
+                        if(!message.channel.guild) {
+                            message.channel.guild = {id: packet.d.guild_id};
+                        }
+                    }
+                } else if(message.poll.results && message.poll.results.answer_counts) {
+                    if(message.poll.results.answer_counts.find((answer) => answer.id === packet.d.answer_id)) {
+                        message.poll.results.answer_counts.find((answer) => answer.id === packet.d.answer_id).count++;
+                        if(packet.d.user_id === this.client.user.id) {
+                            message.poll.results.answer_counts.find((answer) => answer.id === packet.d.answer_id).me_voted = true;
```
bsian03: Can you remove the duplicate `find`s here please?

lib/gateway/Shard.js
```diff
@@ -1258,6 +1260,84 @@ class Shard extends EventEmitter {
...
+                        message.guildID = packet.d.guild_id;
+                        if(!message.channel.guild) {
+                            message.channel.guild = {id: packet.d.guild_id};
+                        }
+                    }
+                } else if(message.poll.results && message.poll.results.answer_counts) {
+                    if(message.poll.results.answer_counts.find((answer) => answer.id === packet.d.answer_id)) {
+                        message.poll.results.answer_counts.find((answer) => answer.id === packet.d.answer_id).count--;
+                        if(packet.d.user_id === this.client.user.id) {
+                            message.poll.results.answer_counts.find((answer) => answer.id === packet.d.answer_id).me_voted = false;
```
bsian03: As well as here

bsian03 (changes requested):

lib/gateway/Shard.js
conorwastakenwastaken: good?

bsian03 (approved):

PR #1497, opened 2024-04-24
feat(ThreadChannel): Thread member pagination

conorwastakenwastaken:
## Add thread member pagination options

Ref:
- discord/discord-api-docs#5834

## Additions/Changes
- Added `after`, `limit` & `withMember` options to `Client#getThreadMembers()` and `ThreadChannel#getMembers()`
- Typed `Client#getThreadMember()`
- Fixed `Client#getThreadMember()` function not sending the `with_member` option properly
- Fixed `Client#getThreadMember()` & `ThreadChannel#getMember()` js doc

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -3505,7 +3505,7 @@ class Client extends EventEmitter {
      * @arg {Boolean} [options.withMember] Whether to include the guild member object for each member of the thread
      * @returns {Promise<Array<ThreadMember>>}
      */
-    getThreadMembers(channelID, options) {
+    getThreadMembers(channelID, options = {}) {
         return this.requestHandler.request("GET", Endpoints.THREAD_MEMBERS(channelID), true, {
             after: options.after,
             limit: options.limit,
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -3499,10 +3499,18 @@ class Client extends EventEmitter {
     /**
      * Get a list of members that are part of a thread channel
      * @arg {String} channelID The ID of the thread channel
+     * @arg {Object} [options] Options for the request
+     * @arg {String} [options.after] Get members after this user ID
+     * @arg {Number} [options.limit=100] The maximum number of thread members to fetch (1-100)
+     * @arg {Boolean} [options.withMember] Whether to include the guild member object for each member of the thread
      * @returns {Promise<Array<ThreadMember>>}
      */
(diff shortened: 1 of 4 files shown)


lib/Client.js
```diff
@@ -3487,22 +3487,30 @@ class Client extends EventEmitter {
...
      * @arg {String} channelID The ID of the thread channel
+     * @arg {Object} [options] Options for the request
+     * @arg {String} [options.after] Get members after this user ID
+     * @arg {Number} [options.limit=100] The maximum number of thread members to fetch (1-100)
+     * @arg {Boolean} [options.withMember] Whether to include the guild member object for each member of the thread
      * @returns {Promise<Array<ThreadMember>>}
      */
-    getThreadMembers(channelID) {
-        return this.requestHandler.request("GET", Endpoints.THREAD_MEMBERS(channelID), true).then((members) => members.map((member) => new ThreadMember(member, this)));
+    getThreadMembers(channelID, options) {
```
```suggestion
    getThreadMembers(channelID, options = {}) {
```
Otherwise will error if we don't pass `options`

lib/Client.js
```diff
@@ -3487,22 +3487,30 @@ class Client extends EventEmitter {
...
+        return this.requestHandler.request("GET", Endpoints.THREAD_MEMBERS(channelID), true, {
+            after: options.after,
+            limit: options.limit,
+            with_member: options.withMember
+        }).then((members) => members.map((member) => new ThreadMember(member, this)));
```
bsian03, line 3513: Need to snipe `id` into the member object otherwise cache cannot be upserted

bsian03 (changes requested):

lib/Client.js
conorwastakenwastaken: ThreadMember structure already handles that?

bsian03: You're thinking about the thread channel ID, not the member

conorwastakenwastaken: it handles the member too

bsian03: There is no `id` in the member object from the api, the guild members collection requires it to update the guild member.

image

conorwastakenwastaken: oh mb forgot discord was acoustic. fixed ??

bsian03 (approved):

PR #1499, opened 2024-04-30
Added missing createdAt prop from Interaction's docs

dmkng:
I noticed that the createdAt property is missing from the Interaction's documentation page, so I added it like it's added in the Channel's docs, but I have a suggestion for better readability of the documentation. Wouldn't it be better to show that the Interaction, Channel, Message etc are extending Base, and don't show the properties from the parent class? I think that the way it is done currently slightly hurts the readability because it doesn't tell the whole exact story, for example, GuildChannel extends Channel, and it is shown in the GuildChannel's docs, but the GuildChannel's docs duplicate properties that are in the Channel's docs, but not always all of them like in the Interaction's case. If this suggestion looks good for y'all, I can close this pull request and create another one with this suggestion. Maybe properties from the parent classes can be shown in a separate section "Inherited from ..." on the docs website, but I don't know if jsdoc supports something like that.

### DIFF
--- a/lib/structures/AutocompleteInteraction.js
+++ b/lib/structures/AutocompleteInteraction.js
@@ -9,6 +9,7 @@ const {InteractionResponseTypes} = require("../Constants");
  * @extends Interaction
  * @prop {Permission?} appPermissions The permissions the app or bot has within the channel the interaction was sent from
  * @prop {DMChannel | TextChannel | NewsChannel} channel The channel the interaction was created in. Can be partial with only the id if the channel is not cached
+ * @prop {Number} createdAt Timestamp of the interaction's creation
  * @prop {Object} data The data attached to the interaction
  * @prop {String} data.id The ID of the Application Command
  * @prop {String} data.name The command name
--- a/lib/structures/CommandInteraction.js
+++ b/lib/structures/CommandInteraction.js
@@ -16,6 +16,7 @@ const {InteractionResponseTypes} = require("../Constants");
  * @extends Interaction
  * @prop {Permission?} appPermissions The permissions the app or bot has within the channel the interaction was sent from
  * @prop {DMChannel | TextChannel | NewsChannel} channel The channel the interaction was created in. Can be partial with only the id if the channel is not cached
+ * @prop {Number} createdAt Timestamp of the interaction's creation
  * @prop {Object} data The data attached to the interaction
  * @prop {String} data.id The ID of the Application Command
  * @prop {String} data.name The command name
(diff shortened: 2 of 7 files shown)


bsian03:
createdAt is not documented in Interaction because it's already documented in Base.

> GuildChannel's docs duplicate properties that are in the Channel's docs, but not always all of them like in the Interaction's case

wdym?

dmkng:
> createdAt is not documented in Interaction because it's already documented in Base.
> 
> > GuildChannel's docs duplicate properties that are in the Channel's docs, but not always all of them like in the Interaction's case
> 
> wdym?

For example, `id` and `createdAt` are documented separately in TextChannel, GuildChannel, Channel and Base, but those properties are from Base. I made this pull request following the same logic but with `createdAt` and Interaction, but I suggested that it doesn't seem very readable.

dmkng:
I noticed just now that the Interaction have `id` property documented, but specific Interaction types does not have `id` documented.

bsian03:
> For example, `id` and `createdAt` are documented separately in TextChannel, GuildChannel, Channel and Base, but those properties are from Base.

Those properties are documented in Channel class only, which extends down to its child classes in the web documentation. so I don't exactly understand what you're trying to do here.

And you still haven't explained how it's not readable. please contact me in the Eris server if you have any feedback about it

either way, I'm going to mark this PR as a won't complete.

dmkng:
Okay, I understand that those extend down in the web documentation and not in the code, so my second commit is unneccesary, but still, the docs hide the fact that Channel, Message, Interaction etc are extending from Base and Interaction doesn't inherit the createdAt property in the documentation likely because it doesn't extend Base in the docs, so I added the createdAt property to the Interaction docs like it is done in e.g. Channel and Message.

dmkng:
> And you still haven't explained how it's not readable

As I said, it is slightly not readable because properties from parent classes are mixed with properties from the specific class, but I don't know if something can be done about it, if something can be done then it would be great. My main issue is missing `createdAt` property from Interaction, but I also think that the Base extend should be shown in the docs.

bsian03:
The docs are built manually. they haven't been updated in a while, so whatever is showing on the dev docs (ErisDev instead of just Eris in URL) is out of date. ~~Also there are already a few big PRs regarding interactions I need to look at which I haven't had the time to do recently, and~~ it's likely I'm going to do a general interaction refactor which will update everything.
correction: I forgot I closed those prs. but above statement still applies

The current release version of Eris (0.17.x) docs correctly reflects what is provided, even with all its flaws.
Again, any feedback please let me know in the Eris server

PR #1500, opened 2024-04-30
Added missing createdAt prop from Interaction's docs

dmkng:
I noticed that the `createdAt` property is missing from the Interaction's documentation page, so I added it like it's added in e.g. the Channel's docs.

### DIFF
--- a/lib/structures/Interaction.js
+++ b/lib/structures/Interaction.js
@@ -7,6 +7,7 @@ const {InteractionTypes} = require("../Constants");
  * Represents an interaction. You also probably want to look at PingInteraction, CommandInteraction, ComponentInteraction, AutocompleteInteraction, ModalSubmitInteraction, and UnknownInteraction.
  * @prop {Boolean} acknowledged Whether or not the interaction has been acknowledged
  * @prop {String} applicationID The ID of the interaction's application
+ * @prop {Number} createdAt Timestamp of the interaction's creation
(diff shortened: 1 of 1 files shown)


conorwastakenwastaken:
Afaik Discord don't send a timestamp with interactions? It's not on Discords documentation atm and even when I look at the raw interaction payload that Discord sends, no timestamp appears

dmkng:
> Afaik Discord don't send a timestamp with interactions? It's not on Discords documentation atm and even when I look at the raw interaction payload that Discord sends, no timestamp appears

`createdAt` property is inherited from Base, it extracts timestamp from `id` property. I tested it and it works like a charm.

conorwastakenwastaken:
> > Afaik Discord don't send a timestamp with interactions? It's not on Discords documentation atm and even when I look at the raw interaction payload that Discord sends, no timestamp appears
> 
> `createdAt` property is inherited from Base, it extracts timestamp from `id` property. I tested it and it works like a charm.

Right ik that, I just couldn't get it to work for me lol so if it works for you then clearly I'm doing something wrong. If it works in your testing then TS docs for the Interaction class will need to include createdAt too. Looks like TS docs only documents createdAt in UnknownInteraction atm

bsian03:
createdAt is already documented in Base

bsian03:
nvm sorry misunderstood this

bsian03 (approved):

PR #1501, opened 2024-05-04
remove localized system message handling

flav-code:


conorwastakenwastaken (approved):

lib/structures/Message.js
```diff
@@ -244,6 +244,22 @@ class Message extends Base {
...
+            case MessageTypes.STAGE_END: {
+                data.content = `${this.author.mention} ended ${this.content}`;
+                break;
+            }
+            case MessageTypes.STAGE_SPEAKER: {
+                data.content = `${this.author.mention} is now a speaker.`;
+                break;
+            }
+            case MessageTypes.STAGE_TOPIC: {
+                data.content = `${this.author.mention} changed the Stage topic: ${this.content}`;
```
```suggestion
                data.content = `${this.author.mention} changed the Stage topic: **${this.content}**`;
```

bsian03 (changes requested):
Just this small bold topic change text. Also will consult abal about that content reassignment since we're directly replacing data from discord.
otherwise LGTM

lib/structures/Message.js
flav-code: image

bsian03:
we decided the best way moving forward would be to remove all system localisation message handling, i.e. we will not be setting the message content if a particular message type required it

flav-code:
you want me to remove all data.content from message system ?

bsian03:
correct, the only thing that should be in content is what's reflected on discord

MCausc78:
Personally I would make a getter like `displayedContent` or `systemContent` rather replacing data.content

flav-code:
> Personally I would make a getter like `displayedContent` or `systemContent` rather replacing data.content

I agree

bsian03:
neither of us saw a legitimate and proper use case for having localised system messages in the message content. I'm happy for you to discuss this with me on discord though.

flav-code:
we could even do away with the user's call function, thus eliminating this unnecessary switching case

It would be nice to remove all functions not intended for bots from the lib, it would lighten the whole and remove unused code.

bsian03:
I'll work on that, the PR is massively out of date so I'll have to look over everything again but it's on my list of things to do asap, unless you're happy doing it of course

bsian03 (approved):

Issue #1498, opened 2024-04-28
guild voiceState note removed when user leave a stage channel


flav-code:
But if the user leave a normal voice channel he is removed from the collection

I don't know if this is a feature, but on my side i don't want to keep user not in a voice channel in the voiceStates collection

*When the user leave a stage channel the property channelID is set to null*

PR #1502, opened 2024-05-04
Don't keep suppressed user in cache because i'ts false by default

flav-code:


flav-code:
#1498

flav-code:
@bsian03 could you check ?
I already tested ok my side

bsian03:
I haven't been able to look at this yet. Once I do I'll let you know

flav-code:
my pr allows to remove the voiceState of a user who disconnects from a stage channel

bsian03 (approved):
stale voice state when leaving stage channel, #559 for precedence

PR #1505, opened 2024-05-12
fix(voice): OggOpusTransformer missing callback in _final

james58899:
I discovered this when I tried to increase `voiceDataTimeout` and noticed the `end` event was later than expected.
Because the stream does not emit `end`, it will stop playing after waiting for timeout, not when the file ends.

ref: https://nodejs.org/docs/latest-v20.x/api/stream.html#writable_finalcallback


bsian03:
Good find, thanks

PR #1508, opened 2024-05-27
fix: export everything under the lib for new versions of node

rubyowo:
I don't know why but apparently this was changed in node 18+ (I'm guessing)

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
       "./index.js"
     ],
     "./": "./",
+    "./*": "./*.js",
     "./esm": "./esm.mjs"
   },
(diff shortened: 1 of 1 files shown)


bsian03:
I'll have a look at this hopefully over the coming week. What export type are you using by the way?

rubyowo:
I am not *directly* using it like this, a [plugin](https://github.com/YetAnotherConnor/ReactionThreadsMenu) for a discord bot I use `require`s the endpoints file

Wdym export type?

bsian03 (approved):

PR #1521, opened 2024-07-20
fix(Member): fix updating from voice state updates

Snazzah:
Since this is kinda active again, this is a mirror of this PR: https://github.com/projectdysnomia/dysnomia/pull/131
> Noticed that voice updates don't also update member's roles, so this should update that stuff (and also update user stuff)
> 
> This is really only useful for use cases where you aren't receiving member/user updates already.

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -78,6 +78,18 @@ class Member extends Base {
     }
 
     update(data) {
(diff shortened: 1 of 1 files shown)


bsian03 (approved):
This seems a little hacky to me but it's fine

PR #1525, opened 2024-08-08
fix(Member): fix voice state not updated when `member` field is missing.

### LINKED ISSUE #1522: Voice state not added when the bot starts
If I'm connected to a channel and it starts, the bot doesn't add a voiceState.
before the update, it worked fine
https://github.com/abalabahaha/eris/pull/1521

babiabeo:
Fix: #1522

Currently, the voice state is only updated if the data has `member` field. However, the `member` field is optional in [Voice State object](https://discord.com/developers/docs/resources/voice#voice-state-object-voice-state-structure), leading to incorrect update handling. This PR ensures that the voice state is updated even when the `member` field is not present.

### DIFF
--- a/lib/structures/Member.js
+++ b/lib/structures/Member.js
@@ -96,7 +96,7 @@ class Member extends Base {
             this.status = data.status;
         }
         if(data.joined_at !== undefined) {
-            this.joinedAt = data.joined_at ?  Date.parse(data.joined_at) : null;
+            this.joinedAt = data.joined_at ? Date.parse(data.joined_at) : null;
         }
         if(data.client_status !== undefined) {
             this.clientStatus = Object.assign({web: "offline", desktop: "offline", mobile: "offline"}, data.client_status);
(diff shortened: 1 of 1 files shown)


bsian03:
Thanks for getting this - will review and test soon
cc @flav-code

TTtie (approved):
From my very brief local testing, it seems to be working, both at startup and while the voice state is updated with the new member data!

lib/structures/Member.js
```diff
@@ -88,7 +88,9 @@ class Member extends Base {
             } else if(data.channel_id || data.mute || data.deaf || data.suppress) {
                 this.guild.voiceStates.update(data);
             }
-            data = data.member;
+            if(data.hasOwnProperty("member")){
```
TTtie: a very small style nit I just noticed, there should be a space between the if condition and the block inside
```suggestion
            if(data.hasOwnProperty("member")) {
```

flav-code:
LGTM

flav-code (changes requested):
Style

flav-code:
image

babiabeo:
It is weird that the linter doesn't detect that one.

bsian03:
Will fix the linter, otherwise lgtm

PR #1530, opened 2024-08-22
typo(InteractionCreateMessage)

flav-code:


DonovanDMC:
initial response returns void

flav-code:
And if you defer it returns the message

DonovanDMC:
I'm aware
the point is that isn't a typo, it's intentional
You should take a look at every other interaction class

flav-code:
I saw that the other interactiosn have the same,
If worst comes to worst, I'll create my own function createMessageAfterDefer(): Promise<Message>

DonovanDMC:
have you considered createFollowup

flav-code:
My bad...

I always thought that the createFollowup created a new message, but actually if there is no message yet the followup edit the message

Thanks we can close this

PR #1534, opened 2024-08-27
Makes channel_types an array of ChannelTypes

areodot:
Fixed the channel_types property so it would not throw a type error when i try to use an array for channel_types (and to throw an error when not) as stated in the [Application Command Option Structure](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-structure)


bit0r1n:
failed attempt of making `channel_types` into an array number 4

bsian03:
this pr was closed, idk why you're reviewing it

PR #1535, opened 2024-08-28
chore: modernize linting

abalabahaha:
All autofix changes are isolated in the last commit for `git blame`. Don't squash.

Supersedes #1531.


lib/Constants.d.ts
bsian03, line 1: Should really add `unknown` comments in audit log actions so constants value line up with each other

lib/voice/Piper.js
```diff
@@ -105,7 +105,8 @@ class Piper extends EventEmitter {
             this.streams.push(source.once("error", (e) => this.stop(e)));
         }
 
-        if(options.format === "opusPackets") { // eslint-disable no-empty
+        // eslint-disable-next-line no-empty
+        if(options.format === "opusPackets") {
```
bsian03: Can probably remove this if it isn't being used? It isn't in the options list nor is it referenced anywhere else

bsian03: Not outdated, still relevant https://github.com/abalabahaha/eris/blob/8e8569ae40a299dcdea52496b65329398d29bc46/lib/voice/Piper.js#L109

babiabeo: Using `switch case` would be better

bsian03: would be out of scope with this PR, but idk if @abalabahaha wants to change it here or leave it for another PR

abalabahaha: Out of scope for a linting PR, but it's useful and doesn't cost much to keep around so should probably fix docs

lib/Constants.d.ts
abalabahaha: It's a fairly long list and there are actual decimal number ranges (unlike the bitfields), so it seems better to not have `APPLICATION_COMMAND_PERMISSION_UPDATE` add 25 spaces to `GUILD_UPDATE` from 60 lines away.

PR #1544, opened 2024-09-22
fix(VoiceConnection): do not reconnect if close code is 4014 (Disconnected)

### LINKED ISSUE #1437: VoiceConnection disconnect event not emitting
When the client is forcefully disconnected from a voice channel (right click -> disconnect), or possibly other mean as well, the disconnect event does not emit. This means there's no way to detect this, streams will just keep trying to play, hilarity ensues.

Whenever the websocket closes with `4014: Disconnected` and there's a channel (not sure why there wouldn't be one), it is setting `reconnecting = true`, therefore not emitting the disconnect event. https://github.com/abalabahaha/eris/blob/e43418dae558a8fc7108fbfd820ff6cd8c6a2db3/lib/voice/VoiceConnection.js#L360-L367

When it eventually tries to reconnect, it's getting stuck here: https://github.com/abalabahaha/eris/blob/e43418dae558a8fc7108fbfd820ff6cd8c6a2db3/lib/voice/VoiceConnection.js#L172-L174

What I personally think should happen: When it gets disconnected for any reason that is not an error, it shouldn't even try to reconnect and just fire the disconnect event. Users should be able to decide themselves if they want to reconnect or not. If this behavior needs to stay for some reason, maybe a "reconnect" option for channel.join(), being true by default for backwards compatibility.

babiabeo:
Fixes: #1437

[As described in the docs](https://discord.com/developers/docs/topics/opcodes-and-status-codes#voice-voice-close-event-codes), voice should not reconnect if close code is 4014 (Disconnected).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -876,7 +876,7 @@ declare namespace Eris {
     channelDelete: [channel: Exclude<AnyChannel, GroupChannel>];
     channelPinUpdate: [channel: TextableChannel, timestamp: number, oldTimestamp: number];
     channelUpdate: [channel: AnyGuildChannel, oldChannel: OldGuildChannel | OldForumChannel | OldGuildTextChannel | OldVoiceChannel]
-    | [channel: GroupChannel, oldChannel: OldGroupChannel];
+      | [channel: GroupChannel, oldChannel: OldGroupChannel];
     connect: [id: number];
     debug: [message: string, id?: number];
     disconnect: [];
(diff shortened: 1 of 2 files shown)


Snazzah:
I'll test this later but there are some possible issues:
1. `.disconnect()` with `reconnecting = false` will update the voice state and disconnect the client, this is only bad because things like voice server updates (changing the voice region) would also give a 4014, which we wouldn't want to full disconnect from and would rather wait for the next `VOICE_SERVER_UPDATE` event. On a 4014, `data.endpoint` would be set to `null` which would signal to `.connect` that it should wait for an event:
```js
if (!data.endpoint) {
  return; // Endpoint null, wait next update.
}
```
2. `VOICE_STATE_UPDATE`s would already update/disconnect the connection if the client was kicked or if the channel was removed.
3. A failed reconnect would properly disconnect the client if it came to it, emitting the `disconnect` event.

I've tried kicking the client from a VC and it seemed to disconnect just fine because of the `VOICE_STATE_UPDATE`, maybe there should be a separate timeout for that update waiting.

bsian03:
~~I didn't read that essay, but ideally the only change here would be just to disconnect, existing functions shouldn't change unless there is a clear benefit or if it is absolutely necessary~~ nvm read context and understand this a bit more now

a 4014 here in this case means you shouldn't reconnect to that websocket. reconnecting to the voice channel on the other hand is an entirely different question. Assuming the disconnect was not intentional client side, we should reconnect to the voice channel on a new websocket

babiabeo:
I see. It seems that I misunderstood the meaning of reconnecting here. Will close this PR for better solution.

PR #1547, opened 2024-09-23
feat: handle voice channel effect send events

Snazzah:
Ref https://github.com/discord/discord-api-docs/pull/6025

PR #1550, opened 2024-09-24
feat(Soundboard): Add Soundboard Sound functionality

bsian03:
~~Note: Currently there is an incompatibility with storing soundboard sounds in cache, as there is no `id` in the soundboard sound structure but required in Collection.~~

~~The PR in it's current state (2024-09-25T03:41Z) will not work, as gateway logic is written as if the collection were to exist~~

#1547 to be folded into this PR once complete


babiabeo:
Perhaps add a getter for `id` that returns `soundID`? (Though some people might think it is redundant)

```js
get id() {
  return this.soundID;
}
```

bsian03:
Sure, however that still doesn't solve the main issue of not being able to properly store the data directly into the collection

I am discussing this with abal to see what our options are

lib/Client.js
```diff
@@ -1874,6 +1905,24 @@ class Client extends EventEmitter {
...
+  /**
+   * Edit a guild soundboard sound
+   * @arg {String} guildID The guild ID where the sound will be edited
+   * @arg {String} soundID The ID of the soundboard sound
+   * @arg {Object} options The properties to edit
+   * @arg {String?} [options.emojiID] The ID of the relating custom emoji (mutually exclusive with options.emojiName)
+   * @arg {String?} [options.emojiName] The name of the relating default emoji (mutually exclusive with options.emojiID)
+   * @arg {String} [options.name] The name of the soundboard sound (2-32 characters)
+   * @arg {Number?} [options.volume] The volume of the soundboard sound, between 0 and 1
+   * @arg {String} [options.reason] The reason to be displayed in audit logs
```
bsian03, line 1919: separate this into its own argument?

PR #1557, opened 2024-09-30
fix(constants): update gateway/rest versions in types

Snazzah:
Updated the types to be the same as the JS file.


bsian03:
the number of times I keep telling myself to do this 🤦‍♂️

PR #1561, opened 2024-10-04
File parameter changes

areodot:
This pull request deprecates the file parameter, renaming it to files and moving it under the options (hope you understood I didn’t know how to put it, if you didn’t please look at the code)


bsian03:
This is something that we still have not discussed internally regarding handling file behaviour, as such I won't be reviewing this PR in its current state.

PR #1564, opened 2024-10-20
Fixed some example codes

mahorabhossainmidul:
HI, this pool have nothing special , i just fix some example codes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,10 @@ Ping Pong Example
 const Eris = require("eris");
 
 // Replace TOKEN with your bot account's token
-const bot = new Eris("Bot TOKEN", {
+const bot = Eris("Bot TOKEN", {
     intents: [
-        "guildMessages"
+        "guildMessages",
+        "messageContent"
     ]
 });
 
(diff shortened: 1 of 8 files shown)


coolcalcacol:
Eris is a class, so should be called with `new`as with the original examples

mahorabhossainmidul:
https://github.com/abalabahaha/eris/blob/dev/index.js#L5 Here you can Eris is a function that returns new Client class.

babiabeo:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new#description):

> 4. If the constructor function returns a non-primitive, this return value becomes the result of the whole new expression.

So if I understand correctly, the `Eris()` function returns a class which is a non-primitive. It means that both `Eris()` and `new Eris()` return the same result.

mahorabhossainmidul:
Yes you are right. Actually when I was using typescript I got a type error so I replaced `new Eris` with `Eris`.

bsian03:
I'd argue that if you're using typescript, you're already familiar with what to do and not to do. That being said this *technically* is a minor bug, but I have my reservations (aside from the point I just made) with this being the solution. I'll loop back to it once the API facing features are sorted out

PR #1583, opened 2024-12-20
Add application to client options in constructor

someramsey:
`Client.application` should be configurable from the constructor. This pull request just adds application property to the options inside the Client constructor.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -507,10 +507,10 @@ declare namespace Eris {
     type: ApplicationRoleConnectionMetadataTypes;
   }
   interface ClientOptions {
-    application?: { id: string; flags?: number };
     /** @deprecated */
     agent?: HTTPSAgent;
     allowedMentions?: AllowedMentions;
+    application?: { id: string; flags?: number };
     autoreconnect?: boolean;
     compress?: boolean;
     connectionTimeout?: number;
--- a/index.d.ts
+++ b/index.d.ts
@@ -507,6 +507,7 @@ declare namespace Eris {
     type: ApplicationRoleConnectionMetadataTypes;
   }
   interface ClientOptions {
+    application?: { id: string; flags?: number };
     /** @deprecated */
     agent?: HTTPSAgent;
     allowedMentions?: AllowedMentions;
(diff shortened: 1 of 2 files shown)


retrouser955 (commented):
Please add JSDoc for the option

index.d.ts
```diff
@@ -507,6 +507,7 @@ declare namespace Eris {
     type: ApplicationRoleConnectionMetadataTypes;
   }
   interface ClientOptions {
+    application?: { id: string; flags?: number };
```
retrouser955: This needs to be sorted alphabetically

someramsey:
You can merge it now

lib/Client.js
```diff
@@ -152,6 +153,13 @@ class Client extends EventEmitter {
       ws: {},
       reconnectDelay: (lastDelay, attempts) => Math.pow(attempts + 1, 0.7) * 20000,
     }, options);
+
+    if (options.application) {
+      this.application = {
+        id: options.application.id,
+        flags: options.application.flags ?? 0,
```
bsian03, line 160: This is not backwards compatible

lib/Client.js
```diff
@@ -79,6 +79,7 @@ class Client extends EventEmitter {
    * Create a Client
    * @arg {String} token The auth token to use. Bot tokens should be prefixed with `Bot` (e.g. `Bot MTExIHlvdSAgdHJpZWQgMTEx.O5rKAA.dQw4w9WgXcQ_wpV-gGA4PSk_bm8`). Prefix-less bot tokens are [DEPRECATED]
    * @arg {Object} options Eris client options
+   * @arg {Object} [options.application] Object containing the bot application's ID and its public flags
```
bsian03, line 82: Need to document child properties

bsian03 (changes requested):

index.d.ts
bsian03: Props also need to be sorted

lib/Client.js
someramsey: Ah. Anyways I made this commit ages ago. I don't even remember what its about anymore. So I wont be pushing any further changes here you can close it

PR #1590, opened 2025-08-04
Fixing and add Stuffs

ratniaprivte:
[Pull request reason]
Adding logo for the library
add some changes and delete some stuffs not useful,
so the library could be at least more nice looking.


ratniaprivte:
thx

bsian03:
Rejecting this, not necessary.

In the future please discuss this with me as per contributing guidelines, thank you

PR #1592, opened 2025-08-24
perf(Channles): removed channelGuildMap to improve performance

ZixeSea:
### What changed
Removed `channelGuildMap` from client to improve RAM usage. Added `guildID` parameter to existing functions to make existing functions work as they did before. Since Discord provided the guild ID (through `packet.d.guild_id`), this change would only impact direct calls by users on functions within the client.

### Functions impacted
- Client#editChannelPosition
- Client#getChannel
- Client#joinVoiceChannel
- Client#leaveVoiceChannel

## BREAKING CHANGE
For the small number of people that call the above-mentioned functions (under functions impacted) directly through the client object. This would require an additional parameter for the guild ID.


flav-code:
+1

with 500k channels you can save more than 200mb of ram

PR #1594, opened 2025-08-25
feat: Support Pin Messages permission

conorwastakenwastaken:
Ref:
- https://github.com/discord/discord-api-docs/pull/7766


bsian03:
Will assume you've calculated the permission value correctly

PR #1597, opened 2025-08-29
feat(Guild): Guild incident actions

conorwastakenwastaken:
Ref:
- https://github.com/discord/discord-api-docs/pull/7360

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2276,7 +2276,7 @@ declare namespace Eris {
       options: { name?: string; roles?: string[] },
       reason?: string
     ): Promise<Emoji>;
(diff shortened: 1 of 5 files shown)


index.d.ts
```diff
@@ -1159,6 +1163,12 @@ declare namespace Eris {
     reason?: string;
     user: User;
   }
+  interface GuildIncidentData {
+    dmsDisabledUntil: number | null;
+    dmSpamDetectedAt: number | null;
```
bsian03, line 1168: gonna nitpick here dmSpam should technically go before dmsDisabled

conorwastakenwastaken: Is the alphabet down south different to the one up here? Up North, D usually comes before P. Or does uppercase take priority for some reason

bsian03: oh no i completely misread that oops. also sheep.

PR #1604, opened 2025-09-28
chore(examples): update bot token placeholder and add components v2 example

marawny:


reinacchi (commented):
I don't see a need to remove the Bot placeholder since Eris supports tokens without the "Bot" prefix.

https://github.com/abalabahaha/eris/blob/dev/lib/gateway/Shard.js#L1951

marawny:
I didn't remove the bot placeholder, the issue is `BOT` will not work as prefix it should be exactly `Bot`.

also see: https://discord.com/channels/831967755447828491/831975480877514822/1421983849612709888

bsian03:
Just to clarify here - prefixless tokens are deprecated and have been so for a while. So while it is still possible to not prefix the token, it is strongly discouraged

marawny:
just merge it 🤠

